PR #37, opened 2020-05-13
use regular pre-commit commit, skip pylint

sebhmg:
no need for pre-commit config specific to CI


sebhmg:
see https://github.com/MiraGeoscience/geoh5py/pull/38 instead
I was just using this PR on top of "more_docs" branch to make sure it passed

PR #40, opened 2020-09-07
Reference data

domfournier:
General updates

- Add reference mapping

### DIFF
--- a/geoh5py/io/h5_reader.py
+++ b/geoh5py/io/h5_reader.py
@@ -257,7 +257,7 @@ class H5Reader:
         return property_groups
 
     @classmethod
-    def fetch_value_map(cls, h5file: Optional[str], uid: uuid.UUID) -> Optional[float]:
+    def fetch_value_map(cls, h5file: Optional[str], uid: uuid.UUID) -> Optional[dict]:
         """
         Get data :obj:`~geoh5py.data.data.Data.value_map`
 
--- a/geoh5py/io/h5_reader.py
+++ b/geoh5py/io/h5_reader.py
@@ -75,6 +75,14 @@ class H5Reader:
                 "Color map"
             ][:]
 
+        if "Value map" in entity["Type"].keys():
+            value_map = entity["Type"]["Value map"][:]
+            mapping = {}
+            for key, value in value_map.tolist():
+                mapping[key] = value
+
+            type_attributes["entity_type"]["value_map"] = mapping
+
         # Check if the entity has property_group
         if "PropertyGroups" in entity.keys():
             for pg_id in entity["PropertyGroups"].keys():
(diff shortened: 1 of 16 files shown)


domfournier:
@sebhmg Whenever you are ready to review this.

geoh5py/io/h5_reader.py
```diff
@@ -248,6 +256,27 @@ def fetch_property_groups(
 
         return property_groups
 
+    @classmethod
+    def fetch_value_map(cls, h5file: Optional[str], uid: uuid.UUID) -> Optional[float]:
```
sebhmg, line 260: **question:** looks like a copy paste of `fetch_values`. Should return ReferenceValueMap instead?

with check on the type of values?

geoh5py/io/h5_writer.py
```diff
@@ -434,12 +457,76 @@ def write_color_map(
...
+                compression_opts=9,
+            )
+
+        if close_file:
+            h5file.close()
+
+    @classmethod
+    def write_value_map(
+        cls,
+        entity_type: "shared.EntityType",
```
sebhmg, line 474: **question (non-blocking):** since we write value map, should not the entity by typed more specifically?

geoh5py/io/h5_writer.py
```diff
@@ -434,12 +457,76 @@ def write_color_map(
...
+        file: Optional[Union[str, h5py.File]] = None,
+        close_file: bool = True,
+    ):
+        """
+        Add :obj:`~geoh5py.data.value_map.ColorMap` to a
```
sebhmg, line 479: **question:** I am confused: is this about color map or reference data?

sebhmg (changes requested):
Looks good. I did not find anything major. Please see some questions/suggestions in the review.

geoh5py/io/h5_reader.py
domfournier: Correct, should return Optional[dict]

geoh5py/io/h5_writer.py
domfournier: Sorry, i m not getting you here? Like passing the value_map itself?

geoh5py/io/h5_writer.py
domfournier: Good eye, typo.

geoh5py/io/h5_writer.py
sebhmg: entity is the object data here right? Should the type be `ReferenceData`?

domfournier: so the `value_map` is stored under the `entity_type`. We could pass the `ReferenceData` and get its `entity_type`. Right now I m just passing the nearest level which the data is under. Up to you.

domfournier:
@sebhmg Can you check pre-commit and pylint again? Looks like something with poetry again

geoh5py/io/h5_writer.py
sebhmg: ho, right the mapping it is stored on the type. So it could take `DataType` instead of `EntityType`

sebhmg (approved):

PR #47, opened 2021-03-06
GEOIO-66, GEOIO-64: Drillholes

domfournier:
Implement the `Drillhole` object class with functionality for adding log and interval data.
Also includes some general maintenance items
- Add no-data-value for integer and float data, convert to nan on import 
- Assign English entity_type name instead of UUID string
- Check/convert binary type strings
- Generalize the coordinate properties (vertices, trace, survey) as numpy.recordarray - simplifies the H5writer 

Related to GEOIO-66, GEOIO-64


geoh5py/data/float_data.py
```diff
@@ -50,11 +51,17 @@ def values(self, values):
         self._values = self.check_vector_length(values)
 
     def check_vector_length(self, values) -> np.ndarray:
```
sebhmg, line 53: **suggestion:** somewhat unclear semantic

since this is a member function, called by another function to fixup `self._values`, semantic would be clearer if it simply updates `self._values` instead of returning it.

Also when it returns, there is no guaranty that `self.n_values == len(self._values)`
Should it just be named `flatten_values()`?
Are there any case where we would keep non-flatten values?

geoh5py/io/h5_reader.py
```diff
@@ -291,6 +309,16 @@ def fetch_values(cls, h5file: Optional[str], uid: uuid.UUID) -> Optional[float]:
         name = list(project.keys())[0]
         if "Data" in list(project[name]["Data"][cls.uuid_str(uid)].keys()):
             values = np.r_[project[name]["Data"][cls.uuid_str(uid)]["Data"]]
+            if not isinstance(values[0], (str, bytes)):
+                if values.dtype in [float, "float64", "float32"]:
+                    ind = values == FLOAT_NDV
+                else:
+                    ind = values == INTEGER_NDV
+                    values = values.astype("float64")
```
sebhmg, line 317: **question:** can really a float64 hold any integer values? Should it not store them in an array of integers ?

It is unlikely the data uses the full min,max range of integer values, but in theory it could use some integer values that do not fit in a float64 representation

geoh5py/data/float_data.py
domfournier: Just for context, something like this is needed because geoh5 can store vectors shorter than n_vertices if the bottom is all ndv.

I m not fully following you here, let's talk about it.

geoh5py/data/integer_data.py
```diff
@@ -15,30 +15,15 @@
...
 
-import numpy as np
 
-from .data import Data, PrimitiveTypeEnum
+from .data import PrimitiveTypeEnum
+from .float_data import FloatData
 
 
-class IntegerData(Data):
+class IntegerData(FloatData):
```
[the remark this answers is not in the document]
domfournier: Done

geoh5py/io/h5_reader.py
domfournier: Issue is how to handle the NDV. Unfortunately, numpy doesn't support nan on integer type. My thought was to convert everything to float, then transfer back on the write?

geoh5py/groups/container_group.py
```diff
@@ -34,6 +34,9 @@ def __init__(self, group_type: GroupType, **kwargs):
         assert group_type is not None
         super().__init__(group_type, **kwargs)
 
+        if self.entity_type.name == "Entity":
```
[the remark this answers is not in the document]
domfournier: Yep, any recommendation on how/where to name those?

geoh5py/io/h5_reader.py
sebhmg: not sure to follow: NDV for integer data is an int (-2147483648)

sebhmg (approved):
(sorry for the delay)
Looks all good. Thanks.
Just this remaining question about using float64 for integers

geoh5py/io/h5_reader.py
domfournier: I meant on the python side ... it is nicer to work with numpy.nan, but that's not supported if the array is type int32...

sebhmg: ok, seems fine then. float64 seems able to store any value from int32

PR #49, opened 2021-03-23
GEOIO-98: Fix reference to python 3.6

domfournier:
GEOIO-98: Bump reference to python 3.6 -> 3.7 in docs.


sebhmg (approved):
and pyproject.toml already tells python ^3.7

looks good

setup.py
```diff
@@ -37,7 +37,7 @@
```
sebhmg, line 37: @domfournier, I would be curious to know which tools requires `setup.py`

Using Poetry is supposed to save the hassle of having to maintain `setup.py`, `requirements.txt`, `setup.cfg`, `MANIFEST.in`, `Pipfile`, …

domfournier: I can move the `environment.yml` to the docs, and we just kill the setup.py?

sebhmg (approved):

PR #59, opened 2021-04-15
GEOPY-168: Empty FROM-TO on subsequent interval data

domfournier:
- Fix for subsequent add_data for drillhole intervals. The issue was on the reset of drillhole cells before adding the new FROM and TO. 
- Attempt to fix missing images in docs


sebhmg (changes requested):
should be covered by tests

domfournier:
@sebhmg Ok ready for review.

tests/create_drillhole_data_test.py
```diff
@@ -47,43 +47,73 @@ def test_create_drillhole_data():
...
+        ), "Error with number of cells on interval data creation."
+        assert well.n_vertices == (
+            from_to_a.size + 4
+        ), "Error with number of vertices on interval data creation."
+        assert not np.any(
+            np.isnan(well.get_data("FROM")[0].values)
+        ), "FROM values not fully set."
+        assert not np.any(
+            np.isnan(well.get_data("TO")[0].values)
+        ), "TO values not fully set."
```
sebhmg, line 89: **question (non-blocking):** should it also assert on the expected data size for `FROM` and `TO`?

sebhmg (approved):

tests/create_drillhole_data_test.py
domfournier: Added

PR #64, opened 2021-04-16
Add files via upload

RichardScottOZ:
Very basic usage


domfournier:
Hi @RichardScottOZ , can you tell me more about your notebook? What kind of text can we add to describe the goal of the example?

RichardScottOZ:
My idea is to have a notebook of examples of  - this is how you get data/information from a project that already exists - e.g. syntax to interface etc., as opposed to the make a new one synthetic variety.

RichardScottOZ:
Just a skeleton currently.

RichardScottOZ:
e.g. here's my geoscience analyst project, how do I grab stuff out of it to use elsewhere in standard python fashion

as opposed to the examples which are all take stuff from elsewhere and put into Analyst?

PR #99, opened 2021-06-03
GEOPY-63

domfournier:
**GEOPY-63 - unit test about saving only modified attributes** 
 Test that only modified properties are being written to disk.

### DIFF
--- a/tests/save_modified_properties_test.py
+++ b/tests/save_modified_properties_test.py
@@ -24,8 +24,6 @@ import numpy as np
 from geoh5py.objects import Points
 from geoh5py.workspace import Workspace
 
-# from geoh5py.io.h5_writer import H5Writer
-
 
 @patch("geoh5py.io.h5_writer.H5Writer.write_data_values")
 @patch("geoh5py.io.h5_writer.H5Writer.write_coordinates")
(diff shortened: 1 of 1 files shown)


sebhmg (changes requested):
the purpose of the test is to check that only necessary parts are updated in the geoh5 file.
I would think the test should be on `h5writer.write_attributes()` (and other function: expected ones are call and other are not called).
Possibly check also a first intermediate state on `Entity.modified_attributes`

tests/save_modified_properties_test.py
```diff
@@ -34,24 +41,35 @@ def test_save_modified_properties():
...
+            not write_data_values.called
+        ), f"{write_data_values} should not have been called."
 
-            workspace.finalize()
+        points.add_data({"rando": {"values": np.ones(n_data)}})
 
-            assert (
-                len(allow_move.mock_calls) == 2
-            ), "Not all attributes were re-written."
+        assert write_data_values.called, f"{write_data_values} should have been called."
```
sebhmg, line 69: **question (non-blocking):** is there no need for a `finalize`?
unclear to me when finalize is required. Only when creating new entities?
The doc says: _Finalize the h5file by checking for **updated** entities_

sebhmg (approved):
👍 

I just have one question in comments. Thanks

PR #101, opened 2021-06-11
GEOPY-249

domfournier:
**GEOPY-249 - Add fails on imported drillhole**

### DIFF
--- a/geoh5py/objects/drillhole.py
+++ b/geoh5py/objects/drillhole.py
@@ -63,7 +63,7 @@ class Drillhole(Points):
         self._deviation_y = None
         self._deviation_z = None
 
-        self._default_colocation_distance = 1e-2
+        self._default_collocation_distance = 1e-2
 
         super().__init__(object_type, **kwargs)
 
@@ -254,16 +254,16 @@ class Drillhole(Points):
         self._locations = None
 
     @property
-    def default_colocation_distance(self):
+    def default_collocation_distance(self):
         """
-        Minimum colocation distance for matching depth on merge
+        Minimum collocation distance for matching depth on merge
         """
-        return self._default_colocation_distance
+        return self._default_collocation_distance
 
-    @default_colocation_distance.setter
-    def default_colocation_distance(self, tol):
+    @default_collocation_distance.setter
+    def default_collocation_distance(self, tol):
         assert tol > 0, "Tolerance value should be >0"
-        self._default_colocation_distance = tol
+        self._default_collocation_distance = tol
 
     @property
     def trace(self) -> Optional[np.ndarray]:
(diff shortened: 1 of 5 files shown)


geoh5py/objects/drillhole.py
```diff
@@ -61,6 +63,8 @@ def __init__(self, object_type: ObjectType, **kwargs):
         self._deviation_y = None
         self._deviation_z = None
 
+        self._tolerance = 1e-2
```
sebhmg, line 66: **suggestion (naming, semantic):** `default_colocation_distance` ? or just `default_colocation`

- "default" because this get used only if not specified in `attrs` of `add_data()` (should also rename tolerance to `colocation` in those `attrs`?)

geoh5py/objects/drillhole.py
```diff
@@ -333,10 +349,11 @@ def add_data(
 
             attr["name"] = name
 
-            tolerance = 1e-2
             if "tolerance" in list(attr.keys()):
                 assert attr["tolerance"] > 0, "Input depth 'tolerance' must be >0."
                 tolerance = attr["tolerance"]
+            else:
+                tolerance = self.tolerance
```
sebhmg, line 356: **suggestion (test):** have a test on using this tolerance when adding data  (test for one data added, and one replaced)

BTW what is the expected behavior: is data replaced if within the tolerance or ignored?

geoh5py/objects/drillhole.py
```diff
@@ -211,7 +211,16 @@ def surveys(self):
         :obj:`numpy.array` of :obj:`float`, shape (3, ): Coordinates of the surveys
         """
         if (getattr(self, "_surveys", None) is None) and self.existing_h5_entity:
-            self._surveys = self.workspace.fetch_coordinates(self.uid, "surveys")
+            surveys = self.workspace.fetch_coordinates(self.uid, "surveys")
+            if len(surveys) == 1:
```
sebhmg, line 219: **suggestion (test):** have a unit test for survey with a single value

sebhmg (changes requested):
code looks good. better with some new unit tests for these changes

geoh5py/objects/drillhole.py
domfournier: Done

geoh5py/objects/drillhole.py
domfournier: No replacement of the data value itself, the entry just gets inserted at a matching existing depth. So I guess, in a way, depth is rounded to the matching value yes.

sebhmg: because is is a new drillhole log, not adding entries into an existing log. right

geoh5py/objects/drillhole.py
domfournier: Added

geoh5py/objects/drillhole.py
```diff
@@ -15,6 +15,8 @@
 #  You should have received a copy of the GNU Lesser General Public License
 #  along with geoh5py.  If not, see <https://www.gnu.org/licenses/>.
 
+# pylint: disable=R0902
```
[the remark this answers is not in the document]
domfournier: Ok

sebhmg (approved):

geoh5py/objects/drillhole.py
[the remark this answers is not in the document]
sebhmg: created GEOPY-260

PR #113, opened 2021-07-23
GEOPY-85

domfournier:
**GEOPY-85 - Support of geoh5 DC object in geoh5py** 
 - Implement the DCIP survey object (PotentialElectrode and CurrentElectrode).
- Add docs
- Unitest


geoh5py/objects/surveys/direct_current.py
```diff
@@ -0,0 +1,282 @@
...
+        try:
+            return self.workspace.get_entity(currents)[0]
+        except IndexError:
+            print("Associated CurrentElectrode entity not found in Workspace.")
+            return None
+
+    @current_electrodes.setter
+    def current_electrodes(self, current_electrodes: CurrentElectrode):
+
+        assert isinstance(current_electrodes, CurrentElectrode), (
```
benk-mira, line 121: I think we should try to stick to try/except raise blocks or if condition raise in production code to raise custom errors.  I know they're more verbose, but assertions a more meant for debugging and testing.

geoh5py/objects/surveys/direct_current.py
```diff
@@ -0,0 +1,282 @@
...
+import uuid
+
+import numpy as np
+
+from ...data import Data, ReferencedData
+from ..curve import Curve
+from ..object_type import ObjectType
+
+
+class PotentialElectrode(Curve):
```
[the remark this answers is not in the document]
benk-mira: Hmm... I see there's also a CurrentElectrode class with both properties, I must be missing the point...

geoh5py/objects/surveys/direct_current.py
```diff
@@ -0,0 +1,282 @@
...
+        :return: Default unique identifier
+        """
+        return cls.__TYPE_UID
+
+    @property
+    def current_electrodes(self):
+        """
+        The associated current electrode object (sources).
+        """
+        ...
```
benk-mira: Is it ok for this to be empty?

geoh5py/objects/surveys/direct_current.py
```diff
@@ -0,0 +1,282 @@
...
+
+    @current_electrodes.setter
+    def current_electrodes(self):
+        ...
```
benk-mira, line 217: ...and this too?

geoh5py/shared/utils.py
```diff
@@ -51,15 +51,15 @@ def fetch_h5_handle(
 
 def match_values(vec_a, vec_b, collocation_distance=1e-4):
     """
-    Find indices of matching values between two vecays, within collocation_distance.
+    Find indices of matching values between two arrays, within collocation_distance.
 
-    :param: vec_a, list or numpy.ndvecay shape (*,)
```
benk-mira, line 56: never seen that type before!

tests/create_survey_dcip_test.py
```diff
@@ -0,0 +1,82 @@
...
+from geoh5py.shared.utils import compare_entities
+from geoh5py.workspace import Workspace
+
+
+def test_survey_dcip():
+
+    name = "TestCurrents"
+    n_data = 12
+
+    with tempfile.TemporaryDirectory() as tempdir:
```
benk-mira, line 35: Could use tmp_path fixture here, obviously not required though..

tests/create_survey_dcip_test.py
```diff
@@ -0,0 +1,82 @@
...
+        path = Path(tempdir) / r"testDC.geoh5"
+
+        # Create a workspace
+        workspace = Workspace(path)
+
+        # Create sources along line
+        x_loc, y_loc = np.meshgrid(np.arange(n_data), np.arange(-1, 3))
+        vertices = np.c_[x_loc.ravel(), y_loc.ravel(), np.zeros_like(x_loc).ravel()]
+        parts = np.kron(np.arange(4), np.ones(n_data)).astype("int")
```
benk-mira, line 44: Isn't this the same as np.tile(np.arange(4), n_data)?? I don't think too many people outside of Eldad's circle know about the kronecker product trick, maybe best to avoid for the 1d array case?

geoh5py/objects/surveys/direct_current.py
```diff
@@ -0,0 +1,282 @@
...
+    """
+    Ground potential electrode (receiver).
+
+    .. warning:: Partially implemented.
+
+    """
+
+    __TYPE_UID = uuid.UUID("{275ecee9-9c24-4378-bf94-65f3c5fbe163}")
+
+    def __init__(self, object_type: ObjectType, **kwargs):
```
benk-mira, line 35: There's no way to know from the docs or the init signature what the attributes for this class are (downside of opening up kwargs).  I would recommend documenting, or replacing the kwargs with a list of arguments if it's short enough, or both!  Another option is to set all the valid attributes for the class to None in the init and then update with kwargs.

benk-mira: ...this would allow you to type the arguments too!

benk-mira (dismissed):
Awesome.  I gave some general feedback for now, and I'm sure I'll have some more useful feedback when I start to interact with it while writing the inversion driver.  I'm going to approve, but suggest at least to address the note about missing docs/attrs list for the new classes since it will probably same time down the road.  Nice work!

geoh5py/objects/surveys/direct_current.py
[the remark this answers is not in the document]
benk-mira: Resolved by talking with Dom.  Want to be able to set from both current and potential!

geoh5py/shared/utils.py
domfournier: Yep, bad find/replace

tests/create_survey_dcip_test.py
domfournier: I ll do it for them all at once on a separate issue?

geoh5py/objects/surveys/direct_current.py
domfournier: isn't implied with the base class though?

tests/create_survey_dcip_test.py
domfournier: Did not know about tile. But no unfortunately. Tile would return [0, 1, 0, 1, 0, 1 ...], while the kron returns [0, 0, 0, ..., 1,1,1]. I'm open to a different function ... if you can think of one

geoh5py/objects/surveys/direct_current.py
domfournier: Done

Issue #119, opened 2021-07-29
geoh5py.objects.drillhole.Drillhole.desurvey fails on holes with one survey point


markgrujic:
**Summary:**
Tables in holes with only one survey point (usually at surface) can't be desurveyed

**Example code:**
```
from geoh5py.groups import ContainerGroup
from geoh5py.workspace import Workspace
from geoh5py.objects import Drillhole

workspace = Workspace("zero_depth.geoh5")
group = ContainerGroup.create(workspace, name='myGroup')

# Create a simple well
total_depth = 100
dist = 0 # if one point, it fails
azm = 45
dip = 70
collar = np.r_[0., 10., 10]

well = Drillhole.create(
    workspace, collar=collar, surveys=np.c_[dist, dip, azm]
)

depths = np.r_[5, 10, 20]

# fails:
ds = pd.DataFrame(well.desurvey(depths), columns=["X", "Y", "Z"])
```

**Expected behaviour:**
The hole should be considered "infinitely" straight as defined by the available survey data.

**Actual behaviour:**
desurvey requires >1 point to define a curve

**Code that seems to fix it:**
Add a dummy point far away.
```
if well.surveys.shape[0] == 1:
    well = Drillhole.create(
        workspace, collar=collar,
        surveys=np.vstack([np.c_[dist, dip, azm], np.c_[9999, dip, azm]])
    )
    
ds = pd.DataFrame(well.desurvey(depths), columns=["X", "Y", "Z"])
```

domfournier:
@sebhmg  That might be a bit dangerous for visualization?
There might be a way to extrapolate from a single point on top ...

Issue #118, opened 2021-07-29
geoh5py.objects.drillhole.Drillhole.desurvey doesn't like depth=0


markgrujic:
**Summary:**
When desurveying a hole, depth = 0 introduces errors by using the last survey point instead of the first.

**Example code:**
```
from geoh5py.groups import ContainerGroup
from geoh5py.workspace import Workspace
from geoh5py.objects import Drillhole

import pandas as pd
import numpy as np

workspace = Workspace("zero_depth.geoh5")
group = ContainerGroup.create(workspace, name='myGroup')

# Create a simple well
total_depth = 100
dist = np.linspace(0, total_depth, 10)
azm = np.ones_like(dist) * 45.
dip = np.linspace(-89, -75, dist.shape[0])
collar = np.r_[0., 10., 10]

well = Drillhole.create(
    workspace, collar=collar, surveys=np.c_[dist, dip, azm]
)

depths_good = np.r_[0.01, 5, 10, 20]
depths_bad = np.r_[0, 0.01, 5, 10, 20]

ds_good = pd.DataFrame(well.desurvey(depths_good), columns=["X", "Y", "Z"])
ds_bad = pd.DataFrame(well.desurvey(depths_bad), columns=["X", "Y", "Z"])

print(ds_good)
print(ds_bad)
```
```
          X          Y         Z
0  0.000219  10.000219  9.990006
1  0.109673  10.109673  5.002867
2  0.219346  10.219346  0.005734
3  0.609148  10.609148 -9.977772
          X          Y         Z
0 -7.554879   2.445122  8.151507
1  0.000219  10.000219  9.990006
2  0.109673  10.109673  5.002867
3  0.219346  10.219346  0.005734
4  0.609148  10.609148 -9.977772
```

**Expected behaviour:**
Desurveyed data at depth=0 should use the collar location. If no survey at depth = 0, either an error or projection of the hole to surface should occur.

**Actual behaviour:**
It looks like the last survey point is used, resulting in incorrect desurveying

**Code that seems to fix it:**
If there is a survey location at 0 it's pretty safe to 
```
    def desurvey(self, depths):
        """
        Function to return x, y, z coordinates from depth.
        """
        assert (
            self.surveys is not None and self.collar is not None
        ), "'surveys' and 'collar' attributes required for desurvey operation"
        
        assert (
            np.min(depths) >= 0
        ), "Minimum depth must be >= 0"
        
        if isinstance(depths, list):
[...]

PR #122, opened 2021-08-11
GEOPY-300

domfournier:
Addressing issue #118, #119

### DIFF
--- a/geoh5py/objects/drillhole.py
+++ b/geoh5py/objects/drillhole.py
@@ -40,7 +40,6 @@ class Drillhole(Points):
     __TYPE_UID = uuid.UUID(
         fields=(0x7CAEBF0E, 0xD16E, 0x11E3, 0xBC, 0x69, 0xE4632694AA37)
     )
-
     _attribute_map = Points._attribute_map.copy()
     _attribute_map.update(
         {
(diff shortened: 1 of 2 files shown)


domfournier:
Addressing issue #119

domfournier:
@markgrujic If you want to have a go at it as well... should fix your two issues there.
The idea was to add internally an extra survey point at depth=0 (with same dip azimuth as last) is the surveys contains a single line.

geoh5py/objects/drillhole.py
```diff
@@ -219,19 +213,16 @@ def surveys(self):
...
 
         if getattr(self, "_surveys", None) is not None:
-            return self._surveys
+            surveys = self._surveys.view("<f4").reshape((-1, 3))
+
+            # Repeat first survey point at surface for de-survey interpolation
+            surveys = np.vstack([surveys[0, :], surveys])
+            surveys[0, 0] = 0.0
+
+            return surveys.astype(float)
```
sebhmg, line 225: **question:** is this always adding an extra point, or only if needed?

sebhmg (approved):
I am not sure to understand each of the changes but if tests are passing, that is good

geoh5py/objects/drillhole.py
domfournier: Always yes, just because it doesn hurt, but will always fix the case where the survey point(s) is only defined below the collar. Has to do with the searchsorted needing a point on the "left"

benk-mira (approved):
Nice work!

PR #123, opened 2021-08-24
GEOPY-246: Parent change

domfournier:
**GEOPY-246 - Parent change not reflected in GA tree**

### DIFF
--- a/geoh5py/io/h5_writer.py
+++ b/geoh5py/io/h5_writer.py
@@ -104,6 +104,27 @@ class H5Writer:
         )
 
     @staticmethod
+    def remove_child(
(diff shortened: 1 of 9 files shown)


geoh5py/io/h5_writer.py
```diff
@@ -103,6 +103,27 @@ def create_dataset(cls, entity_handle, dataset: np.array, label: str):
             compression_opts=9,
         )
 
+    @staticmethod
```
benk-mira, line 106: nice docstring and typing!

benk-mira (approved):
Looks good to me.

PR #127, opened 2021-09-01
GEOPY-312: Fix input argument parent=None

domfournier:
**GEOPY-312 - Increase geoh5py coverage**

### DIFF
--- a/tests/create_entity_parent_test.py
+++ b/tests/create_entity_parent_test.py
@@ -1,0 +1,43 @@
+#  Copyright (c) 2021 Mira Geoscience Ltd.
+#
+#  This file is part of geoh5py.
+#
+#  geoh5py is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  geoh5py is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


tests/create_entity_parent_test.py
```diff
@@ -0,0 +1,43 @@
...
+from pathlib import Path
+
+from geoh5py.groups import ContainerGroup
+from geoh5py.objects import Points
+from geoh5py.workspace import Workspace
+
+
+def test_create_point_data():
+
+    with tempfile.TemporaryDirectory() as tempdir:
```
sebhmg, line 28: simpler with the `tmp_path` fixture (thanks to @benk-mira for having discover that one)
https://docs.pytest.org/en/6.2.x/tmpdir.html

tests/create_entity_parent_test.py
```diff
@@ -0,0 +1,43 @@
...
+        h5file_path = Path(tempdir) / r"test.geoh5"
```
sebhmg, line 29: with the fixture, will become
```suggestion
        h5file_path = tmp_path / r"test.geoh5"
```

tests/create_entity_parent_test.py
domfournier: Agreed, we just thought of making an issue and re-vamping them all at once.
Up to you.

sebhmg (approved):

PR #131, opened 2021-09-23
GEOPY-330

domfournier:
**GEOPY-330 - Fix geoh5py DC survey docs**


benk-mira (approved):
user_guide dc survey images looking good now.

PR #135, opened 2021-10-08
GEOPY-346

domfournier:
**GEOPY-346 - Copy direct_current fails**

### DIFF
--- a/geoh5py/objects/surveys/direct_current.py
+++ b/geoh5py/objects/surveys/direct_current.py
@@ -117,6 +117,8 @@ class PotentialElectrode(Curve):
             self, parent, copy_children=copy_children, omit_list=omit_list
         )
         setattr(new_entity, "_ab_cell_id", None)
+        if new_entity.ab_cell_id is None and self.ab_cell_id is not None:
+            self.ab_cell_id.copy(parent=new_entity)
         new_currents = parent.workspace.copy_to_parent(
             self.current_electrodes,
             parent,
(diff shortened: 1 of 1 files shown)


sebhmg (changes requested):
see some suggestions

geoh5py/objects/surveys/direct_current.py
```diff
@@ -118,29 +117,21 @@ def copy(self, parent=None, copy_children: bool = True):
             self, parent, copy_children=copy_children, omit_list=omit_list
         )
         setattr(new_entity, "_ab_cell_id", None)
-
+        if new_entity.ab_cell_id is None and self.ab_cell_id is not None:
```
sebhmg: **question:** given the `setattr` call above, is this test `new_entity.ab_cell_id is None` not always true?

sebhmg: ho, `ab_cell_id()` is maybe getting the ID from somewhere else. Fair enough.

sebhmg (approved):
see my question about setting and testing ab_cell_id, but maybe be I also have the answer ;)

Looks good thanks

geoh5py/objects/surveys/direct_current.py
domfournier: correct yes, the ab_cell_id will go through the property again and try to pull from geoh5. The reason being that if copy_children==True, it will already be available

PR #149, opened 2021-11-18
GEOPY-382

domfournier:
**GEOPY-382 - StatsCache remains when property values are edited through geoh5py**

### DIFF
--- a/geoh5py/io/h5_writer.py
+++ b/geoh5py/io/h5_writer.py
@@ -559,6 +559,10 @@ class H5Writer:
                     compression="gzip",
                     compression_opts=9,
                 )
+                entity_type_handle = H5Writer.fetch_handle(h5file, entity.entity_type)
+                stats_cache = entity_type_handle.get("StatsCache")
+                if stats_cache is not None:
+                    del entity_type_handle["StatsCache"]
 
     @classmethod
     def write_entity(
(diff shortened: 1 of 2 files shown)


geoh5py/io/h5_writer.py
```diff
@@ -559,6 +559,10 @@ def write_data_values(
                     compression="gzip",
                     compression_opts=9,
                 )
+                entity_type_handle = H5Writer.fetch_handle(h5file, entity.entity_type)
```
sebhmg, line 562: **question:** is it per entity type? I was expecting it to be per "data type"

@jincandescent , do you confirm?

sebhmg (commented):
waiting for @jincandescent to comment

geoh5py/io/h5_writer.py
domfournier: here the entity is a "data" entity.

sebhmg: mmh... however the data type is not the `entity_type`, or is it?

sebhmg: data type is how the user regroups data for statistics. Typically done automatically based  on the name of the data.
@jincandescent  correct me if I am wrong

domfournier: Under geoh5py, the data_type is just a sub-class of entity_type, without a fix uuid
image

sebhmg (approved):

PR #152, opened 2021-11-30
GEOPY-90

domfournier:
**GEOPY-90 - Support of geoh5 MT object in geoh5py**


geoh5py/objects/surveys/magnetotellurics.py
```diff
@@ -0,0 +1,229 @@
...
+        :return: List of new Data objects.
+        """
+        data_objects = []
+        if self.channels is None or not self.channels:
+            raise AttributeError(
+                "The 'channels' property defining frequencies must be set before adding data."
+            )
+
+        if not isinstance(data, dict):
+            raise TypeError(
```
benk-mira, line 179: Would be nice to put a more thorough validator here to check the component names, nesting structure, etc.  Maybe once we get the validator module moved over?

benk-mira: Oh, I see, you're doing the extra level validation in the for loop..  Great!

tests/create_survey_mt_test.py
```diff
@@ -0,0 +1,130 @@
...
+        mt_survey = Magnetotellurics.create(workspace, vertices=vertices, name=name)
+
+        for key, value in {
+            "input_type": "Rx Only",
+            "survey_type": "Magnetotellurics",
+            "unit": "Hertz (Hz)",
+        }.items():
+            assert getattr(mt_survey, key) == value, f"Error setting defaults for {key}"
+
+        with pytest.raises(TypeError):
```
benk-mira, line 60: Nice addition!  These can be a little stronger if you check that the output is what you programmed it to be.  Rules out a TypeError popping up in an unexpected place before your code hits:

with pytest.raises(TypeError) as excinfo:
    call_method()

assert "some string from my custom exception" in str(excinfo)

benk-mira (dismissed):
Looks great!  Nothing blocking in my comments.  Address as you see fit.

tests/create_survey_mt_test.py
domfournier: Done

benk-mira (dismissed):
Yep yep!

tests/create_survey_mt_test.py
benk-mira: Sweeeet!

PR #170, opened 2022-02-16
Geopy 347 reorganize

benk-mira:
**GEOPY-347 - ui json parsing and validation in geoh5py**
Allow passing validations and validators into InputValidation class.  Move validations from ui_json into InputValidations class.

### DIFF
--- a/geoh5py/ui_json/input_file.py
+++ b/geoh5py/ui_json/input_file.py
@@ -11,14 +11,13 @@ import json
 import os
 import warnings
 from copy import deepcopy
-from typing import Any, Callable, cast
+from typing import Any, Callable
 from uuid import UUID
 
 import numpy as np
 
-from geoh5py.groups import ContainerGroup, PropertyGroup
+from geoh5py.groups import ContainerGroup
 from geoh5py.io.utils import as_str_if_uuid, entity2uuid, str2uuid, uuid2entity
-from geoh5py.shared import Entity
 from geoh5py.shared.exceptions import BaseValidationError, JSONParameterValidationError
 from geoh5py.shared.validators import UUIDValidator
 from geoh5py.workspace import Workspace
--- a/geoh5py/ui_json/input_file.py
+++ b/geoh5py/ui_json/input_file.py
@@ -19,7 +19,7 @@ import numpy as np
 from geoh5py.groups import ContainerGroup, PropertyGroup
 from geoh5py.io.utils import as_str_if_uuid, entity2uuid, str2uuid, uuid2entity
 from geoh5py.shared import Entity
-from geoh5py.shared.exceptions import JSONParameterValidationError
+from geoh5py.shared.exceptions import BaseValidationError, JSONParameterValidationError
 from geoh5py.shared.validators import UUIDValidator
 from geoh5py.workspace import Workspace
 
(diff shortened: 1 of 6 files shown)


geoh5py/ui_json/input_file.py
```diff
@@ -166,7 +165,9 @@ def validations(self, valid_dict: dict | None):
     def validators(self):
```
domfournier, line 165: I would propose to rename this to `input_validation` and `ui_validation` instead of `validators` and `ui_validators`. Just so we make a distincation with `validators` of the `InputValidation` class.

geoh5py/ui_json/validation.py
```diff
@@ -51,17 +46,54 @@ class InputValidation:
...
+    def validations(self, val: dict[str, Any]):
+        self._validations = val
+
+    @property
+    def validators(self):
+        return self._validators
+
+    @validators.setter
+    def validators(self, val):
+        v = {} if val is None else val
```
domfournier, line 89: We should probably check for `isinstance(val, "BaseValidator")`, otherwise the merge with required_validators won't make much sense.

geoh5py/ui_json/validation.py
```diff
@@ -51,17 +46,54 @@ class InputValidation:
...
+        self.ui_json: dict[str, Any] = ui_json
+        self.validators: dict[str, BaseValidator] = validators
         self.workspace: Workspace | None = workspace
-        self.validations = validations
         self.ignore_list: tuple = ignore_list
         self.ignore_requirements: bool = ignore_requirements
 
+    @property
+    def ui_json(self):
+        return self._ui_json
```
domfournier, line 65: Propose to NOT store ui_json, but just call the `infer_validations` method directly in the init. Just scared of having the ui_json lingering in many classes...

geoh5py/ui_json/validation.py
```diff
@@ -72,38 +104,64 @@ def workspace(self, value: Workspace | None):
...
                 else:
-                    raise ValueError(f"No validator implemented for argument '{key}'.")
-        elif val is None:
-            self._validators = None
-        else:
-            raise ValueError(
-                "Input 'validations' must be of of type 'dict' or None. "
-                f"Argument of type {type(val)} provided"
-            )
+                    check_type = cast(Any, type(item["value"]))
```
domfournier, line 159: What's the use case for this?

domfournier (changes requested):
Nice, I like the changes. Getting there

geoh5py/ui_json/validation.py
benk-mira: I don't know, it's your line of code that I moved over from input file class.

geoh5py/ui_json/validation.py
benk-mira: ok, I've eliminated the setting and added a raise attribute error on the getter.

geoh5py/ui_json/input_file.py
benk-mira: Let's talk through the naming, I have a different opinion..

geoh5py/ui_json/validation.py
benk-mira: val isn't expected to be a BaseValidator here, but a dictionary of validators.  How about typing this with `val: dict[str, BaseValidator]`

domfournier: Oh yep, sure. That makes more sense.

geoh5py/ui_json/validation.py
domfournier: ahha ok. Must be a reason...

geoh5py/ui_json/validation.py
benk-mira: Done

domfournier (approved):

PR #171, opened 2022-02-23
Geopy 463

benk-mira:
**GEOPY-463 - Handle None values in Type/Value/UUID validators**
Handle None values in validators


domfournier (commented):
Need to chat about this one

domfournier (dismissed):
Grand

PR #172, opened 2022-02-24
Geopy 464

benk-mira:
**GEOPY-464 - UUID validation only checking for valid uuid string when run from geoapps where workspace or parent isn't known ahead of time**
Simplify UUIDValidator to check for valid UUID string only


geoh5py/ui_json/constants.py
```diff
@@ -37,42 +38,42 @@
...
-    "dependency": {"types": [str]},
-    "dependencyType": {"types": [str], "values": ["enabled", "disabled"]},
-    "enabled": {"types": [bool]},
-    "group": {"types": [str]},
-    "label": {"types": [str], "required": True},
-    "main": {"types": [bool]},
-    "meshType": {"types": [str, UUID]},
-    "optional": {"types": [bool]},
-    "parent": {"types": [str, UUID]},
+    "dataType": {"types": [str, type(None)], "values": ["Float", "Text", "Integer"]},
```
domfournier, line 51: That's not allowed for ui.json, those fields have to be one of the `values`

geoh5py/ui_json/validation.py
```diff
@@ -52,10 +52,8 @@ def __init__(
         ui_json: dict[str, Any] = None,
         **validation_options,
     ):
-        self._inferred_validations = (
-            {} if ui_json is None else self.infer_validations(ui_json)
-        )
-        self.validations: dict[str, Any] | None = validations
+        self.validations = {} if ui_json is None else self.infer_validations(ui_json)
+        self.validations = self._merge_validations(validations)
```
domfournier, line 56: Not a big fan on this syntax, sort of repeating the same assignment twice in a row. And this would brake if validations is None since not protected in _merge_validations

geoh5py/ui_json/validation.py
```diff
@@ -116,16 +115,21 @@ def _required_validators(validations):
...
         for key, item in ui_json.items():
             if not isinstance(item, dict):
                 continue
 
             if "isValue" in item:
                 validations[key] = {
-                    "types": [UUID, int, float, Entity],
-                    "association": item["parent"],
+                    "types": [str, UUID, int, float, Entity],
+                    "association": None if item["isValue"] else item["parent"],
```
domfournier, line 128: Don't think that's possible. "isValue" means that this can be a dropdown for data, which always requires an association...

domfournier (changes requested):
We need to talk about this.

geoh5py/ui_json/constants.py
benk-mira: you're right, there's no need for a types validation if there's also a values validation.  Removed the former.

geoh5py/ui_json/validation.py
benk-mira: Updated this section as we discussed: passing validations into infer_validations and handling internally

geoh5py/ui_json/validation.py
benk-mira: Yep, changed to only attach association is not isVAlue instead of setting to None

PR #173, opened 2022-02-24
GEOPY-470

domfournier:
**GEOPY-470 - Crash on geoh5 without Types**

### DIFF
--- a/geoh5py/io/h5_reader.py
+++ b/geoh5py/io/h5_reader.py
@@ -87,6 +87,9 @@ class H5Reader:
             for key, value in entity.attrs.items():
                 attributes["entity"][key] = value
 
+            if "Type" not in entity:
+                entity.create_group("Type")
+
             for key, value in entity["Type"].attrs.items():
                 type_attributes["entity_type"][key] = value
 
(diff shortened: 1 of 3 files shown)


geoh5py/io/h5_reader.py
```diff
@@ -87,6 +87,9 @@ def fetch_attributes(
             for key, value in entity.attrs.items():
                 attributes["entity"][key] = value
 
+            if "Type" not in entity:
```
benk-mira, line 90: In the writer, you are creating a group called 'Types', but here it is just 'Type'.  Is this correct?

benk-mira: hmm.. Based on the tests, I'm guessing that this is correct, but I'll leave the comment in here so you can explain to me what's going on!!

domfournier: That's just geoh5 naming convention. "Types" is a group containing the "Type" of all objects... not my choice

benk-mira: gotcha

benk-mira (approved):
All good

PR #177, opened 2022-02-25
GEOPY-278

domfournier:
**GEOPY-278 - Colormap ignored in kwargs on DataType creation**
# Conflicts:
#	tests/color_map_test.py


geoh5py/data/color_map.py
```diff
@@ -50,27 +61,39 @@ def values(self) -> np.ndarray:
...
-            assert values.dtype.names is not None and all(
-                name in names for name in (values.dtype.names)
-            ), f"Input 'values' must contain fields with types {names}"
-            self._values = np.asarray(values, dtype=list(zip(names, formats)))
+        if not isinstance(values, np.ndarray):
+            raise TypeError(f"Input 'values' of ColorMap must be of type {np.ndarray}.")
+
+        if np.issubdtype(values.dtype.base, np.number):
+            if values.shape[1] != 5:
+                raise ShapeValidationError("values", values.shape, "(*, 5)")
```
benk-mira, line 76: This is great for now, but we should think about how we would incorporate the larger validation framework in here.  I think this might point to having a lightweight constructor to support validations per param.  That way we could construct a validation here with a dictionary containing the types and shape rules and then call the .validate method to accomplish these two if condition: validate patterns.

geoh5py/data/data_type.py
```diff
@@ -88,10 +88,23 @@ def color_map(self) -> ColorMap | None:
...
+
+        if isinstance(color_map, dict):
+            color_map = ColorMap(**color_map)
+
+        elif isinstance(color_map, np.ndarray):
+            color_map = ColorMap(values=color_map)
+
+        elif not isinstance(color_map, ColorMap):
+            raise TypeError(
+                f"Input value for 'color_map' must be of type {ColorMap} or dict with 'values'."
```
benk-mira, line 101: ... or a numpy array.

geoh5py/data/color_map.py
benk-mira: I'll add that the use case here should be incorporated into the shape validator, ie: pass in one or more free dimensions and enforce the rest.

geoh5py/data/data_type.py
domfournier: Nice catch

benk-mira (dismissed):
Great, thanks

PR #179, opened 2022-03-07
Geopy 466

benk-mira:
**GEOPY-466 - InputValidation class needs to properly handle merging inferred validations and external ones.**
Add unit test to show merging behaviour of external validations into InputFile class


domfournier (approved):
Looks all good to me.

PR #185, opened 2022-03-30
GEOPY-67

domfournier:


geoh5py/objects/geo_image.py
```diff
@@ -33,13 +41,204 @@ class GeoImage(ObjectBase):
...
+            locations.ndim != 2
+            or reference.shape[0] != locations.shape[0]
+            or locations.shape[1] != 3
+        ):
+            raise ValueError(
+                "Input 'locations' must be a 2D array of shape(*, 3) "
+                "with the same number of rows as the control points."
+            )
+        param_x, _, _, _ = np.linalg.lstsq(
+            np.c_[np.ones(3), reference], locations[:, 0], rcond=None
```
benk-mira, line 197: should this be np.ones(len(reference))?

benk-mira (commented):
Have a look at the comment in the georeference method - I think I caught a bug there.  Otherwise looks great!

geoh5py/objects/geo_image.py
domfournier: Very nice catch brother. That's what I am talking about!

domfournier: Fixed

benk-mira (approved):
Sweet!

PR #190, opened 2022-04-11
Geopy 388

benk-mira:
**GEOPY-388 - Add a validation type to make sure at least one channel of data and uncertainty are provided to inversion.**
Needs to go in before the GEOPY-388 on geoapps


domfournier (approved):
Greeeat

PR #195, opened 2022-04-19
Pillow update

domfournier:


domfournier:
@andrewg-mira: Do we need an 0.2.0rc2, or just release straight with 0.2.0?

andrewg-mira (approved):

andrewg-mira:
I think that if the checks pass OK we can just release 0.2.0 without doing an rc2

PR #200, opened 2022-05-13
GEOPY-514

benk-mira:
**GEOPY-514 - The AtLeastOneValidation doesn't work for the natural sources**


domfournier (approved):
Seems reasonable!!

domfournier:
I added a unit test to replicate the error I see on my side.
We basically need to ignore the validation on dependent variables if part of a group. Otherwise we get the wrong error message. 

image

PR #201, opened 2022-05-18
Change logic on dependency variables

domfournier:
**GEOPY-514 - The AtLeastOneValidation doesn't work for the natural sources**


benk-mira (approved):
I think this will work in our current state.  Could be a problem if we ever write code relying on the params setter_validators, but we'll cross that bridge when we get there..

PR #202, opened 2022-05-19
Geopy 508

domfournier:


geoh5py/workspace/workspace.py
```diff
@@ -827,6 +827,13 @@ def objects(self) -> list[objects.ObjectBase]:
         return self._all_objects()
 
     def open(self, mode: str | None = None):
+        if isinstance(self._geoh5, h5py.File) and self._geoh5:
```
sebhmg, line 830: **suggestion:** no need to test `self._geoh5`
if `isinstance(self._geoh5, h5py.File)` is true, then `self._geoh5` is not None.

geoh5py/workspace/workspace.py
```diff
@@ -827,6 +827,13 @@ def objects(self) -> list[objects.ObjectBase]:
...
+            warnings.warn(
+                f"Workspace already opened in mode {self._geoh5.mode}. "
+                "Consider closing the file first with 'close()'."
```
sebhmg, line 833: **suggestion:** not sure this second message line is necessary.
If the file is already open, there is no need to close it.
Unless the file is open as read-only and it tries to open read/write. But in that case, I assume there would be an error already.

sebhmg (changes requested):
please see suggestions/questions

sebhmg (approved):

geoh5py/workspace/workspace.py
domfournier: Indeed!

geoh5py/workspace/workspace.py
domfournier: That's the case I had in mind yes. I am fine with removing the hint.

sebhmg (approved):

PR #207, opened 2022-06-06
GEOPY-543: always import annotations from future

sebhmg:
**GEOPY-543 - make sure typehints do no affect runtime and do not require imports**
and makes sure typehint are fully ignored at runtime
and not forcing imports


domfournier (approved):
Sure thing

PR #208, opened 2022-06-07
GEOPY-539

jamiefb:
**GEOPY-539 - Create dash interface to Scatter plot**


domfournier (commented):
Looking good!
We will need to add a unitest for it. You can just append to the `create_workspace_test.py` file. We basically need to re-read an existing one as a byte string, and make sure that the Workspace builds the same as from a path.

domfournier (approved):
Fantastic work. Going in.

PR #221, opened 2022-07-09
GEOPY-587: fix new pylint reports in geoh5py

sebhmg:
**GEOPY-587 - fix new pylint reports in geoh5py**


domfournier (approved):
Nice cleanup!

PR #231, opened 2022-08-10
Geopy 329

domfournier:
**GEOPY-329 - Implement drape models**


geoh5py/objects/drape_model.py
```diff
@@ -0,0 +1,157 @@
...
+    @property
+    def layers(self) -> np.ndarray | None:
+        """
+        :obj:`~geoh5py.objects.object_base.ObjectBase.layers`
+        """
+        if self._layers is None and self.on_file:
+            self._layers = self.workspace.fetch_array_attribute(self, "layers")
+
+        if self._layers is not None:
+            return np.asarray(self._layers.tolist())
```
benk-mira, line 101: why convert to list then back to array?

benk-mira: oh... is the original self._layers a recarray which
 becomes an ndarray?

benk-mira (approved):
Looks good, just one small remark, but nothing blocking.

PR #252, opened 2022-10-13
GEOPY-644: Add clipping by extent method

domfournier:
**GEOPY-644 - Add clipping by extent method**

### DIFF
--- a/geoh5py/groups/group.py
+++ b/geoh5py/groups/group.py
@@ -75,9 +75,7 @@ class Group(Entity):
         else:
             self.comments.values = self.comments.values + [comment_dict]
 
-    def clip_by_extent(
-        self, bounds: np.ndarray, parent=None, copy_children: bool = True
-    ) -> Group | None:
+    def clip_by_extent(self, bounds: np.ndarray, parent=None) -> Group | None:
         """
         Find indices of vertices within a rectangular bounds.
 
@@ -86,7 +84,15 @@ class Group(Entity):
             with shape(2, 3) defining the top and bottom limits.
         :param attributes: Dictionary of attributes to clip by extent.
         """
-        new_entity = self.copy(parent=parent, copy_children=copy_children)
+        new_entity = self.copy(parent=parent, copy_children=False)
+        for child in self.children:
+            child.clip_by_extent(bounds, parent=new_entity, copy_children=True)
+
+        if len(new_entity.children) == 0:
+            new_entity.workspace.remove_entity(new_entity)
+            del new_entity
+            return None
+
         return new_entity
 
     @property
--- a/geoh5py/groups/group.py
+++ b/geoh5py/groups/group.py
@@ -75,7 +75,9 @@ class Group(Entity):
         else:
             self.comments.values = self.comments.values + [comment_dict]
 
-    def clip_by_extent(self, bounds: np.ndarray, parent=None) -> Group | None:
+    def clip_by_extent(
+        self, bounds: np.ndarray, parent=None, copy_children: bool = True
+    ) -> Group | None:
         """
         Find indices of vertices within a rectangular bounds.
 
@@ -86,7 +88,7 @@ class Group(Entity):
         """
         new_entity = self.copy(parent=parent, copy_children=False)
         for child in self.children:
-            child.clip_by_extent(bounds, parent=new_entity, copy_children=True)
+            child.clip_by_extent(bounds, parent=new_entity, copy_children=copy_children)
 
         if len(new_entity.children) == 0:
             new_entity.workspace.remove_entity(new_entity)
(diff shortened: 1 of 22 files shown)


geoh5py/groups/group.py
```diff
@@ -73,6 +75,28 @@ def add_comment(self, comment: str, author: str = None):
...
+    ) -> Group | None:
+        """
+        Find indices of vertices within a rectangular bounds.
+
+        :param bounds: shape(2, 2) Bounding box defined by the South-West and
+            North-East coordinates. Extents can also be provided as 3D coordinates
+            with shape(2, 3) defining the top and bottom limits.
+        :param attributes: Dictionary of attributes to clip by extent.
+        """
+        new_entity = self.copy(parent=parent, copy_children=False)
```
sebhmg, line 89: **suggestion (test):** add tests for the cases pointed below by codecov

geoh5py/objects/curve.py
```diff
@@ -26,17 +26,74 @@
...
+    Base class for object with cells.
+    """
+
+    _attribute_map = Points._attribute_map.copy()
+
+    def __init__(self, object_type: ObjectType, **kwargs):
+
+        self._cells: np.ndarray | None = None
+
+        super().__init__(object_type, **kwargs)
```
sebhmg, line 40: **suggestion:** usually, calling parent constructor comes first

geoh5py/objects/curve.py
```diff
@@ -26,17 +26,74 @@
 from .points import Points
 
 
-class Curve(Points):
+class CellObject(Points):
```
sebhmg, line 29: **suggestion:** move class to a dedicated file, as it will be shared by other object types

geoh5py/objects/object_base.py
```diff
@@ -211,13 +213,39 @@ def comments(self):
     def default_type_uid(cls) -> uuid.UUID:
         ...
 
+    def clip_by_extent(
+        self, bounds: np.ndarray, parent=None, copy_children: bool = True
+    ) -> ObjectBase | None:
+        """
+        Find indices of vertices within a rectangular bounds.
```
sebhmg, line 220: **suggestion (maintainability):** make it clear it must be re-implemented
* either add to the doc that default implementation does not clip and returns a copy of the object as-is
* or annotate the method as abstract here, then sub-classes can throw a non-implemented exception

geoh5py/shared/utils.py
```diff
@@ -317,7 +316,42 @@ def dict_mapper(
...
+def mask_by_extent(locations: np.ndarray, extent: np.ndarray) -> np.ndarray:
+    """
+    Find indices of locations within a rectangular extent.
+
+    :param locations: shape(*, 3) Coordinates to be evaluated.
+    :param extent: shape(2, 2) Limits defined by the South-West and
+        North-East corners. Extents can also be provided as 3D coordinates
+        with shape(2, 3) defining the top and bottom limits.
+    """
+    if isinstance(extent, list):
```
sebhmg, line 328: **suggestion (test):** add tests for the cases pointed below by codecov

sebhmg (changes requested):
Nice ! ideally, have some more tests

Please also see comment about default implementation of the clip function

geoh5py/objects/object_base.py
domfournier: Sounds good. I might turn it into abstract actually, and just implement the copy for anything that is not vertex based?

domfournier: Sounds good. I might turn it into abstract actually, and just implement the copy for anything that is not vertex based?

geoh5py/groups/group.py
domfournier: Done

geoh5py/objects/curve.py
domfournier: Ok maybe, but the base `Entity` applies the kwargs, so `_cells` would be lost.

geoh5py/objects/curve.py
domfournier: Done

geoh5py/objects/object_base.py
domfournier: Added a warning for all non-vertex entities for now.

geoh5py/objects/object_base.py
```diff
@@ -224,8 +226,30 @@ def clip_by_extent(
...
+        """
+        Find indices of cells within a rectangular bounds.
+
+        :param bounds: shape(2, 2) Bounding box defined by the South-West and
+            North-East coordinates. Extents can also be provided as 3D coordinates
+            with shape(2, 3) defining the top and bottom limits.
+        :param attributes: Dictionary of attributes to clip by extent.
+        """
+
+        # TODO Clip entity within bounds.
```
sebhmg, line 247: **chore:** is there a JIRA issue to track this?

sebhmg (approved):
👍 🏆

geoh5py/objects/object_base.py
domfournier: Riding with the devil!!! GEOPY-666

PR #267, opened 2022-11-15
Geopy 700

MatthieuCMira:
**GEOPY-700 - Add GeoImage conversion method to Grid2D**
Change GeoImage object:
- if the image is a tiff, save the coordinates in the vertices automatically
- add a function .to_grid2d() to the object in order to convert it to a geoh5py GRID2D object


domfournier (commented):
Thanks for the work. See notes below. We will also need to add a unittest to cover the new code. We can go over it together.

geoh5py/objects/geo_image.py
```diff
@@ -158,6 +166,10 @@ def image(self, image: str | np.ndarray | BytesIO | Image.Image):
 
         self.vertices = self.default_vertices
 
+        # if the image is a tiff, georeference the image
+        if "image_copy" in locals():
```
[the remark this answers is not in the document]
MatthieuCMira: Yes, but the self.default_vertices depends on the self.image_data
Then the georeferencing has to be at the end.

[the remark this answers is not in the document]
domfournier: Ok i am not 100% sure why the header would not be accessible by the object. Technically it is stored as a blob copy of the original. Feels like the `georeferencing_from_tiff` should more `georeferencing_from_header` and not take any argument - just pulls it from it's image stored on file.

geoh5py/objects/geo_image.py
```diff
@@ -238,3 +252,134 @@ def vertices(self, xyz: np.ndarray | list):
...
+            warn("The 'tif.' image has no referencing information.")
+
+    def to_grid2d(
+        self,
+        name: str = None,
+        rotation: float = 0.0,
+        dip: float = 0.0,
+        elevation: float = 0.0,
+        transform: str = "GRAY",
+        parent: Group = None,
```
domfournier, line 298: As discussed, let's pass `**grid2d_kwargs` from here to the `.create(**grid2d_kwargs)` instead and remove `name`, `rotation`, `dip` and `parent` from this list.

geoh5py/objects/geo_image.py
```diff
@@ -238,3 +252,134 @@ def vertices(self, xyz: np.ndarray | list):
...
+    ) -> Grid2D:
```
domfournier, line 299: Great job on the return type.

domfournier (changes requested):

geoh5py/objects/geo_image.py
```diff
@@ -238,3 +250,127 @@ def vertices(self, xyz: np.ndarray | list):
...
+        grid = Grid2D.create(
+            self.workspace,
+            name=name,
+            origin=[u_origin, v_origin, elevation],
+            u_cell_size=u_cell_size,
+            v_cell_size=v_cell_size,
+            u_count=u_count,
+            v_count=v_count,
+            rotation=float(rotation),
+            dip=float(dip),
```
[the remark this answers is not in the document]
MatthieuCMira: parent will  be passed in the **kwargs

geoh5py/objects/geo_image.py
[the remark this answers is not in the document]
MatthieuCMira: In the original code, the vertices are reset after the image is loaded. 
If I want to set it before, then I have to change this part. 
Seems easier to just have the georeferencing line calling locals at the end than change the structure.

geoh5py/objects/geo_image.py
```diff
@@ -379,3 +379,54 @@ def to_grid2d(
...
+            for id_ in self.tag:
+                image.getexif()[id_] = self.tag[id_]
+
+            return image
+
+        return None
+
+    def export_as_georeferenced_tif(
+        self, name: str, path: str = ""
+    ):  # todo: export as a normal tiff if no tag?
```
domfournier, line 399: Can we generalize this and just call `export`? We could check if the outgoing name is `tif`, then if `tag` exists we try to save as geotif, otherwise plain image with PIL?

geoh5py/objects/geo_image.py
```diff
@@ -238,3 +280,181 @@ def vertices(self, xyz: np.ndarray | list):
...
+                        [u_origin, v_origin, 0.0],
+                        [u_oposite, v_origin, 0.0],
+                        [u_oposite, v_oposite, 0.0],
+                    ]
+                )
+
+                # georeference the raster
+                self.georeference(reference, locations)
+
+        except (AttributeError, KeyError):
```
domfournier, line 339: Can remove the `AttributeError` from this `try`, and just be more specific with logic for `if self.tag is None`?

geoh5py/objects/geo_image.py
```diff
@@ -238,3 +280,181 @@ def vertices(self, xyz: np.ndarray | list):
...
+    def to_grid2d(
+        self,
+        transform: str = "GRAY",
+        **grid2d_kwargs,
+    ) -> Grid2D:
+        """
+        Create a geoh5py :obj:geoh5py.objects.grid2d.Grid2D from the geoimage in the same workspace.
+        :param transform: the type of transform ; if "GRAY" convert the image to grayscale ;
+        if "RGB" every band is sent to a data of a grid.
+        :param grid2d_kwargs: **kwargs arguments of Grid2D: name, parent, rotation, dip, elevation.
```
domfournier, line 351:
```suggestion
        :param **grid2d_kwargs: Any argument supported by :obj:`geoh5py.objects.grid2d.Grid2D`.
```

geoh5py/objects/geo_image.py
```diff
@@ -238,3 +280,181 @@ def vertices(self, xyz: np.ndarray | list):
...
+                image.getexif()[id_] = self.tag[id_]
+
+            return image
+
+        return None
+
+    def save_as(self, name: str, path: str = ""):
+        """
+        Function to save the geoimage into a file.
+        It the name ends by '.tif' or '.tiff' and the tag is not None
```
domfournier, line 436: Missing some comment here?

geoh5py/objects/geo_image.py
```diff
@@ -238,3 +280,181 @@ def vertices(self, xyz: np.ndarray | list):
...
+        """
+
+        # verifications
+        if self.image is None:
+            raise AttributeError("The object contains no image data")
+        if not isinstance(name, str):
+            raise TypeError(
+                f"The 'name' has to be a string; a '{type(name)}' was entered instead"
+            )
+        if not isinstance(path, str):
```
domfournier, line 448: Should probably allow for type `Path` as well

geoh5py/objects/geo_image.py
domfournier: Just so we can hyperlink to Grid2D

domfournier (approved):
To test!
image

domfournier (approved):
Worked great!

PR #284, opened 2023-01-18
Geopy 721 v df

MatthieuCMira:
**GEOPY-721 - Refactor grid2d to geoimage into object converter class**


geoh5py/shared/conversion/base.py
```diff
@@ -0,0 +1,169 @@
...
+        """Output object"""
+        return self._output
+
+    @output.setter
+    def output(self, value):
+        """
+        Set the output object.
+        :param value: any values to be pass to output.
+        """
+        self._output = value
```
domfournier, line 101: I don't think we want to necessarily store one output. Remember that the converter can technically output to many type of objects. So we could call the same converter (for a given object) multiple times.

geoh5py/shared/conversion/base.py
```diff
@@ -0,0 +1,169 @@
...
+
+    @property
+    def name(self) -> str | None:
+        """Name of the output object"""
+        return self._name
```
domfournier, line 106: Same here, not sure we need this since the output can change.

geoh5py/shared/conversion/base.py
```diff
@@ -0,0 +1,130 @@
...
+        # get the properties
+        properties = cls.verify_kwargs(input_entity, **kwargs)
+
+        # create the point object
+        output = objects.Points.create(
+            properties["workspace"], vertices=input_entity.centroids, **kwargs
+        )
+
+        # copy the properties of the original object
+        cls.copy_properties(input_entity, output)
```
domfournier, line 125: The order of operation won't work here because the copy would over-write the user input values in kwargs. I would suggest moving this inside the `verify_kwargs` as part of the kwargs prepping phase.

geoh5py/shared/conversion/geo_image.py
```diff
@@ -0,0 +1,238 @@
...
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with geoh5py.  If not, see <https://www.gnu.org/licenses/>.
+#
+#  This file is part of geoh5py.
+#
+#  geoh5py is free software: you can redistribute it and/or modify
```
domfournier, line 20: Copyright duplicated.

geoh5py/shared/conversion/grid2d.py
```diff
@@ -0,0 +1,244 @@
...
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with geoh5py.  If not, see <https://www.gnu.org/licenses/>.
+#
+#  This file is part of geoh5py.
+#
+#  geoh5py is free software: you can redistribute it and/or modify
```
domfournier, line 20: Here too, duplicate copyright

geoh5py/shared/utils.py
```diff
@@ -367,3 +367,15 @@ def get_attributes(entity, omit_list=(), attributes=None):
...
+
+
+def overwrite_kwargs(to_overwrite: dict, kwargs: dict) -> dict:
+    """
+    Overwrite kwargs with overwrite.
+    :param to_overwrite: Dictionary of kwargs to overwrite.
+    :param kwargs: Dictionary of kwargs.
+    """
+    for key, value in kwargs.items():
+        if key in kwargs:
```
domfournier, line 379:
```suggestion
        if key in to_overwrite:
```

geoh5py/objects/geo_image.py
```diff
@@ -222,6 +226,7 @@ def georeference(self, reference: np.ndarray | list, locations: np.ndarray | lis
 
         self.set_tag_from_vertices()
 
+    @typing.no_type_check
```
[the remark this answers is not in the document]
MatthieuCMira: Just to not being annoyed by mypy when adding data in the dictionary.

geoh5py/shared/conversion/base.py
```diff
@@ -0,0 +1,136 @@
...
+        return workspace
+
+    @classmethod
+    def copy_properties(cls, input_entity: Entity, output: Entity, **kwargs):
+        """
+        Copy all the properties from the original entity to the new one.
+        :param input_entity: The input entity to convert.
+        :param output: the new entity.
+        """
+        for property_ in entity_properties:
```
domfournier, line 81: `entity_properties` isn't defined? This can't be working?

domfournier (approved):
image

PR #288, opened 2023-01-26
GEOPY-843

domfournier:
**GEOPY-843 - Drillhole group update to 4.2 changes**

### DIFF
--- a/geoh5py/io/h5_reader.py
+++ b/geoh5py/io/h5_reader.py
@@ -230,6 +230,14 @@ class H5Reader:
 
                     return json.loads(as_str_if_utf8_bytes(attribute))
 
+                if label == "Attributes Jsons":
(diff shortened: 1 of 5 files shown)


geoh5py/io/h5_reader.py
```diff
@@ -230,6 +230,14 @@ def fetch_concatenated_attributes(
 
                     return json.loads(as_str_if_utf8_bytes(attribute))
 
+                if label == "Attributes Jsons":
+                    attribute = group[label][()]
```
benk-mira, line 234: what does [()] do?

benk-mira (approved):

geoh5py/io/h5_reader.py
domfournier: That's an h5py syntax to get an array out

Issue #275, opened 2022-12-08
Drillhole.get_data() doesn't return values


cardinalgeo:
Hi everyone! I'm enjoying using geoh5py so far, though I've found a potential bug (described below). Any idea what's going on? 
Best, Robert Collar

### Environment data
OS (Windows | Mac | Linux distro) and version: Darwin arm64 21.3.0
Type of virtual environment used (N/A | venv | virtualenv | conda | ...): conda
python version:    3.9.13
geoh5py version: 0.4.0

### Expected behavior
When I create a `Drillhole` object using, for example, `well = Drillhole.create(workspace)`, add attributes (e.g., `well.name = "Drillhole"`) and interval data, close the workspace, and then reopen the workspace (e.g., in a different jupyter notebook), I expect the interval data values to be retrievable. 

### Actual behavior
Instead, when I follow the steps above, the interval data values are `None`. Interestingly, if I initialize a `Drillhole` object and its attributes all at once (and then subsequently add the interval data like before), the interval data values are retrievable. 

### Steps to reproduce
In the example below, after closing and reopening the workspaces, the `interval_data` values are only printed for the first well. 
```
from geoh5py.workspace import Workspace
from geoh5py.groups import DrillholeGroup
from geoh5py.objects import Drillhole
import numpy as np

import pandas as pd

workspace1 = Workspace("../bug_mre1.geoh5")
workspace2 = Workspace("../bug_mre2.geoh5")

dh_group1 = DrillholeGroup.create(workspace1)
dh_group2 = DrillholeGroup.create(workspace2)

# Create a simple well
total_depth = 100
dist = np.linspace(0, total_depth, 10)
azm = np.ones_like(dist) * 45.
dip = np.linspace(-89, -75, dist.shape[0])
collar = np.r_[0., 10., 10]

# data intervals
from_to = np.vstack([
    [0.25, 25.5],
    [30.1, 55.5],
    [56.5, 80.2]
])

# create Drillhole object for first well
well1 = Drillhole.create(
    workspace1, collar=collar, surveys=np.c_[dist, azm, dip], name="Drillhole", parent=dh_group1
)

well1.add_data({
    "random_values": {
        "values": np.array([1, 1, 1],dtype="float64"),
        "from-to": from_to,
    }
})

# create Drillhole object for second well
well2 = Drillhole.create(workspace2)

well2.name = "Drillhole"
well2.parent = dh_group2
well2.collar=collar
[...]

PR #302, opened 2023-02-15
GEOPY-723

domfournier:
**GEOPY-723 - Drillhole.get_data() doesn't return values**
Deals with issue raised on #275


benk-mira (approved):
hmm.. that looked familiar.  Guess this one should go in before 667

PR #304, opened 2023-02-17
GEOPY-857

domfournier:
**GEOPY-857 - clip by extent fails with memory error on large file**
@sebhmg I added an argument to clear attributes on copy. Should solve your issue... unless you can't load the original object with its data. Otherwise will need to slice at the source.

### DIFF
--- a/geoh5py/groups/group.py
+++ b/geoh5py/groups/group.py
@@ -75,7 +75,11 @@ class Group(Entity):
             self.comments.values = self.comments.values + [comment_dict]
 
     def copy_from_extent(
-        self, bounds: np.ndarray, parent=None, copy_children: bool = True
+        self,
+        bounds: np.ndarray,
+        parent=None,
+        copy_children: bool = True,
+        clear_cache: bool = False,
     ) -> Group | None:
         """
         Find indices of vertices within a rectangular bounds.
@@ -88,7 +92,10 @@ class Group(Entity):
         new_entity = self.copy(parent=parent, copy_children=False)
         for child in self.children:
             child.copy_from_extent(
-                bounds, parent=new_entity, copy_children=copy_children
+                bounds,
+                parent=new_entity,
+                copy_children=copy_children,
+                clear_cache=clear_cache,
             )
 
         if len(new_entity.children) == 0:
--- a/geoh5py/objects/object_base.py
+++ b/geoh5py/objects/object_base.py
@@ -239,11 +239,12 @@ class ObjectBase(Entity):
         ):
             return None
 
-        new_entity = self.copy(parent=parent, copy_children=copy_children)
+        new_entity = self.copy(
+            parent=parent, copy_children=copy_children, clear_cache=clear_cache
+        )
         new_entity.clip_by_extent(bounds)
 
         if clear_cache:
-            clear_array_attributes(self, recursive=copy_children)
             clear_array_attributes(new_entity, recursive=copy_children)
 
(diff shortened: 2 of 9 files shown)


tests/clip_vertex_data_test.py
```diff
@@ -136,10 +136,15 @@ def test_clip_groups(tmp_path):
                 },
             }
         )
+
         with Workspace(tmp_path / r"testClipPoints_copy.geoh5") as new_workspace:
-            group_a.copy_from_extent(extent, parent=new_workspace)
+            group_a.copy_from_extent(extent, parent=new_workspace, clear_cache=True)
```
sebhmg, line 141: **suggestion (tests):** have a test test with `clear_cache=False`

and also tests for `Workspace.copy_to_parent()` and  implementation `Entity.copy()` with `clear_cache=True` and `clear_cache=False`

geoh5py/groups/group.py
```diff
@@ -75,7 +75,11 @@ def add_comment(self, comment: str, author: str | None = None):
             self.comments.values = self.comments.values + [comment_dict]
 
     def copy_from_extent(
-        self, bounds: np.ndarray, parent=None, copy_children: bool = True
+        self,
+        bounds: np.ndarray,
+        parent=None,
+        copy_children: bool = True,
+        clear_cache: bool = False,
```
sebhmg, line 82: **issue (doc):** obsolete doc of parameters

The 3 last parameters are not documented, while the doc mentions non existing parameter `attributes`

geoh5py/objects/object_base.py
```diff
@@ -220,7 +220,11 @@ def default_type_uid(cls) -> uuid.UUID:
         ...
 
     def copy_from_extent(
-        self, bounds: np.ndarray, parent=None, copy_children: bool = True
+        self,
+        bounds: np.ndarray,
+        parent=None,
+        copy_children: bool = True,
+        clear_cache: bool = False,
```
sebhmg, line 227: **issue (doc):** obsolete doc of parameters

The 3 last parameters are not documented, while the doc mentions non existing parameter `attributes`

geoh5py/objects/surveys/direct_current.py
```diff
@@ -103,13 +103,20 @@ def ab_map(self) -> dict | None:
...
+        clear_cache: bool = False,
+        **kwargs,
+    ):
         """
         Function to copy a survey to a different parent entity.
 
         :param parent: Target parent to copy the entity under. Copied to current
             :obj:`~geoh5py.shared.entity.Entity.parent` if None.
         :param copy_children: Create copies of all children entities along with it.
+        :param clear_cache: Clear array attributes after copy.
```
sebhmg, line 119: **suggestion:** remove ambiguity in the doc: does it clear attributes on `self` or on the `copy`

Also should the doc says clears the attributes _cache_ array?

I do not quite understand where attributes are stored if cleared in memory and the file has not been saved to disk yet,

sebhmg (changes requested):
thanks !
please see some suggestions and questions

sebhmg:
I ma testing with on a 3 GB geoh5, and it eats 11 GB right away, before being killed with exit code 137

image

domfournier:
> I am testing with on a 3 GB geoh5, and it eats 11 GB right away, before being killed with exit code 137
> 
> image

So I guess you can't even load the object with it's data in memory. Will have to slice one data at a time.

PR #307, opened 2023-02-28
GEOPY-857

domfournier:
**GEOPY-857 - clip by extent fails with memory error on large file**


geoh5py/objects/cell_object.py
```diff
@@ -89,16 +107,66 @@ def remove_vertices(self, indices: list[int]):
...
         self.remove_cells(np.where(~np.all(vert_index[self.cells], axis=1)))
         setattr(self, "cells", new_index[self.cells])
+
+    def copy(
+        self,
+        parent=None,
+        copy_children: bool = True,
+        clear_cache: bool = False,
+        mask: list[float] | np.ndarray | None = None,
+        cell_mask: list[float] | np.ndarray | None = None,
```
sebhmg, line 125: **thought (maintainability, performances, for-later):** I suppose `mask` and `cell_mask` always have to be consistent consistent together.
This could be enforced by having `class Mask`, with properties `cell_indices` and `vertex_indices`.

Possibly even subclassing?
- `class VertexMask` with `vertex_indices` only
- `class CellMask` with both `vertex_indices` and `cell_indices`

BTW, also consider using bitarray. The `Mask` class would hide the actual implementation, so that the choice of regular array or bitarray can be revisited later without touching the client code.

`bitarray` will beat `list[int]` in storage efficiency as soon as more than 1/32 of vertices are masked.
On object with n vertices, masking k vertices:
- bitarray takes n bits in memory
- list[int] takes 32*k bits in memory

Masking all or none would take special handling to not store a full bitarray of 0 or 1. Something like
```py
class Mask:
   def __init__(self):
       self.vertex_mask : bitarray | None  = None  # None equivalent to all 0
       self.vertex_mask_all : bool  = false  # if true, equivalent ot bitarray with all 1
```

geoh5py/objects/surveys/direct_current.py
```diff
@@ -213,6 +272,64 @@ def metadata(self, values):
...
 
+    @property
+    @abstractmethod
+    def potential_electrodes(self):
+        """
+        The associated potential_electrodes (receivers)
+        """
+
+
+class PotentialElectrode(BaseElectrode):
```
sebhmg, line 283: *question (out-of-scope):** from another PR? out of scope for this PR

geoh5py/objects/surveys/electromagnetics/ground_tem.py
```diff
@@ -181,44 +334,44 @@ def default_receiver_type(self):
...
-            if values.parent is not None and values.parent.uid != self.uid:
-                values = values.copy(parent=self)
-
-        else:
-            raise TypeError(
-                "Input value for 'tx_id_property' should be a np.ndarray or ReferenceData.)"
-            )
-
-        self._tx_id_property = values
+    # @property
```
sebhmg, line 337: **suggestion (clean-code):** remove code commented out

geoh5py/shared/utils.py
```diff
@@ -161,6 +161,22 @@ def merge_arrays(
     return np.r_[head, tail]
 
 
+def clear_array_attributes(entity: Entity, recursive: bool = False):
+    """
+    Clear all attributes from an entity.
```
sebhmg, line 166: **suggestion (doc)**: clarify this is only a cache and that it does not affect the actual attribute values of the entity

geoh5py/shared/utils.py
```diff
@@ -407,17 +423,24 @@ def get_attributes(entity, omit_list=(), attributes=None):
...
 
-            attributes[key] = getattr(entity, key)
+            attr = getattr(entity, key)
+            attributes[key] = attr
+
     return attributes
 
 
-def overwrite_kwargs(to_overwrite: dict, kwargs_to_add: dict) -> dict:
+def xy_rotation_matrix(angle: float) -> np.ndarray:
```
sebhmg, line 432: **question:** out of scope? is this part of GEOPY-857

geoh5py/workspace/workspace.py
```diff
@@ -15,7 +15,7 @@
 #  You should have received a copy of the GNU Lesser General Public License
 #  along with geoh5py.  If not, see <https://www.gnu.org/licenses/>.
 
-# pylint: disable=R0904
+# pylint: disable=R0904, R0913, R0914
 # pylint: disable=C0302
```
sebhmg, line 19: **suggestion:** disable pylint statement with short names instead of numbers
```suggestion
# pylint: disable=too-many-lines, too-many-public-methods, too-many-arguments, too-many-locals
```

sebhmg (changes requested):
- Would probably take a dedicated Mask class. See comment on https://github.com/MiraGeoscience/geoh5py/pull/307/files#diff-c3f7c9d137c766a1a5079e37d8748c400450ffcd6b0bec07ce19728580870cccR125
Such a refactoring can be addressed in another dedicated issue.

- See some comments on the doc for the new functions (return type hint, explanation on when or why to use `clear_cache`). Could be documented once only in base class, with a reference  from inherited classes, instead of copying the doc to each function function doc. See suggestion of PEP- to use "extends" or "overrides" when documenting functions in inherited classes ( [PEP-257](
https://peps.python.org/pep-0257/#:~:text=If%20a%20class,its%20own%20behavior) ).

- Some questions regarding cache invalidation for `_extent`

geoh5py/shared/entity.py
```diff
@@ -168,26 +171,14 @@ def clipping_ids(self) -> list[uuid.UUID] | None:
         """
         return self._clipping_ids
 
-    def copy(self, parent=None, copy_children: bool = True, **kwargs):
+    @abstractmethod
+    def mask_by_extent(self, extent: list[float]):
```
sebhmg, line 175: **issue (type-hint, inconsistency):** all implementations seem to have type `np.ndarray`

geoh5py/groups/group.py
```diff
@@ -74,30 +74,92 @@ def add_comment(self, comment: str, author: str | None = None):
         else:
             self.comments.values = self.comments.values + [comment_dict]
 
-    def copy_from_extent(
-        self, bounds: np.ndarray, parent=None, copy_children: bool = True
-    ) -> Group | None:
+    def mask_by_extent(self, extent: list[float]):
```
sebhmg, line 77: **issue (type-hint, inconsistency):** all other implementations have type `np.ndarray`

geoh5py/objects/grid2d.py
```diff
@@ -141,6 +140,90 @@ def centroids(self) -> np.ndarray | None:
...
+        :param parent: Target parent to copy the entity under. Copied to current
+            :obj:`~geoh5py.shared.entity.Entity.parent` if None.
+        :param copy_children: (Optional) Create copies of all children entities along with it.
+        :param clear_cache: Clear array attributes after copy.
+        :param kwargs: Additional keyword arguments to pass to the copy constructor.
+
+        :return entity: Registered Entity to the workspace.
+        """
+        if isinstance(extent, list):
+            extent = np.vstack(extent)
```
sebhmg, line 164: **issue:** does not convert flat list `list[float]` to a 2 dimension array

maybe not need to have the list type for this param...

```
>>> import numpy
>>> numpy.vstack([0,1,2,3,4,5])
array([[0],
       [1],
       [2],
       [3],
       [4],
       [5]])
```

geoh5py/shared/entity.py
```diff
@@ -210,6 +201,58 @@ def create(cls, workspace, **kwargs):
...
+        :param copy_children: (Optional) Create copies of all children entities along with it.
+        :param clear_cache: Clear array attributes after copy.
+        :param kwargs: Additional keyword arguments to pass to the copy constructor.
+
+        :return entity: Registered Entity to the workspace.
+        """
+
+    def copy_from_extent(
+        self,
+        extent: list[float] | ndarray,
```
sebhmg, line 226: **suggestion (YAGNI):** not necessary to support list as param type

else, would be `list[list[float]]` and requires its own set of unit tests...

geoh5py/objects/object_base.py
```diff
@@ -214,47 +212,84 @@ def comments(self):
 
         return None
 
+    def copy(
+        self,
+        parent=None,
+        copy_children: bool = True,
+        clear_cache: bool = False,
+        mask: list[float] | np.ndarray | None = None,
```
sebhmg, line 220: **question:** if the mask is a list of vertex indices, why does it contain float instead of integer?

sebhmg (changes requested):
please see some more questions and possible issues where using type list instead of ndarray

geoh5py/groups/group.py
```diff
@@ -74,30 +74,92 @@ def add_comment(self, comment: str, author: str | None = None):
...
         """
-        Find indices of vertices within a rectangular bounds.
+        Mask data by extent.
 
-        :param bounds: shape(2, 2) Bounding box defined by the South-West and
-            North-East coordinates. Extents can also be provided as 3D coordinates
-            with shape(2, 3) defining the top and bottom limits.
-        :param attributes: Dictionary of attributes to clip by extent.
+        :param extent: [xmin, ymin, xmax, ymax]
```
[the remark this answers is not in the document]
domfournier: It does, I will beefup the docs

geoh5py/objects/surveys/direct_current.py
domfournier: Allowed me to reduce code duplication for the copy by creating a sub-class. It's definitely a bit of a creep, but overall less lines to test.

geoh5py/objects/grid2d.py
domfournier: Was mainly for convenience of supplying [pt_1, pt_2], instead of forcing to be a numpy array... I just think I need to change the type hint to list[list]

geoh5py/shared/entity.py
domfournier: Ok fine. We can force 2D array, will be less risky

geoh5py/objects/object_base.py
domfournier: Should be `int`

sebhmg:
testing the application, I am getting this error with 2D grids:
image

domfournier:
clearly a typo there, but what's the error?

sebhmg:
> clearly a typo there, but what's the error? Need to see the end of the log.

E       ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 2 dimensions. The detected shape was (3, 3) + inhomogeneous part.

domfournier:
Ok but where does it error. Show me the actual log

sebhmg:
> Ok but where does it error. Show me the full traceback, or send me your file.

I will share my code...

I have tracked down the origin of the problem to the creation of the 2D grid, here: _rotation is an array. If I understand right `astype` convert the array values, not the array itself
image

Note the same pattern is used for _cell_size and in octree:
image

sebhmg:
> Wow ok, that's a new one. Rotation usually comes in as a numpy array shape(1,) in float (as checked by the setter). We will have to look at the geoh5, and figure out how/where it was created, and protect against that I guess.

So, it appear the problem does not happen with numpy 1.23.5 (as in geoapps locked env) or 1.21.6 (geoh5py poetry.lock), but does with numpy 1.24.1
My guess is that there used to be type coercion when creating a nd.array with a mix of scalars and 1d arrays as values.

Fixes in https://github.com/MiraGeoscience/geoh5py/pull/309 works for all these versions.
For safety, I would suggest to constraint numpy to ~1.23 (that is 1.23.*)

geoh5py/objects/grid2d.py
sebhmg: possibly, as long as tests are covering both the numpy and the list input

benk-mira (approved):
Looks good.  Just one comment, more of a question

geoh5py/groups/group.py
```diff
@@ -74,30 +74,92 @@ def add_comment(self, comment: str, author: str | None = None):
...
+        return None
+
+    def copy(
+        self,
+        parent=None,
+        copy_children: bool = True,
+        clear_cache: bool = False,
+        mask: list[float] | np.ndarray | None = None,
+        **kwargs,
+    ):
```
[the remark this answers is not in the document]
domfournier: Done

geoh5py/data/data.py
```diff
@@ -55,6 +58,63 @@ def __init__(
...
+        )
+
+        return new_entity
+
+    @property
+    def extent(self) -> np.ndarray | None:
+        """
+        Geography bounding box of the parent object.
+
+        :return: shape(2, 3) Bounding box defined by the bottom South-West and
```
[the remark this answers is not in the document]
domfournier: No, it will always be (2, 3) since objects have 3 coordinates.

geoh5py/objects/cell_object.py
domfournier: Will implement at the same time as GEOPY-776

geoh5py/objects/drillhole.py
```diff
@@ -115,7 +115,7 @@ def collar(self, value: list | np.ndarray | None):
             self.workspace.update_attribute(self, "attributes")
 
         self._locations = None
-
+        self._extent = None
```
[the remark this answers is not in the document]
domfournier: Getting rid of the stashed attribute

geoh5py/data/data.py
```diff
@@ -55,6 +58,63 @@ def __init__(
...
+        mask: np.ndarray | None = None,
+        **kwargs,
+    ):
+        """
+        Function to copy data to a different parent entity.
+
+        :param parent: Target parent to copy the entity under. Copied to current
+            :obj:`~geoh5py.shared.entity.Entity.parent` if None.
+        :param copy_children: (Optional) Create copies of all children entities along with it.
+        :param clear_cache: Clear array attributes after copy.
```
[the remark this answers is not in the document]
domfournier: It would if you are going to re-use the object - otherwise you need to re-load

geoh5py/objects/grid_object.py
```diff
@@ -0,0 +1,162 @@
...
+    @property
+    def extent(self):
+        """
+        Geography bounding box of the object.
+
+        :return: shape(2, 3) Bounding box defined by the bottom South-West and
+            top North-East coordinates.
+        """
+        if self._extent is None and self.centroids is not None:
+            self._extent = np.c_[
```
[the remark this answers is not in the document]
domfournier: Will be removed

geoh5py/objects/cell_object.py
domfournier: Will be tackled at the same time as GEOPY-776

geoh5py/shared/entity.py
```diff
@@ -210,6 +201,58 @@ def create(cls, workspace, **kwargs):
         )
         return new_object
 
+    @abstractmethod
+    def copy(
+        self,
+        parent=None,
+        copy_children: bool = True,
+        clear_cache: bool = False,
```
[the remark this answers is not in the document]
domfournier: Done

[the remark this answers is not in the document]
domfournier: Done

geoh5py/objects/surveys/electromagnetics/ground_tem.py
domfournier: Done

geoh5py/shared/utils.py
domfournier: Done

sebhmg (dismissed):

geoh5py/objects/object_base.py
```diff
@@ -247,14 +245,24 @@ def copy(
...
+                ):
+                    # if child.name in ["A-B Cell ID", "Transmitter ID"]:
+                    #     continue
+                    #
+                    # child_mask = mask
+                    # if child.association is DataAssociationEnum.CELL:
+                    #     if self.cells is not None and mask is not None:
+                    #         child_mask = np.all(mask[self.cells], axis=1)
+                    # elif child.association is not DataAssociationEnum.VERTEX:
+                    #     child_mask = None
```
sebhmg, line 260: **suggestion (clean-code):** remove code commented out?

sebhmg (approved):

geoh5py/objects/object_base.py
domfournier: Done. Thanks for that.

sebhmg (dismissed):

domfournier:
Re-tested with

```
def test_file(tmp_path):
    extent = np.vstack([[313600, 6071500], [315500, 6072400]])

    with Workspace(r"geoapps\assets\FlinFlon_natural_sources.geoh5") as ws:
        with Workspace(tmp_path / r"test_clipping.geoh5") as new_ws:
            for obj in ws.objects:
                obj.copy_from_extent(extent, parent=new_ws)

    with Workspace(r"geoapps\assets\FlinFlon_dcip.geoh5") as ws:
        with Workspace(tmp_path / r"test_clipping_dcip.geoh5") as new_ws:
            for obj in ws.objects:
                obj.copy_from_extent(extent, parent=new_ws)

    with Workspace(r"geoapps\assets\FlinFlon.geoh5") as ws:
        with Workspace(tmp_path / r"test_clipping_all.geoh5") as new_ws:
            for obj in ws.objects:
                obj.copy_from_extent(extent, parent=new_ws)
```

PR #311, opened 2023-03-15
GEOPY-877

domfournier:
**GEOPY-877 - Improve intersection detection in `mask_by_extent`**

### DIFF
--- a/geoh5py/shared/utils.py
+++ b/geoh5py/shared/utils.py
@@ -382,6 +382,34 @@ def dict_mapper(
     return val
 
 
+def box_intersect(
+    extent_a: np.ndarray, extent_b: np.ndarray, return_array=False
+) -> bool | np.ndarray:
+    """
+    Compute the intersection of two axis-aligned bounding extents defined by their
+    bottom south-west corner and top north-east corner.
+
+    :param extent_a: First extent
+    :param extent_b: Second extent
+    :param return_array: Return the intersection extent
+
+    :return: True if the box extents intersect,
+        or intersection extent coordinates if `return_array` is True.
+    """
+    min_ext = []
+    max_ext = []
+    for comp_a, comp_b in zip(extent_a.T, extent_b.T):
+        min_ext.append(max(min(comp_a[0], comp_a[1]), min(comp_b[0], comp_b[1])))
+        max_ext.append(min(max(comp_a[0], comp_a[1]), max(comp_b[0], comp_b[1])))
+
+    overlap = np.array([min_ext, max_ext])
+
+    if return_array:
+        return overlap
+
+    return not (np.diff(overlap, axis=0) < 0).any()
+
+
 def mask_by_extent(locations: np.ndarray, extent: np.ndarray) -> np.ndarray:
     """
     Find indices of locations within a rectangular extent.
--- a/geoh5py/shared/utils.py
+++ b/geoh5py/shared/utils.py
@@ -403,10 +403,10 @@ def box_intersect(extent_a: np.ndarray, extent_b: np.ndarray) -> bool:
             )
 
     for comp_a, comp_b in zip(extent_a.T, extent_b.T):
-        min_ext = max(min(comp_a[0], comp_a[1]), min(comp_b[0], comp_b[1]))
-        max_ext = min(max(comp_a[0], comp_a[1]), max(comp_b[0], comp_b[1]))
+        min_ext = max(comp_a[0], comp_b[0])
+        max_ext = min(comp_a[1], comp_b[1])
 
-        if (max_ext - min_ext) < 0:
+        if min_ext > max_ext:
             return False
 
     return True
--- a/geoh5py/shared/utils.py
+++ b/geoh5py/shared/utils.py
@@ -382,23 +382,19 @@ def dict_mapper(
     return val
 
 
-def box_intersect(
-    extent_a: np.ndarray, extent_b: np.ndarray, return_array=False
-) -> bool | np.ndarray:
+def box_intersect(extent_a: np.ndarray, extent_b: np.ndarray) -> bool:
     """
     Compute the intersection of two axis-aligned bounding extents defined by their
-    arrays of minimum and maximum bounds in ND-space.
+    arrays of minimum and maximum bounds in N-D space.
 
     :param extent_a: First extent or shape (2, N)
     :param extent_b: Second extent or shape (2, N)
-    :param return_array: Return the intersection extent
 
-    :return: True if the box extents intersect, or if `return_array` is True
-        returns the intersection extent array with same shape as the input extents.
+    :return: Logic if the box extents intersect along all dimensions.
     """
     for extent in [extent_a, extent_b]:
         if not isinstance(extent, np.ndarray) or extent.ndim != 2:
-            raise TypeError("Input extents must be a 2D numpy.ndarray.")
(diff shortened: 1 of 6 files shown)


geoh5py/shared/utils.py
```diff
@@ -382,6 +382,34 @@ def dict_mapper(
...
+    min_ext = []
+    max_ext = []
+    for comp_a, comp_b in zip(extent_a.T, extent_b.T):
+        min_ext.append(max(min(comp_a[0], comp_a[1]), min(comp_b[0], comp_b[1])))
+        max_ext.append(min(max(comp_a[0], comp_a[1]), max(comp_b[0], comp_b[1])))
+
+    overlap = np.array([min_ext, max_ext])
+
+    if return_array:
+        return overlap
```
sebhmg, line 408: **issue:** seem to return an non null extent if input extents are disjoined

tests/shared_utils_test.py
```diff
@@ -56,3 +61,36 @@ def test_mask_by_extent():
     assert not mask_by_extent(
         np.vstack([points]), np.vstack(corners[:2])
     ), "Point should have been outside."
+
+
+def test_box_intersect():
```
sebhmg, line 66: **suggestion (readability, tests):** better split the function in dedicated tests

- test_disjoint
- test_fully_inside
- test_corner_inside (4 cases in 2D, 8 in 3D)
- test_edge_inside (4 cases in 2D, 12 in 3D)
- test_face inside (6 cases in 3D)

also test the symmetric each time (`box_intersect(a,b)` and `box_intersect(b,a)`)

for test with multiple cases, use something like
```py
@pytest.mark.parameterize("box_a, box_b, expected_box", boxes_for_corner_intersect)
def test_corner_intersect(box_a, box_b, expected_box)
```

geoh5py/shared/utils.py
```diff
@@ -382,6 +382,34 @@ def dict_mapper(
...
+    :param extent_a: First extent
+    :param extent_b: Second extent
+    :param return_array: Return the intersection extent
+
+    :return: True if the box extents intersect,
+        or intersection extent coordinates if `return_array` is True.
+    """
+    min_ext = []
+    max_ext = []
+    for comp_a, comp_b in zip(extent_a.T, extent_b.T):
```
sebhmg, line 401: make it complciate becaseu there is no axumtion that it point are ordered (min_x, min_y) (max_x, max_y)
Could just assert if it is no the case. Would also make less test cases to write...

Extent should always be correctly ordered when computed. If user provided, it is the responsibility of the entry point, in upstream code, to fixup the box or raise an exception

sebhmg (changes requested):
seems to return an non null extent if boxes are disjoint.
See suggestion for test cases, and simplification

geoh5py/shared/utils.py
```diff
@@ -382,6 +382,34 @@ def dict_mapper(
...
 
 
+def box_intersect(
+    extent_a: np.ndarray, extent_b: np.ndarray, return_array=False
+) -> bool | np.ndarray:
+    """
+    Compute the intersection of two axis-aligned bounding extents defined by their
+    bottom south-west corner and top north-east corner.
+
+    :param extent_a: First extent
```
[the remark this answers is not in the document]
domfournier: The `zip()` will stop at the lowest common number. So if one is (2,2) and the other (2,3), then only the x and y will be checked. Makes sense?

tests/shared_utils_test.py
domfournier: Very nice. Will do.

geoh5py/shared/utils.py
```diff
@@ -382,6 +382,34 @@ def dict_mapper(
...
+    :param extent_b: Second extent
+    :param return_array: Return the intersection extent
+
+    :return: True if the box extents intersect,
```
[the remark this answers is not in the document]
domfournier: Done.

geoh5py/shared/utils.py
domfournier: Correct, will add a check on order.

geoh5py/shared/utils.py
```diff
@@ -382,6 +382,34 @@ def dict_mapper(
...
+def box_intersect(
+    extent_a: np.ndarray, extent_b: np.ndarray, return_array=False
+) -> bool | np.ndarray:
+    """
+    Compute the intersection of two axis-aligned bounding extents defined by their
+    bottom south-west corner and top north-east corner.
+
+    :param extent_a: First extent
+    :param extent_b: Second extent
+    :param return_array: Return the intersection extent
```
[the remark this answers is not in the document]
sebhmg: totally. There is not need to compute the box. The algorithm can return as soon as the test fails in one dimension
(Testing if a vertex or a cell is in a big or small bounding box does not change anything in term of performance).

geoh5py/shared/utils.py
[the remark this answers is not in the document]
sebhmg: that is correct if just testing the intersection, but returns the wrong bounding box if computing the resulting intersection box.

geoh5py/shared/utils.py
```diff
@@ -382,32 +382,34 @@ def dict_mapper(
...
+        min_ext = max(min(comp_a[0], comp_a[1]), min(comp_b[0], comp_b[1]))
+        max_ext = min(max(comp_a[0], comp_a[1]), max(comp_b[0], comp_b[1]))
 
-    if return_array:
-        return overlap
+        if (max_ext - min_ext) < 0:
+            return False
 
-    return not (np.diff(overlap, axis=0) < 0).any()
+    return True
```
sebhmg: **suggestion (non-blocking):** no need to resort to zip and translate. Can also simplify min/max  knowing that points are sorted
```py
    for i in range(min(extent.shape[0], extent.shape[1])):
        min_ext = max(comp_a[0][i], comp_b[0][i])
        max_ext = min(comp_a[1][i], comp_b[1][i])
        if min_ext > max_ext:
            return False
    return True
```

sebhmg (approved):

geoh5py/shared/utils.py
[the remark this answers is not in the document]
sebhmg: not relevant since not returning the box anymore, but just to explain my comment:
with the code in place, computing the intersection between a 2D box and a 3D box would have created a 2D box, while it must preserve the limits on z and thus returns a 3D box

geoh5py/shared/utils.py
domfournier: I ll go half way. I think the zip is more elegant but granted that yours is more efficient.
Pushing.

geoh5py/shared/utils.py
```diff
@@ -382,32 +382,34 @@ def dict_mapper(
...
+            )
 
-    overlap = np.array([min_ext, max_ext])
+    for comp_a, comp_b in zip(extent_a.T, extent_b.T):
+        min_ext = max(comp_a[0], comp_b[0])
+        max_ext = min(comp_a[1], comp_b[1])
 
-    if return_array:
-        return overlap
+        if min_ext > max_ext:
```
sebhmg, line 409: **question:** add a tolerence?
```suggestion
        if min_ext > max_ext or np.isclose(min_ext, max_ext):
```

sebhmg (approved):

PR #312, opened 2023-03-15
GEOPY-876: validate geoh5 format when using 4.2 save as to 4.1

domfournier:
**GEOPY-876 - Validate geoh5 format when using 4.2 save as to 4.1**

### DIFF
--- a/geoh5py/shared/concatenation.py
+++ b/geoh5py/shared/concatenation.py
@@ -19,7 +19,6 @@
 
 from __future__ import annotations
 
-import re
 import uuid
 import warnings
 from typing import TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


domfournier:
@sebhmg @jincandescent 
Need to confirm that ga.version == 2.1 is reversed for the new drillhole format using the array "Attributes Jsons", and 2.0 stays with the long "Attributes" json string.

geoh5py/shared/concatenation.py
```diff
@@ -137,13 +136,8 @@ def concat_attr_str(self) -> str:
...
         if self._concat_attr_str is None:
             self._concat_attr_str = "Attributes"
-            if (
-                self.workspace.ga_version is not None
-                and "4." in self.workspace.ga_version
-            ):
-                version = re.findall(r"\d+\.\d+", self.workspace.ga_version)
-                if version and float(version[0]) >= 4.2:
-                    self._concat_attr_str = "Attributes Jsons"
+            if self.workspace.version is not None and self.workspace.version > 2.0:
```
sebhmg, line 139: **question:** @jincandescent, are there case where the version of Analyst matters to interpret the content of geoh5 ?

sebhmg (approved):

sebhmg:
🤔**thought:** some of the questions raised during the code review would be solved by introducing a BoundingBox class.

- **allow for various input types:** BoundingBox can be created from a np.ndarray, or 2 tuples or 2 arrays. No need to carry on the Union type in subsequent fonctions. Then it stores and exposes as np.ndarray (or whatever is the simpler / most efficient)
- **checking whether 2D or 3D boxes and combining them:** BoundingBox is always  3D. If proving only x,y, then z is considered -inf, +inf. The code work seamlessly whether initial input  was 2D or 3D
- **checking the coordinates order**: upon creation, BoundingBox orders the points so that the min point has all its component smaller than the max point. No need to test for proper ordering down the road.

geoh5py/shared/concatenation.py
domfournier: Confirmed with Julien

PR #326, opened 2023-04-03
GEOPY-652: Input file: Handle multi select options uijson

domfournier:
**GEOPY-652 - Input file: Handle multi select options uijson**


benk-mira (approved):
Merge this sucka!

PR #335, opened 2023-04-18
GEOPY-846: Support visual parameters

MatthieuCMira:
**GEOPY-846 - Support visual parameters**
add visual parameter as a data type and set color management


geoh5py/data/visual_parameters.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def check_child(self, child: str, attribute: str) -> bool:
+        """
+        Check if the child is a valid parameter.
+        :param child: the key of the child to check.
+        :param attribute: the attribute to check.
+        :return: True if the child and attribute are valid.
+        """
+        if child not in PARAMETERS:
+            warn(f"{child} is not a valid parameter.")
```
domfournier, line 104:
```suggestion
            warn(f"{child} is not a supported parameter. Must be one of {PARAMETERS}.")
```

geoh5py/data/visual_parameters.py
```diff
@@ -0,0 +1,146 @@
...
+            return False
+
+        if not isinstance(self.xml, ET.Element):
+            warn(f"XML is not set for {self}.")
+            return False
+
+        if self.xml.find(child) is None:
+            warn(f"{child} is not present in the XML.")
```
domfournier, line 112: Maybe we should allow adding new ones if they are supported? For instance what would happen if Visual Parameter is not set initially?

geoh5py/data/visual_parameters.py
```diff
@@ -0,0 +1,146 @@
...
+from warnings import warn
+
+from .text_data import TextData
+
+PARAMETERS = ["Colour"]
+ATTRIBUTES = ["tag", "text", "attrib"]
+
+
+class VisualParameters(TextData):
+    _xml: ET.Element | None = None
```
domfournier, line 45: Should we provide a default xml for convenience?

domfournier (commented):
Very well written brother. See comments below.

geoh5py/data/visual_parameters.py
MatthieuCMira: I don't think we should allow users to pass anything. 
Visual paramter is an already created object. 

We could have a function create_visual_parameters(**kwargs) somewhere to set up visual parameters if they are not created? 

However I agree we should allow the user to define an autorized parameter if it's not in the XML, I'll work on that.

geoh5py/data/visual_parameters.py
MatthieuCMira: Why not, it's could answer your previous comment too.

geoh5py/data/visual_parameters.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @property
+    def colour(self) -> None | str | dict:
+        """
+        :obj:`dict` the Colour dictionary.
+        """
+        return self.get_child("Colour", "text")
+
+    @colour.setter
+    def colour(self, values: str):
```
MatthieuCMira, line 77: I want to test the format of the colour to set (and same for other parameters we will add).
So no one could corrupt the xml file using those functions. 
Where could I get this information?

MatthieuCMira (commented):
test the value of the parameters

geoh5py/data/visual_parameters.py
domfournier: @jincandescent Can you give us list or pattern for what's allowed by GA?

geoh5py/data/visual_parameters.py
```diff
@@ -0,0 +1,141 @@
...
+
+        value = int.from_bytes(bytes.fromhex(byte_string), "little")
+
+        self.set_tag("Colour", str(value))
+
+    def get_tag(self, tag: str) -> None | ET.Element:
+        """
+        Check if the tag is a valid parameter.
+        :param tag: the key of the tag to check.
+        :return: True if the tag is a valid parameter.
```
sebhmg, line 117: **issue (doc):** wrong doc? Does not check the validity of the tag but returns its value.

geoh5py/objects/object_base.py
```diff
@@ -525,3 +526,31 @@ def validate_data_type(attribute_dict):
...
+                        "association": "OBJECT",
+                    },
+                    "entity_type": {"name": "XmlData", "primitive_type": "TEXT"},
+                },
+            )
+
+        return self._visual_parameters
+
+    @visual_parameters.setter
+    def visual_parameters(self, value):
```
sebhmg, line 552: **suggestion (type-hint):** add value type hint

geoh5py/data/visual_parameters.py
```diff
@@ -0,0 +1,132 @@
...
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with geoh5py.  If not, see <https://www.gnu.org/licenses/>.
+
+
+from __future__ import annotations
+
+import xml.etree.ElementTree as ET
```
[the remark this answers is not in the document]
sebhmg: @sonatype-lift ignore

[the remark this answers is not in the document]
sebhmg: @sonatype-lift ignore

geoh5py/data/visual_parameters.py
```diff
@@ -0,0 +1,141 @@
...
+        self.workspace.update_attribute(self, "values")
+
+    @property
+    def colour(self) -> None | list:
+        """
+        Colour of the object in [Alpha, Red, Green, Blue] format.
+
+        Each value is an integer between 0 and 255.
+        The colour value is stored as a single integer converted from
+        a byte string of the form 'BBGGRR00' where 'BB' is the blue value,
```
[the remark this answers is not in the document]
domfournier: hmm no, alpha basically increases the "gray" of the color.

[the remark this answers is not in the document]
sebhmg: @jincandescent, @mathewpmira , how is the "alpha" component supposed to affect the color? Should it be used, or is transparency handled through another parameter ?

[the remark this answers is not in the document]
mathewpmira: Alpha is used primarily to show and hide NDVs in ANALYST, in conjunction with the show/hide no-data values from the data colours panel. It shouldn't increase how close the colour is to dark grey. There are a couple of exceptions where we use them for filtering, but they're specific and rare. In general, the behaviour should be:

- If NDVs are shown, then they will appear as grey.
- If NDVs are hidden, then the value is hidden using the alpha value.
- Opacity (eg: 1 - Transparency) is handled at an object-level, using a shader uniform.

See this (slightly modified) sample section of code from one of our shaders.
```
uniform bool are_nodata_values_transparent;
...
bool is_nodata_value = in_colour.a < 0.5;
if((is_nodata_value && are_nodata_values_transparent) || opacity == 0)
{
    discard;
}
vec4 adjusted_color;
if(is_nodata)
{
    adjusted_color = vec4(0.35,0.35,0.35,1);
}
else
{
    adjusted_color = in_colour;
}
...
```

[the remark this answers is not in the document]
domfournier: After testing it looks like 0 to 127 makes the object transparent, and 128 to 255 makes it appear. So not fully supported.

sebhmg (approved):

PR #367, opened 2023-07-05
Geopy 695

MatthieuCMira:
**GEOPY-695 - Expose Coordinate Reference System as property of objects**
Add a crs property setter and getter in entity.
Test implemented.

### DIFF
--- a/geoh5py/shared/entity.py
+++ b/geoh5py/shared/entity.py
@@ -34,6 +34,8 @@ if TYPE_CHECKING:
     from .. import shared
     from ..workspace import Workspace
 
+DEFAULT_CRS = {"Code": "Unknown", "Name": "Unknown"}
+
 
 class Entity(ABC):
     """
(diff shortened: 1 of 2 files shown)


geoh5py/shared/entity.py
```diff
@@ -212,6 +214,53 @@ def create(cls, workspace, **kwargs):
         )
         return new_object
 
+    @property
+    def coordinate_reference_system(self) -> dict | None:
+        """
+        Coordinate reference system attached to the entity.
+        """
+        coordinate_reference_system = None
```
domfournier, line 222:
```suggestion
        coordinate_reference_system = DEFAULT_CRS
```

geoh5py/shared/entity.py
```diff
@@ -212,6 +214,53 @@ def create(cls, workspace, **kwargs):
...
+                "Coordinate Reference System", None
+            )
+
+            if coordinate_reference_system is not None:
+                coordinate_reference_system = coordinate_reference_system.get(
+                    "Current", None
+                )
+
+        if coordinate_reference_system is None:
+            coordinate_reference_system = DEFAULT_CRS
```
domfournier, line 235:
```suggestion
        if self._metadata is not None and "Coordinate Reference System" in self._metadata:
           coordinate_reference_system = self._metadata["Coordinate Reference System"]["Current"]
```

geoh5py/shared/entity.py
```diff
@@ -212,6 +214,53 @@ def create(cls, workspace, **kwargs):
...
+
+        return coordinate_reference_system
+
+    @coordinate_reference_system.setter
+    def coordinate_reference_system(self, value: dict):
+        # assert value is a dictionary containing "Code" and "Name" keys
+        if not isinstance(value, dict):
+            raise TypeError("Input coordinate reference system must be a dictionary")
+        if value.keys() != {"Code", "Name"}:
```
domfournier, line 244:
```suggestion
        
        if value.keys() != {"Code", "Name"}:
```

tests/coordinate_system_test.py
```diff
@@ -0,0 +1,87 @@
...
+    277: (3,),
+    278: (5,),
+    284: (1,),
+    34737: ("WGS 84 / UTM zone 34N|WGS 84|",),
+}
+
+DEFAULT_CRS = {"Code": "Unknown", "Name": "Unknown"}
+
+
+def test_georeference_image(tmp_path):
```
domfournier, line 46: I don't think we need to be specific about image here. Any entity would do.

tests/coordinate_system_test.py
```diff
@@ -0,0 +1,87 @@
...
+
+    grid2 = grid2d_rgb.copy()
+
+    # set reference coordinate system
+    coordinate_system = {
+        "Code": "bidon",
+        "Name": "mais vraiment bidon",
+    }
+
+    grid2.coordinate_reference_system = coordinate_system
```
domfournier, line 87: We should test the raise with bad dict values from the setter.

domfournier (commented):
Looking good. Let's just beef up the test a bit.

geoh5py/shared/entity.py
MatthieuCMira: it will raise an error if "Current" not in the coordinate system. 
Accepted and changed to:

        coordinate_reference_system = DEFAULT_CRS

        if self._metadata is not None and "Coordinate Reference System" in self._metadata:
            coordinate_reference_system = self._metadata["Coordinate Reference System"].get(
                "Current", DEFAULT_CRS)

tests/coordinate_system_test.py
MatthieuCMira: Just do this test thinking of implementing it with the tag image but it's definitely for a next step. 
We could keep it there hopping we'll find a solution.

domfournier: Sounds good. just change the name of the test then, so it's not duplicating in our test suite

geoh5py/shared/entity.py
domfournier: Good thinking. Not sure if it's ever allowed to have coordinate system without "Current", but that's safer for sure.

PR #370, opened 2023-07-12
GEOPY-1002: Have a dip property to geoimage.

MatthieuCMira:
**GEOPY-1002 - dip and rotation**
Geoimage has now a dip getter and setter, based on the vertices + minor cleaning of the class

New functions to utils

Minor change in utils.compare_entities


geoh5py/objects/geo_image.py
```diff
@@ -171,6 +176,56 @@ def default_vertices(self):
             )
         return None
 
+    @property
+    def dip(self) -> float:
+        """
+        The dip of the image in degrees
```
domfournier, line 182:
```suggestion
        Calculated dip of the image in degrees from the vertices position.
```

geoh5py/objects/geo_image.py
```diff
@@ -521,61 +586,37 @@ def to_grid2d(
...
-            self._vertices["y"] = rotated_vertices_centered[:, 1] + center[1]
+        if not isinstance(xyz, np.ndarray) or xyz.shape != (4, 3):
+            raise ValueError("Input 'vertices' must be a numpy array of shape (4, 3)")
 
-            # Update the stored rotation
-            self._rotation = new_rotation
+        xyz = np.asarray(
+            np.core.records.fromarrays(xyz.T, names="x, y, z", formats="<f8, <f8, <f8")
+        )
+        self._vertices = xyz
```
domfournier, line 621: You should probably reset the `dip=None` if the vertices change?

tests/geo_image_test.py
```diff
@@ -411,3 +411,25 @@ def test_clipping_rotated_image(tmp_path):
...
+    geoimage2 = geoimage.copy()
+    geoimage2.rotation = 66
+
+    assert geoimage2.rotation == 66
+
+    geoimage3 = geoimage.copy()
+
+    geoimage3.dip = 44
+
+    assert geoimage3.dip == 44
```
domfournier, line 435: Can we validate the vertices position on change of dip and azimuth?

domfournier (commented):
Great. See comments.

geoh5py/objects/geo_image.py
MatthieuCMira: Sure, but calling dip setter will set vertices, that will set dip to None.

It's okay, as far as dip can be recomputed, but a bit heavy isn't it?

I'll change it your way for now.

domfournier: Hmm no, you can just do this here

`self._dip = None`

So next time dip is requested, it will have to be recomputed.

tests/geo_image_test.py
MatthieuCMira: Done.

geoh5py/objects/geo_image.py
MatthieuCMira: Is it a big deal to recompute _dip and _rotation every time like I did?
It's not consuming, keep the code clear and avoid this kind of errors.

domfournier: Yep that's fine. Not super efficient, but makes sense since we are actually not storing the dip value on file.

MatthieuCMira: it's a 4 points matrix, and storing means running it 2 times anyway.

geoh5py/objects/geo_image.py
```diff
@@ -175,6 +174,46 @@ def default_vertices(self):
...
+        :return: the dip angle.
+        """
+        if self.vertices is None or self.rotation is None:
+            raise AttributeError("The image has no vertices")
+
+        # Get rotation matrix
+        rotation_matrix = xy_rotation_matrix(np.deg2rad(-self.rotation))
+
+        # Rotate the vertices
+        rotated_vertices = np.dot(self.vertices, rotation_matrix.T).T
```
domfournier, line 190: Usually affine transformations are written as y = Ax...

```suggestion
        rotated_vertices = rotation_matrix @ self.vertices.T
```

geoh5py/objects/geo_image.py
```diff
@@ -399,65 +434,50 @@ def mask_by_extent(
...
+    @property
+    def origin(self) -> np.array | None:
+        """
+        The origin of the image.
+        :return: an array of the origin of the image in x, y, z.
+        """
+        if self.vertices is not None:
+            return np.array(
+                [self.vertices[3, 0], self.vertices[3, 1], self.vertices[3, 2]]
+            )
```
domfournier, line 446:
```suggestion
            return self.vertices[3, :]
```

geoh5py/objects/geo_image.py
```diff
@@ -399,65 +434,50 @@ def mask_by_extent(
...
+
+        return None
+
     @property
     def rotation(self) -> float | None:
         """
         The rotation of the image in degrees
```
domfournier, line 453:
```suggestion
        The rotation of the image in degrees, counter-clockwise.
        
```

tests/geo_image_test.py
```diff
@@ -411,3 +411,33 @@ def test_clipping_rotated_image(tmp_path):
...
+
+    np.testing.assert_array_almost_equal(geoimage3.dip, 44)
+
+    geoimage4 = geoimage.copy()
+
+    geoimage4.dip = 44
+    geoimage4.rotation = 66
+
+    np.testing.assert_array_almost_equal(geoimage4.dip, 44)
+    np.testing.assert_array_almost_equal(geoimage4.rotation, 66)
```
domfournier, line 443: Still only checking the values have changed, but we should also test the side effects on `vertices`, and vice versa.

domfournier (commented):
We should still improve our tests a bit.

geoh5py/objects/geo_image.py
MatthieuCMira: Okay I did a mistake here.

The correct function should be:
np.dot(rotation_matrix, self.vertices.T).T

or :
(rotation_matrix @ self.vertices.T).T

which syntax do you prefer?
np dot was used everywhere else, that's why I used it

domfournier: The new preferred syntax for array dot operation is with @ I believe. You could also skip the last transpose if you index your array differently afterwards. But whatever, both works. I have been picky enough ;)

MatthieuCMira: Sure but vertices setter should be of shape (4, 3).
So anyway, we have to transpose it. 
I'll let the dot for this PR, and change massively the dot to @ in a next PR.

MatthieuCMira:
Okay, Now GEOPY-1002B and C are merged in this branch. 
Lets finish the correction here :-)

tests/grid_2d_test.py
```diff
@@ -151,3 +153,31 @@ def test_grid2d_to_geoimage(tmp_path):
...
+    grid1.rotation = 66
+
+    grid2 = grid.copy()
+
+    grid2.dip = 90
+
+    grid3 = grid.copy()
+
+    grid3.rotation = 66
+    grid3.dip = 44
```
domfournier, line 183: Still got this weird test. Mind fixing?

domfournier (commented):
Code looks good, but testing the result seems to fail on my side.
Looks like we got our centroid locations wrong on the grid2d, which results in the wrong cells being cut out.

image

domfournier:
@MatthieuCMira Would you mind taking a look at the centroids calc?

MatthieuCMira:
> @MatthieuCMira Would you mind taking a look at the centroids calc?

It's seems at the right location on my side.
Moreover, the cell you show is outside the extent anyway.
I don't get what is wrong?

domfournier:
Maybe outside the extent, but they are still shifted by 10 m.

MatthieuCMira:
> 

It's because the centroid is at the center of the cell, not at the bottom right.

domfournier:
Just looking at the wrong grid. Lets split this test into 3, too many things happening in one.

MatthieuCMira:
"vertical" property was missing around with macos. 
Another bug bites the dust

MatthieuCMira:
Should we merge or still some work to do?

domfournier:
Well testing didn't work since I still got a crash at the end.

Should at least fix this 
image

domfournier (approved):
Great. Let's address the few kinks in GEOPY-1013 and GEOPY-1014

PR #372, opened 2023-07-13
Geopy 1002 c

MatthieuCMira:
**GEOPY-1002 - dip and rotation**
add dip to the conversion of geoimage and grid2d.
Should be extended tested.


domfournier (approved):
Perfect. Thanks for the test.

PR #373, opened 2023-07-13
GEOPY-999

benk-mira:
**GEOPY-999 - Raise ValueError for non-monotically increasing prism indices in drape model layers setter**


jamiefb (approved):
Looks good!

PR #379, opened 2023-07-20
GEOPY-1000: change uuid for all metadata on copy of EM objects

domfournier:
**GEOPY-1000 - Need to change uuid for all metadata on copy of EM objects**


domfournier:
@sebhmg Sorry Seb, but I lost my reviewer. Any chance you can take a quick look before we close the sprint?

geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -219,55 +231,61 @@ def copy(  # pylint: disable=too-many-arguments
...
-            ] = new_complement.uid
-            new_complement.metadata = metadata
-
-        new_entity.metadata = metadata
         return new_entity
 
+    def copy_complement(
+        self,
+        new_entity,
+        parent=None,
```
sebhmg, line 262: **suggestion:** add type hints for params and return type
possibly also some doc

geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -219,55 +231,61 @@ def copy(  # pylint: disable=too-many-arguments
...
+        copy_children: bool = True,
+        clear_cache: bool = False,
+        mask: np.ndarray | None = None,
+    ):
+        base_object = (
+            self.base_transmitter_type  # pylint: disable=no-member
+            if isinstance(self, self.default_receiver_type)
+            else self.base_receiver_type  # pylint: disable=no-member
+        )
+        new_complement = super(  # pylint: disable=bad-super-call
```
sebhmg, line 272: **question:** I do not understand the use of `super()` here.

Regular usage if for calling the method on the parent class of the same instance
```python
class C(B):
    def method(self, arg):
        super().method(arg)    # This does the same thing as: super(C, self).method(arg)
        // do something more than parent class...
```

geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -314,46 +332,57 @@ def edit_metadata(self, entries: dict[str, Any]):
...
+        )
+
         for value in values:
-            if self.property_groups is not None and value not in self.property_groups:
+            if self.property_groups is None:
+                continue
+
+            if not isinstance(value, (PropertyGroup, str)):
+                raise TypeError(
+                    "Input value for 'Property groups' must be a PropertyGroup, "
```
sebhmg, line 376: **issue:** exception message is inconsistent with the types used inthe `isinstance()` test

geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -314,46 +332,57 @@ def edit_metadata(self, entries: dict[str, Any]):
...
+                    "list of PropertyGroup or None."
+                )
+
+            if (value not in groups) and (value not in groups.values()):
```
sebhmg, line 380: **suggestion (readability, non-blocking):** `if not(value in groups or value in groups.values())`

geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -219,55 +231,61 @@ def copy(  # pylint: disable=too-many-arguments
...
             mask=mask,
             cell_mask=cell_mask,
+            omit_list=OMIT_LIST,
             **kwargs,
         )
 
-        metadata["EM Dataset"][new_entity.type] = new_entity.uid
+        for key, value in self.metadata["EM Dataset"].items():
+            if not isinstance(value, (uuid.UUID, type(None))):
+                new_entity.edit_metadata({key: value})
```
sebhmg, line 246: **suggestion (maintanability):** add some comment

quoting you 😉 
> UUIDs will get re-set when the copy of the transmitter gets set, then the metadata gets updates

sebhmg (changes requested):
please see some suggestions and questions

geoh5py/objects/surveys/electromagnetics/base.py
domfournier: Done

geoh5py/objects/surveys/electromagnetics/base.py
domfournier: Done. Replaced with base_copy

geoh5py/objects/surveys/electromagnetics/base.py
domfournier: Replaced with base_copy

geoh5py/objects/surveys/electromagnetics/base.py
domfournier: Done. Simpler logic

geoh5py/objects/surveys/electromagnetics/base.py
domfournier: Done

geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -173,6 +177,27 @@ def add_validate_component_data(self, name: str, data_block: list | dict):
...
+    def base_copy(
+        self,
+        parent: Group | Workspace | None = None,
+        copy_children: bool = True,
+        clear_cache: bool = False,
+        mask: np.ndarray | None = None,
+        **kwargs,
+    ):
+        """
+        Call the super().copy of the class.
```
sebhmg, line 189: **suggestion (doc):** explain the difference between `base_copy()` and `copy()`, or when to call one rather than the other.

Or maybe `base_copy()` can be renamed to convey a better clue on its role.

sebhmg (changes requested):
please see suggestion for improving doc or name of `base_copy()`

geoh5py/objects/surveys/electromagnetics/base.py
domfournier: How about we make a private method `_super_copy`. Should only be used as internal mechanics of the class.

sebhmg (approved):

PR #380, opened 2023-07-21
GEOPY-1004

domfournier:


jamiefb (dismissed):
Looks good :))

tests/h5_non_ascii_filename_test.py
```diff
@@ -65,3 +69,24 @@ def test_existing_non_ascii_filename(tmp_path: Path):
 
     assert file_path.is_file()
     assert not h5py.is_hdf5(file_path)
+
+
+def test_non_ascii_path_geoh5(tmp_path: Path):
+    path = tmp_path / NON_ASCII_FILENAME
+    path.mkdir(parents=True, exist_ok=True)
+    workspace = Workspace.create(path / "test.geoh5")
```
sebhmg, line 77: **question:** should there be another test with a string as input instead of Path?

tying to cover the fix  in the Workspace setter for h5file. Other said, would this test break if the code in workspace.py had not been changed?

sebhmg (commented):
(non-blocking) possibly consider adding a test using string as input instead of Path.

sebhmg (approved):

PR #381, opened 2023-07-24
Creating a Boolean datatype

MatthieuCMira:
**GEOPY-776 - Support masking attributes**
creating a boolean datatype.

This is a premice of the masking object.
Now, the boolean can be imported and exported to/from GA.

/!\ The boolean are imported as integer and exported as booleans


geoh5py/data/boolean_data.py
```diff
@@ -0,0 +1,77 @@
...
+import warnings
+
+import numpy as np
+
+from .data import PrimitiveTypeEnum
+from .integer_data import IntegerData
+
+
+class BooleanData(IntegerData):
+    @classmethod
```
domfournier, line 29:
```suggestion
    """
    Data class for logical (bool) values.
    """
    
    @classmethod
```

tests/boolean_test.py
```diff
@@ -0,0 +1,51 @@
...
+        values = values.astype(bool)
+
+        grid.add_data(
+            {
+                "my_boolean": {
+                    "association": "CELL",
+                    "values": values,
+                }
+            }
+        )
```
domfournier, line 51: Can you add a piece for re-loading the data in a new workspace and checking the type?

domfournier (commented):
Looking good. Minor suggestions below.

domfournier (dismissed):
Looking good. Moving on to testing

geoh5py/data/boolean_data.py
```diff
@@ -0,0 +1,90 @@
...
+                f"Input 'values' for {self} must be of type {np.ndarray} or None."
+            )
+
+        # removing no data values from the array and replace them by "False"
+        values[np.isnan(values)] = self.ndv
+        values = np.where(values == self.ndv, 0, values)
+
+        if isinstance(values, np.ndarray) and values.dtype not in [
+            np.uint32,
+            np.int32,
```
domfournier, line 73: Only except without warning bool type or int8 from geoh5

geoh5py/data/boolean_data.py
```diff
@@ -0,0 +1,90 @@
...
+            bool,
+        ]:
+            warnings.warn(
+                f"Values provided in {values.dtype} are converted to int for "
+                f"{self.primitive_type()} data '{self.name}.'"
+            )
+
+        values = values.astype(np.int32)
```
domfournier, line 81:
```suggestion
        values = values.astype(bool)
```

geoh5py/data/boolean_data.py
```diff
@@ -0,0 +1,90 @@
...
+
+        if set(np.unique(values)) - {0, 1} != set():
+            raise ValueError(
+                f"Values provided by {self.name} are not containing only 0 or 1"
+            )
+
+        self._values = values.astype(bool)
```
domfournier, line 88:
```suggestion
```

geoh5py/data/boolean_data.py
```diff
@@ -0,0 +1,90 @@
...
+            bool,
+        ]:
+            warnings.warn(
+                f"Values provided in {values.dtype} are converted to int for "
+                f"{self.primitive_type()} data '{self.name}.'"
+            )
+
+        values = values.astype(np.int32)
+
+        if set(np.unique(values)) - {0, 1} != set():
```
domfournier, line 83:
```suggestion
        if set(values) - {False, True} != set():
```

domfournier: Nice trick to check for possible matches!

geoh5py/data/boolean_data.py
domfournier, line 73:
```suggestion
            np.int8
```

domfournier (commented):
Couple more details. Thanks!

geoh5py/data/boolean_data.py
MatthieuCMira: technically {False, True} == {0, 1}

you prefer {False, True} anyway?

domfournier: I think it would make more sense since we are going to convert to bool type above.

geoh5py/data/boolean_data.py
MatthieuCMira: You mean raising an error if dtype != bool or np.int8?

domfournier (approved):
Looking good.

PR #383, opened 2023-07-27
Creating a save_as function in workspace with path checking.

MatthieuCMira:
**GEOPY-993 - Add save_as option to create a new file **
Replace "save" functions to "save_as" everywhere

Add Warning deprecation to save functions in both workspace and entity.

PROBLEM: save_as function in entity does not allow the creation of a dataframe. maybe remove savee and save_as in entity?


geoh5py/shared/entity.py
```diff
@@ -515,11 +518,29 @@ def remove_data_from_group(
...
+    #     :param filepath: The path of the new workspace.
+    #     :param add_children: Option to also save the children.
+    #     :return: the entity in the new workspace
+    #     """
+    #     # create a new workspace
+    #     workspace = Workspace.create(filepath)
+    #
+    #     # save the entity in the new workspace
+    #     return workspace.save_entity(self, add_children=add_children)
+
```
domfournier, line 543: Ok well, looks like that's not possible without a major re-write of the class. Let's just remove save_as here then...

tests/copy_entity_test.py
```diff
@@ -88,9 +88,14 @@ def test_copy_entity(tmp_path: Path):
         with pytest.raises(AssertionError, match="Input metadata must be of type"):
             entity.metadata = 0
 
-        _ = entity.save()
+        # _ = entity.save_as(tmp_path / r"testProject2.geoh5")
```
domfournier, line 91:
```suggestion
```

domfournier (approved):
Pretty simple if we keep it on the workspace. We can re-visit this later for the entity.

geoh5py/shared/entity.py
MatthieuCMira: removing the commented part of the code.

geoh5py/shared/entity.py
```diff
@@ -28,6 +29,8 @@
 
 from geoh5py.shared.utils import str2uuid
 
+# from geoh5py.workspace import Workspace
+
```
domfournier, line 33:
```suggestion
```

domfournier (dismissed):

domfournier (approved):
Great, that worked!

PR #384, opened 2023-07-28
Geopy 776

MatthieuCMira:
**GEOPY-776 - Support masking attributes**
Ensure values type of Boolean Data type is bool


domfournier (approved):
Good catch

PR #385, opened 2023-07-31
Update doctstrings for prisms and layers properties

benk-mira:
**GEOPY-1025 - Improve docs strings for drape_model object**


domfournier (approved):
Thanks for this.

PR #388, opened 2023-08-02
Geopy 1014

MatthieuCMira:
**GEOPY-1014 - Crop geoimage by extent yields large files. Look into compression format of PIL**
geoimage compression.
Geoimage is now a png file.

### DIFF
--- a/geoh5py/objects/geo_image.py
+++ b/geoh5py/objects/geo_image.py
@@ -358,7 +358,9 @@ class GeoImage(ObjectBase):
                 value -= value.min()
                 value *= 255.0 / value.max()
                 value = value.astype("uint8")
+
             image = Image.fromarray(value)
+
         elif isinstance(image, str):
             if not Path(image).is_file():
                 raise ValueError(f"Input image file {image} does not exist.")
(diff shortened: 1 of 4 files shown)


geoh5py/objects/geo_image.py
```diff
@@ -380,12 +382,16 @@ def image(self, image: str | np.ndarray | BytesIO | Image.Image):
                 f"Get type {type(image)} instead."
             )
 
-        with TemporaryDirectory() as tempdir:
-            ext = getattr(image, "format")
-            temp_file = (
-                Path(tempdir) / f"image.{ext.lower() if ext is not None else 'tiff'}"
+        if image.mode not in ["RGBA", "RGB", "L"]:
```
domfournier, line 385: I think "CMYK" would be supported as well

geoh5py/shared/conversion/grid2d.py
```diff
@@ -211,6 +211,9 @@ def convert_to_pillow(data: np.ndarray, mode: str | None = None) -> Image:
             if mode is None:
                 raise IndexError("Only 1, 3, or 4 layers can be selected")
 
+        if mode not in ["L", "RGB", "RGBA"]:
```
domfournier, line 214: Same here.

domfournier (commented):
See comment about CMYK

geoh5py/objects/geo_image.py
MatthieuCMira: No it won't. 
PNG files does not support it. 

A workaround could be to manage the creation of a jpeg if mode = "CMYK".
However, it takes other compression options, and that's more code. 

Which option do we choose?

geoh5py/shared/conversion/grid2d.py
MatthieuCMira: Same answer

geoh5py/objects/geo_image.py
MatthieuCMira: Create a dictionary to manage compression options

geoh5py/shared/conversion/grid2d.py
```diff
@@ -207,10 +211,16 @@ def convert_to_pillow(data: np.ndarray, mode: str | None = None) -> Image:
                 mode = "RGB"
             elif data.shape[-1] == 4:
                 mode = "RGBA"
+                # todo: I do not like RGBA as default, rather CMYK.
+                #  crop image by extent should return the same format,
+                #  not RGBA, and 0 for nan values.
```
domfournier, line 216: I don't agree, but we should have that conversation. Remove comment here.

PR #391, opened 2023-08-08
Geopy 438

MatthieuCMira:
Allow get_entity to accept Property_group

Add a property_group all along the workspace entity getter.

Test it in pytests.

(Not 100% sure about the good usage of some functions).

### DIFF
--- a/geoh5py/objects/object_base.py
+++ b/geoh5py/objects/object_base.py
@@ -343,7 +343,9 @@ class ObjectBase(Entity):
             ):
                 kwargs["property_group_type"] = "Multi-element"
 
-            prop_group = PropertyGroup(self, **kwargs)
+            prop_group = self.workspace.create_property_group(
+                PropertyGroup(self, **kwargs)
+            )
 
             property_groups += [prop_group]
 
--- a/geoh5py/workspace/workspace.py
+++ b/geoh5py/workspace/workspace.py
@@ -75,6 +75,7 @@ if TYPE_CHECKING:
     from ..shared import EntityType
 
 
+# pylint: disable=too-many-instance-attributes
 class Workspace(AbstractContextManager):
     """
     The Workspace class manages all Entities created or imported from the
@@ -94,7 +95,6 @@ class Workspace(AbstractContextManager):
     """
 
     _active_ref: ClassVar[ReferenceType[Workspace]] | type(None) = type(None)  # type: ignore
-
     _attribute_map = {
         "Contributors": "contributors",
         "Distance unit": "distance_unit",
@@ -115,18 +115,19 @@ class Workspace(AbstractContextManager):
     ):
         self._data: dict[uuid.UUID, ReferenceType[data.Data]] = {}
         self._distance_unit: str = distance_unit
-        self._contributors: list[str] = np.asarray(
+        self._contributors: np.ndarray = np.asarray(
             contributors, dtype=h5py.special_dtype(vlen=str)
         )
         self._ga_version: str = ga_version
         self._geoh5: h5py.File | bool = False
-        self._groups: dict[uuid.UUID, ReferenceType[group.Group]] = {}
+        self._groups: dict[uuid.UUID, ReferenceType[Group]] = {}
+        self._property_groups: dict[uuid.UUID, ReferenceType[PropertyGroup]] = {}
         self._h5file: str | Path | BytesIO | None = None
         self._mode: str = mode
         self._name: str = name
-        self._objects: dict[uuid.UUID, ReferenceType[object_base.ObjectBase]] = {}
+        self._objects: dict[uuid.UUID, ReferenceType[ObjectBase]] = {}
         self._repack: bool = repack
-        self._root: Entity | None = None
+        self._root: Entity | PropertyGroup | None = None
         self._types: dict[uuid.UUID, ReferenceType[EntityType]] = {}
         self._version: float = version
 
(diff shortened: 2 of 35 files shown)


geoh5py/workspace/workspace.py
```diff
@@ -524,12 +543,12 @@ def data(self) -> list[data.Data]:
 
     def fetch_or_create_root(self):
         root = self.load_entity(uuid.uuid4(), "root")
-        if root:
+        if root is not None:
```
domfournier, line 546:
```suggestion
        if root is not None and not isinstance(self._root, PropertyGroup):
```

geoh5py/workspace/workspace.py
```diff
@@ -524,12 +543,12 @@ def data(self) -> list[data.Data]:
...
             self._root = root
-            self._root.on_file = True
-            self._root.entity_type.on_file = True
+            if not isinstance(self._root, PropertyGroup):
+                self._root.on_file = True
+                self._root.entity_type.on_file = True
```
domfournier, line 550:
```suggestion
            self._root.on_file = True
            self._root.entity_type.on_file = True
```

domfournier (commented):
Originally we wanted to access the property_groups from the object.get_entity(), rather than from the workspace. Any chance you could expand the implementation to `ObjectBase`?

geoh5py/objects/object_base.py
```diff
@@ -381,6 +384,43 @@ def get_data_list(self, attribute="name") -> list[str]:
...
+        name_list = [
+            child.name for child in self.children if isinstance(child, entity_type)
+        ]
+
+        if self.property_groups is not None:
+            name_list += [
+                group.name
+                for group in self.property_groups
+                if isinstance(group, entity_type)
+            ]
```
domfournier, line 420: We wouldn't need this if we just treat PropertyGroups like a child.

geoh5py/workspace/workspace.py
```diff
@@ -1064,24 +1097,25 @@ def load_entity(
...
         :param entity_type: One of entity type 'group', 'object', 'data' or 'root'
-
+        :param parent: Parent entity.
         :return entity: Entity loaded from geoh5
         """
-        if isinstance(self.get_entity(uid)[0], Entity):
-            return self.get_entity(uid)[0]
+        get_entity_ = self.get_entity(uid)[0]
+        if isinstance(get_entity_, (Entity, PropertyGroup)):
+            return get_entity_
```
domfournier, line 1110:
```suggestion
        child = self.get_entity(uid)[0]
        if isinstance(child, (Entity, PropertyGroup)):
            return child
```

Not really PEP8 compliant.

geoh5py/objects/object_base.py
```diff
@@ -438,6 +480,25 @@ def remove_property_groups(
 
         self.workspace.update_attribute(self, "property_groups")
 
+    def get_property_group(self, name: uuid.UUID | str) -> list:
```
domfournier, line 483: I like that. We should eventually deprecate the `find_or_create_property_group` method in favor of two distinct methods...

domfournier (commented):
Can we give another shot at simply including PropertyGroups in the children list of ObjectBase?

geoh5py/objects/object_base.py
MatthieuCMira: The problem is children based on shared entity. 
If we having property_group to children, we should have the property of entity to property_group. 
It can be done, but does it makes sens from a GA perspective?
For exemple, remove_entity should takes property_group as input. 
Or having exception everywhere in the code?

geoh5py/objects/object_base.py
MatthieuCMira: Find or create property is interesting in a way. 
I just split this function in find_property and create_property and merge them both in the find_or_create_property group. 
Make the code clearer.

geoh5py/objects/object_base.py
domfournier: you just need to worry about the methods of ObjectBase, as it overrides the method of the base entity class.

MatthieuCMira: Sure but Property_group is not an object base.
In the end, or I have code in object_base to redefine get_entity functions
Or I put exception everywhere children is called to be saved, which means adding a lot of code too. 
 => warnings and pass, or save parent entity when property groups is trying to be saved.

domfournier: I will moke it, you can decide

MatthieuCMira: Removed

geoh5py/objects/object_base.py
```diff
@@ -415,6 +459,15 @@ def property_groups(self) -> list[PropertyGroup] | None:
...
         """
+        property_groups = [
+            child for child in self.children if isinstance(child, PropertyGroup)
+        ]
+
+        if self._property_groups is not None:
+            self._property_groups = list(set(self._property_groups + property_groups))
+        elif len(property_groups) > 0:
+            self._property_groups = property_groups
+
```
domfournier, line 470: Why do we need to check if `self._property_group` is not None and append?
Can maybe simplify this with:

```suggestion
        if self._property_groups is None:
            property_groups = [
                child for child in self.children if isinstance(child, PropertyGroup)
            ]

            if len(property_groups) > 0:
                self._property_groups = property_groups
                
        return self._property_groups

```

domfournier (commented):
Small comment. Looking great otherwise...

geoh5py/objects/object_base.py
MatthieuCMira: This function handle the scenario where a property group is in children but not in the _property_group. I don't know when it can occurs, but the function was made to handle this scenario.

If this scenario is impossible, we could choose to not handle it, will make the function simpler.

domfournier: might need to double check that it behaves as intended.

domfournier (dismissed):
Way better than it was. Great collab!

PR #392, opened 2023-08-09
GEOPY-1030

benk-mira:
**GEOPY-1030 - Concatenated drillhole data stored in property group can have different lengths when it should be collocated**

### DIFF
--- a/geoh5py/data/numeric_data.py
+++ b/geoh5py/data/numeric_data.py
@@ -76,7 +76,11 @@ class NumericData(Data, ABC):
         :param values: the values to check.
         :return: the values with the right length.
         """
-        if self.n_values is not None:
+
+        if (
+            not any(k in self.name for k in ["DEPTH", "FROM", "TO"])
+            and self.n_values is not None
+        ):
             if len(values) < self.n_values:
                 full_vector = (
                     np.ones(self.n_values, dtype=values.dtype) * self.nan_value
--- a/geoh5py/data/numeric_data.py
+++ b/geoh5py/data/numeric_data.py
@@ -77,19 +77,19 @@ class NumericData(Data, ABC):
         :return: the values with the right length.
         """
 
-        if self.n_values is not None:
-            if len(values) < self.n_values:
-                full_vector = (
-                    np.ones(self.n_values, dtype=values.dtype) * self.nan_value
-                )
-                full_vector[: len(np.ravel(values))] = np.ravel(values)
-                return full_vector
+        if self.n_values is None:
+            return values
 
-            if len(values) > self.n_values:
-                raise ValueError(
-                    f"Input 'values' of shape({self.n_values},) expected. "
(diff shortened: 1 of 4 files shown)


geoh5py/data/numeric_data.py
```diff
@@ -76,7 +76,11 @@ def format_length(self, values: np.ndarray) -> np.ndarray:
         :param values: the values to check.
         :return: the values with the right length.
         """
-        if self.n_values is not None:
+
+        if (
+            not any(k in self.name for k in ["DEPTH", "FROM", "TO"])
```
domfournier, line 81: Would be better if the check happened in the concatenated data, so that we don't cluter the base class with drillhole stuff. Maybe `return self.n_values = None` if it's a DEPTH, FROM or TO data, then we just return the value here without extending.

geoh5py/shared/concatenation.py
```diff
@@ -695,6 +695,20 @@ def parent(self, parent):
         if f"Property:{self.name}" not in parental_attr:
             parental_attr[f"Property:{self.name}"] = as_str_if_uuid(self.uid)
 
+    @property
+    def n_values(self) -> np.ndarray:
+        """Number of values in the data."""
+
+        n_values = None
+        depths = getattr(self.property_group, "depth_", None)
+        if depths:
```
domfournier, line 704:
```suggestion
        if depths and depths is not self:
```

geoh5py/shared/concatenation.py
```diff
@@ -695,6 +695,20 @@ def parent(self, parent):
...
+    @property
+    def n_values(self) -> np.ndarray:
+        """Number of values in the data."""
+
+        n_values = None
+        depths = getattr(self.property_group, "depth_", None)
+        if depths:
+            n_values = len(depths.values)
+        intervals = getattr(self.property_group, "from_", None)
+        if intervals:
```
domfournier, line 707:
```suggestion
        if intervals and intervals is not self:
```

geoh5py/shared/concatenation.py
benk-mira: doesn't work because the getattr results in a reference back to the n_values property.

domfournier: `n_values` of the Concatenated class yes?

benk-mira: ..ok, maybe it works after all... verifying

geoh5py/data/numeric_data.py
benk-mira: Done

geoh5py/shared/concatenation.py
benk-mira: done.

geoh5py/shared/concatenation.py
benk-mira: Done.

domfournier (approved):

PR #394, opened 2023-08-14
add PropertyGroup to objects children

MatthieuCMira:
**GEOPY-438 - Allow property_groups query from workspace.get_entity**
pass property groups to children.

Changes in a bunch of files and in the tests.


domfournier (commented):
Looking good for merging, we can finish testing on GEOPY-438

PR #397, opened 2023-08-21
GEOPY-438: Update

domfournier:


geoh5py/groups/property_group.py
```diff
@@ -108,31 +137,34 @@ def name(self) -> str:
...
-    def parent(self) -> Entity | None:
+    def on_file(self):
         """
-        The parent :obj:`~geoh5py.objects.object_base.ObjectBase`
+        Property group is on geoh5 file.
         """
-        return self._parent
+        return self._on_file
 
-    @parent.setter
```
MatthieuCMira, line 120: Shouldn't we keep the parent setter?
The parent property is still there.

geoh5py/groups/property_group.py
```diff
@@ -156,6 +198,22 @@ def property_group_type(self) -> str:
     def property_group_type(self, group_type: str):
         self._property_group_type = group_type
 
+    def remove_properties(self, data: Data | list[Data]):
```
MatthieuCMira, line 201: what if we want to remove property from a property name or uid?

geoh5py/groups/property_group.py
```diff
@@ -60,6 +71,24 @@ def __init__(self, parent: ObjectBase, **kwargs):
             except AttributeError:
                 continue
 
+        self.parent.workspace.register_property_group(self)
+
+    def add_properties(self, data: Data | list[Data]):
```
MatthieuCMira, line 76: I am not sure about this but should the data already present on a workspace aren't they?
So accepting uid and name to add properties could be uselfull?

geoh5py/groups/property_group.py
```diff
@@ -41,16 +41,27 @@ class PropertyGroup(ABC):
...
+    ):
+        self.name = name or "property_group"
+        self.uid = uid or uuid.uuid4()
         self._allow_delete = True
+        self.on_file = on_file
         self._association: DataAssociationEnum = DataAssociationEnum.VERTEX
-        self._properties: list[uuid.UUID] = []
+        self._properties: list[uuid.UUID] | None = None
         self._property_group_type = "Multi-element"
-        self.parent: ObjectBase = parent
```
MatthieuCMira, line 53: Why removing parents?
Property Groups are associated to an object anyway?

geoh5py/objects/object_base.py
```diff
@@ -72,8 +72,17 @@ def add_children(self, children: list[Entity] | list[PropertyGroup]):
         :param children: Add a list of entities as
             :obj:`~geoh5py.shared.entity.Entity.children`
         """
-        super().add_children(children)
-        self._property_groups = None
+        property_groups = self._property_groups or []
+
```
MatthieuCMira, line 76: Shouldn't we test the type of children? Ensure it's a data, object of PropertyGroup?
super() children was already doing that.

geoh5py/objects/object_base.py
```diff
@@ -175,31 +184,28 @@ def add_data_to_group(
...
-        association = None
-        template = self.workspace.get_entity(uids[0])[0]
-        if isinstance(template, Data):
-            association = template.association
+        children = []
+        associations = []
+        for entity in data:
+            if isinstance(entity, (uuid.UUID, str)):
+                entity = self.get_entity(entity)[0]
+            children.append(entity)
```
MatthieuCMira, line 195: We should check to ensure children typeis Data.

geoh5py/objects/object_base.py
```diff
@@ -386,11 +392,11 @@ def find_or_create_property_group(self, **kwargs) -> PropertyGroup:
         """
 
         prop_group = None
-        if "name" in kwargs:
-            prop_group = self.get_property_group(kwargs["name"])[0]
+        if name is not None or uid is not None:
```
MatthieuCMira, line 395: what if the user pass a name of an existing property group (or multiple property groups exist with the same name) and another uid? We should  check if uid, then for name.

geoh5py/objects/object_base.py
```diff
@@ -492,8 +479,24 @@ def remove_children(self, children: list[Entity] | list[PropertyGroup]):
...
+                continue
+
+            self._children.remove(child)
+
+            if not self._property_groups:
+                continue
+
+            if isinstance(child, PropertyGroup):
+                self._property_groups.remove(child)
+            elif isinstance(child, Data):
```
MatthieuCMira, line 496: Are data always associated to property groups?
What if a Data is a 'raw' children of an object, we just want to remove it aren't we?

geoh5py/objects/object_base.py
```diff
@@ -609,6 +612,24 @@ def add_default_visual_parameters(self):
...
+        :obj:`~geoh5py.groups.property_group.PropertyGroup` of the object.
+
+        :param data: :obj:`~geoh5py.data.data.Data` object,
+            :obj:`~geoh5py.shared.entity.Entity.uid` or
+            :obj:`~geoh5py.shared.entity.Entity.name` of data.
+        """
+        if not isinstance(data, list):
+            data = [data]
+
+        if not self.property_groups:
```
MatthieuCMira, line 627: self._property_groups ?

MatthieuCMira (commented):
My firsts review ever :') Hope my comment are accurate... 
They are mostly about input type assertion and preventing users' bad uses.
A bit difficult for me to understand how the whole code is working so I'll trust the tests.

geoh5py/groups/property_group.py
domfournier: We could allow for uuid, true. Maybe not from `name` though since too ambiguous.

geoh5py/groups/property_group.py
domfournier: Same as above - cool with uuid, probably not name.

geoh5py/objects/object_base.py
domfournier: We are not allowing two property groups to have the same name - it makes GA angry. But I like the idea of prioritizing uuid. 

Done

geoh5py/objects/object_base.py
domfournier: The plain removal from the list happens above, then from the file below.

geoh5py/objects/object_base.py
domfournier: Probably yes, mostly for consistency with Concatenated class.
Done

domfournier (commented):
See reply and edits. Ready for Round 2

geoh5py/objects/object_base.py
MatthieuCMira: Children cannot be "Objects"?

MatthieuCMira (commented):
Seems good on my side.
I let you merge!

PR #400, opened 2023-08-24
catch properties bad instantiation when copying

MatthieuCMira:
**GEOPY-1042 - cannot create several property groups**
An error was raised when a property group with no associated data was added to an object, and when the user tried to copy such object. 

Catch by verifying the input of properties.


domfournier (approved):
Nice catch!

PR #401, opened 2023-08-30
GEOPY-1017

jamiefb:
**GEOPY-1017 - Profile OMF large file conversion**

### DIFF
--- a/geoh5py/io/h5_writer.py
+++ b/geoh5py/io/h5_writer.py
@@ -215,6 +215,7 @@ class H5Writer:
         cls,
         file: str | h5py.File,
         entity,
+        compression: int = 5,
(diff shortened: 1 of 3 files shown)


geoh5py/io/h5_writer.py
```diff
@@ -620,7 +638,7 @@ def write_data_values(  # pylint: disable=too-many-branches
                     name_map,
                     data=out_values,
                     compression="gzip",
-                    compression_opts=9,
```
domfournier, line 623: @MatthieuCMira Can you remind me why we specified compression to 9? Remember you asking me for the images...

domfournier (commented):
We should probably unit-test this....

PR #402, opened 2023-08-30
GEOPY-1055

benk-mira:
**GEOPY-1055 - ConcatenatedDrillhole object from and to properties don't handle empty property list correctly**


domfournier (approved):
Oh wow, yep. That one slipped. We should probably unit test with a ConcatenatedPropertyGroup that has no properties set.

domfournier:
I did not mean to write `Ass minimal test`

domfournier (approved):

benk-mira:
haha, I thought it was a euphemism for REALLY minimal

PR #406, opened 2023-09-06
add group type to monitory_directory_copy

MatthieuCMira:
**GEOPY-1039 - create uijson for curve depth mapper**


domfournier (approved):
Easy pease

PR #408, opened 2023-09-07
GEOPY-1048

benk-mira:
**GEOPY-1048 - UIJson class implementation with parental validation**


domfournier (approved):
To be continued on GEOPY-1068

PR #412, opened 2023-09-12
GEOPY-1068

benk-mira:
**GEOPY-1068 - Add RequiredEnforcer class to validate label/value exist in all FormParameters**

### DIFF
--- a/geoh5py/shared/exceptions.py
+++ b/geoh5py/shared/exceptions.py
@@ -36,9 +36,9 @@ class Geoh5FileClosedError(ABC, Exception):
 class BaseValidationError(ABC, Exception):
     """Base class for custom exceptions."""
 
-    @staticmethod
+    @classmethod
     @abstractmethod
-    def message(name, value, validation):
+    def message(cls, name, value, validation):
         """Builds custom error message."""
         raise NotImplementedError()
 
@@ -49,8 +49,8 @@ class JSONParameterValidationError(Exception):
     def __init__(self, name: str, err: str):
         super().__init__(JSONParameterValidationError.message(name, err))
 
-    @staticmethod
-    def message(name, err):
+    @classmethod
+    def message(cls, name, err):
         return f"Malformed ui.json dictionary for parameter '{name}'. {err}"
 
 
@@ -58,8 +58,8 @@ class UIJsonFormatError(BaseValidationError):
     def __init__(self, name, msg):
         super().__init__(f"Invalid UIJson format for parameter '{name}'. {msg}")
 
-    @staticmethod
-    def message(name, value, validation):
+    @classmethod
+    def message(cls, name, value, validation):
         pass
 
 
@@ -71,8 +71,8 @@ class AggregateValidationError(BaseValidationError):
     ):
         super().__init__(AggregateValidationError.message(name, value))
 
-    @staticmethod
-    def message(name, value, validation=None):
+    @classmethod
+    def message(cls, name, value, validation=None):
         msg = f"\n\nValidation of '{name}' collected {len(value)} errors:\n"
         for i, err in enumerate(value):
             msg += f"\t{i}. {str(err)}\n"
@@ -90,8 +90,8 @@ class OptionalValidationError(BaseValidationError):
     ):
         super().__init__(OptionalValidationError.message(name, value, validation))
 
-    @staticmethod
-    def message(name, value, validation):
+    @classmethod
+    def message(cls, name, value, validation):
         return f"Cannot set a None value to non-optional parameter: {name}."
 
 
@@ -106,8 +106,8 @@ class AssociationValidationError(BaseValidationError):
     ):
         super().__init__(AssociationValidationError.message(name, value, validation))
 
-    @staticmethod
-    def message(name, value, validation):
+    @classmethod
+    def message(cls, name, value, validation):
         return (
             f"Property '{name}' with value: '{value}' must be "
             f"a child entity of parent {validation}"
@@ -120,8 +120,8 @@ class PropertyGroupValidationError(BaseValidationError):
     def __init__(self, name: str, value: PropertyGroup, validation: str):
         super().__init__(PropertyGroupValidationError.message(name, value, validation))
 
-    @staticmethod
-    def message(name, value, validation):
+    @classmethod
+    def message(cls, name, value, validation):
         return (
             f"Property group for '{name}' must be of type '{validation}'. "
             f"Provided '{value.name}' of type '{value.property_group_type}'"
@@ -132,8 +132,8 @@ class AtLeastOneValidationError(BaseValidationError):
     def __init__(self, name: str, value: list[str]):
         super().__init__(AtLeastOneValidationError.message(name, value))
 
-    @staticmethod
-    def message(name, value, validation=None):
+    @classmethod
+    def message(cls, name, value, validation=None):
         opts = "'" + "', '".join(str(k) for k in value) + "'"
         return f"Must provide at least one {name}.  Options are: {opts}"
 
@@ -142,9 +142,45 @@ class RequiredValidationError(BaseValidationError):
     def __init__(self, name: str):
         super().__init__(RequiredValidationError.message(name))
 
-    @staticmethod
-    def message(name, value=None, validation=None):
+    @classmethod
+    def message(cls, name, value=None, validation=None):
         return f"Missing required parameter: '{name}'."
+
+
+class InCollectionValidationError(BaseValidationError):
+    collection = "Collection"
+    item = "data"
+
+    def __init__(self, name: str, value: list[str]):
+        super().__init__(self.message(name, value))
+
+    @classmethod
+    def message(cls, name, value, validation=None):
+        _ = validation
+        return (
+            f"{cls.collection}: '{name}' "
+            f"is missing required {cls.item}(s): {value}."
+        )
+
+
+class RequiredFormMemberValidationError(InCollectionValidationError):
+    collection = "Form"
+    item = "member"
+
+
+class RequiredUIJsonParameterValidationError(InCollectionValidationError):
+    collection = "UIJson"
+    item = "parameter"
+
+
+class RequiredWorkspaceObjectValidationError(InCollectionValidationError):
+    collection = "Workspace"
+    item = "object"
+
+
+class RequiredObjectDataValidationError(InCollectionValidationError):
+    collection = "Object"
+    item = "data"
 
 
 class ShapeValidationError(BaseValidationError):
--- a/geoh5py/ui_json/enforcers.py
+++ b/geoh5py/ui_json/enforcers.py
@@ -239,7 +239,7 @@ class TypeEnforcer(Enforcer):
 
 class ValueEnforcer(Enforcer):
     """
-    Enforces restricted value choice.
+    Enforces restricted value choices.
 
     :param validations: Valid parameter value(s).
     :raises ValueValidationError: If value is not a valid value
@@ -303,7 +303,7 @@ class UUIDEnforcer(Enforcer):
 
 class RequiredEnforcer(Enforcer):
     """
-    Enforces required items exist in a collection.
+    Enforces required items in a collection.
 
     :param validations: items that are required in the
         collection.
--- a/geoh5py/ui_json/enforcers.py
+++ b/geoh5py/ui_json/enforcers.py
@@ -166,16 +166,10 @@ class Enforcer(ABC):
     @abstractmethod
     def rule(self, value: Any):
         """True if 'value' adheres to enforcers rule."""
-        raise NotImplementedError(
-            "Enforcer is abstract. Must be implemented in subclass"
-        )
 
     @abstractmethod
     def enforce(self, name: str, value: Any):
         """Enforces rule on 'name' parameter's 'value'."""
-        raise NotImplementedError(
-            "Enforcer is abstract. Must be implemented in subclass"
-        )
 
     @property
     def validations(self):
--- a/geoh5py/ui_json/enforcers.py
+++ b/geoh5py/ui_json/enforcers.py
@@ -17,7 +17,6 @@
 
 from __future__ import annotations
 
-import uuid
 from abc import ABC, abstractmethod
 from typing import Any
 
@@ -33,6 +32,7 @@ from geoh5py.shared.exceptions import (
     UUIDValidationError,
     ValueValidationError,
 )
+from geoh5py.shared.utils import is_uuid
 
 
 class EnforcerPool:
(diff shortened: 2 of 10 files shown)


geoh5py/ui_json/enforcers.py
```diff
@@ -0,0 +1,350 @@
...
+        try:
+            enforcer.enforce(self.name, value)
+        except BaseValidationError as err:
+            self._errors.append(err)
+
+    def _raise_errors(self):
+        """Raise errors if any exist, aggregate if more than one."""
+        if self._errors:
+            if len(self._errors) > 1:
+                raise AggregateValidationError(self.name, self._errors)
```
domfournier, line 149: Is there an efficiency reason to not always just go through the AggregateValidationError when just one error? Would make the code simpler.

geoh5py/ui_json/enforcers.py
```diff
@@ -0,0 +1,350 @@
...
+    enforcer_type: str = ""
+
+    def __init__(self, validations: Any | None = None):
+        self._validations = validations
+
+    @abstractmethod
+    def rule(self, value: Any):
+        """True if 'value' adheres to enforcers rule."""
+        raise NotImplementedError(
+            "Enforcer is abstract. Must be implemented in subclass"
```
domfournier, line 170: Looks like a copy from below. We have so far never raised anything in geoh5py in the `abstractmethod`, as it is implied that the subclass needs to implement it. Also less to test.

geoh5py/ui_json/enforcers.py
```diff
@@ -0,0 +1,350 @@
...
+        )
+
+    @abstractmethod
+    def enforce(self, name: str, value: Any):
+        """Enforces rule on 'name' parameter's 'value'."""
+        raise NotImplementedError(
+            "Enforcer is abstract. Must be implemented in subclass"
+        )
```
domfournier, line 178:
```suggestion
```

geoh5py/ui_json/enforcers.py
```diff
@@ -0,0 +1,350 @@
...
+
+        return is_equal
+
+    def __str__(self):
+        return f"<{type(self).__name__}> : {self.validations}"
+
+
+class TypeEnforcer(Enforcer):
+    """
+    Enforces type validation.
```
domfournier, line 199:
```suggestion
    Enforces parameter value type.

```

geoh5py/ui_json/enforcers.py
```diff
@@ -0,0 +1,350 @@
...
+            raise ValueValidationError(name, value, self.validations)
+
+    def rule(self, value: Any) -> bool:
+        """True if value is a valid choice."""
+        return value in self.validations
+
+
+class UUIDEnforcer(Enforcer):
+    """
+    Enforces uuid validation.
```
domfournier, line 266:
```suggestion
    Enforces that a value is a valid uuid string or instance.

```

geoh5py/ui_json/enforcers.py
```diff
@@ -0,0 +1,350 @@
...
+        if not self.rule(value):
+            raise UUIDValidationError(
+                name,
+                value,
+            )
+
+    def rule(self, value: Any) -> bool:
+        """True if value is a valid uuid string."""
+
+        if self.validations == "optional":
```
domfournier, line 289: Feels like we are mixing responsability here? The UUIDEnforcer could always accept None, but the form validator could raise an error if not optional instead?

geoh5py/ui_json/enforcers.py
```diff
@@ -0,0 +1,350 @@
...
+    Enforces uuid validation.
+
+    :param validations: None is considered a valid uuid if
+        validations is 'optional'.
+    :raises UUIDValidationError: If value is not a valid uuid string.
+    """
+
+    enforcer_type = "uuid"
+
+    def __init__(self, validations: str | None = None):
```
domfournier, line 275: Could this be a sub-class of TypeEnforcer, where we call the super().rule for check for type (UUID, str, None), then perform the check it's a valid uuid?

geoh5py/ui_json/enforcers.py
```diff
@@ -0,0 +1,350 @@
...
+            return True
+
+        if not isinstance(value, str):
+            return False
+
+        try:
+            uuid.UUID(value)
+            is_uuid = True
+        except ValueError:
+            is_uuid = False
```
domfournier, line 299: I think we could re-use `shared.utils.is_uuid`

geoh5py/ui_json/enforcers.py
```diff
@@ -0,0 +1,350 @@
...
+
+        return is_uuid
+
+
+class RequiredEnforcer(Enforcer):
+    """
+    Enforces required items exist in a collection.
+
+    :param validations: items that are required in the
```
domfournier, line 308:
```suggestion
    :param validations: Items that are required in the

```

geoh5py/ui_json/forms.py
```diff
@@ -0,0 +1,499 @@
...
+
+    def __init__(self, name, value=None, **kwargs):
+        value = StringParameter("value", value=value)
+        super().__init__(name, value=value, **kwargs)
+
+
+class BoolFormParameter(FormParameter):
+    """Boolean parameter type."""
+
+    def __init__(self, name, value=None, **kwargs):
```
domfournier, line 274: Should this not be a bool by default? I would expect the default state to be False?

geoh5py/ui_json/forms.py
```diff
@@ -0,0 +1,499 @@
...
+        self,
+        name: str,
+        value: Any | None = None,
+        **kwargs,
+    ):
+        self.name: str = name
+        self._value: Parameter = self._set_value_parameter(value)
+        self._label = StringParameter("label")
+        self._enabled = BoolParameter("enabled", value=True)
+        self._optional = BoolParameter("optional", value=False)
```
domfournier, line 144: See comment about default state of BoolParameter. Maybe we just specify only if changed from default?

geoh5py/ui_json/forms.py
```diff
@@ -0,0 +1,499 @@
...
+            if len(error_list) == 1:
+                raise error_list.pop()
+            raise AggregateValidationError(self.name, error_list)
+
+        self._extra_members.update(members)
+
+    def validate(self):
+        """Validates data against the pool of enforcers."""
+        enforcers = EnforcerPool.from_validations(self.name, self.validations)
+        enforcers.validate(self.form())
```
domfournier, line 205: Is this old? I don't see a validate method on the enforcer.

geoh5py/ui_json/forms.py
```diff
@@ -0,0 +1,499 @@
...
+    """
+
+    identifier_members: list[str] = ["file_description", "file_type", "file_multi"]
+    validations = {"required_form_members": ["file_description", "file_type"]}
+
+    def __init__(self, name, value=None, **kwargs):
+        self._file_description = StringListParameter("file_description")
+        self._file_type = StringListParameter("file_type")
+        self._file_multi = BoolParameter("file_multi")
+        value = StringParameter("value", value=value)
```
domfournier, line 370: Should probably turn this into a PathParameter eventually, similar behaviour as UUID where we can check if Path or str, and check if it exists.

geoh5py/ui_json/enforcers.py
```diff
@@ -0,0 +1,350 @@
...
+        if enforcer_type == "required_form_members":
+            enforcer = RequiredFormMemberEnforcer(validation)  # type: ignore
+        if enforcer_type == "required_workspace_objects":
+            enforcer = RequiredWorkspaceObjectEnforcer(validation)  # type: ignore
+        if enforcer_type == "required_object_data":
+            enforcer = RequiredObjectDataEnforcer(validation)  # type: ignore
+
+        return enforcer
+
+    def validate(self, value: Any):
```
domfournier, line 130: Suggest that we keep the same `enforce` method name as the individual enforcers?

geoh5py/ui_json/forms.py
domfournier: Ok now I get that the method is named different on the Pool. Maybe we keep the name `enforce`? We could even just make it a `__call__(self)`

domfournier (commented):
Epic! See comments.

geoh5py/ui_json/enforcers.py
benk-mira: I think this is perfectly valid code.  It is sole reason that the NotImplementedError exists (https://docs.python.org/3/library/exceptions.html#NotImplementedError).  That said, I'm willing to remove for coverage, consistency reasons if you like..

geoh5py/ui_json/enforcers.py
benk-mira: Except it doesn't have to be a parameter.  How about 'Enforcer valid type(s)'

geoh5py/ui_json/forms.py
benk-mira: renamed

geoh5py/ui_json/forms.py
benk-mira: Exactly, it's on my list of Enforcers to add!

geoh5py/ui_json/enforcers.py
benk-mira: done

geoh5py/ui_json/forms.py
benk-mira: using the __call__ method here makes the code less readable.  It's more obvious what's going on if we just use `enforcers.enforce`.

geoh5py/shared/exceptions.py
```diff
@@ -142,11 +142,47 @@ class RequiredValidationError(BaseValidationError):
...
+class RequiredWorkspaceObjectValidationError(InCollectionValidationError):
+    collection = "Workspace"
+    item = "object"
+
+
+class RequiredObjectDataValidationError(InCollectionValidationError):
+    collection = "Object"
+    item = "data"
+
+
```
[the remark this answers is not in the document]
benk-mira: Thx!

geoh5py/ui_json/enforcers.py
benk-mira: Updated with 'Enforces valid uuid string.' to keep consistent with others

geoh5py/ui_json/enforcers.py
domfournier: That's fair, although I have mostly seen it for the second reason (work in progress of a class). You technically can't reach that chunk of code though, since the abstract class can't be instantiated without the method being implemented.

image

And for consistency, I would say yes, we remove.

@sebhmg Do you have an opinion generally on this.

geoh5py/ui_json/enforcers.py
benk-mira: It's just more readable when the raise occurs for the user.  I think I prefer complexity in the code rather than in an error message that we needs to parse just to find out that a TypeValidationError was thrown on the value of parameter x. 

The plan is to extract a class here that we can re-use when we implement the uijson update values.   since we'll need to aggregate the errors there too.  That will restore the single responsability principle for this method!

geoh5py/ui_json/enforcers.py
benk-mira: Yes - thank you!

geoh5py/ui_json/enforcers.py
benk-mira: Good point.  Updated.

geoh5py/ui_json/enforcers.py
benk-mira: I'll just use two enforcers on the UUIDParameter.  I have removed TypedParameter subclass and the optional attr for all parameters with type enforcers.  I'm now passing NoneType in TypeEnforcer.rule() and value=None for UUIDEnforcer.rule().  I'll implement an NoneNotOptionalEnforcer for form-level validations.

geoh5py/ui_json/enforcers.py
benk-mira: Just had a closer look at your example.  The point of the NotImplementedError is to prevent a developer from defining a concrete class and forgetting to implement one of the abstract methods.  So to test, we could subclass test_abstract_class, then call implement and check that the NotImplementedError is thrown.

domfournier: Did you run it? Don't think you can instantiate your subclass either without defining `message`

image


Seems redundant anyway, but that's just my opinion.

@sebhmg We can differ to your third-party opinion?

sebhmg: `NotImplementedError` is thrown at runtime, so it will not "prevent a developer from defining a concrete class and forgetting to implement one of the abstract methods".
Form unit test maybe. However would think, as @domfournier point it out, that Python will refuse to instantiate a class with an abstract method. Only case to worry about is for `class` or `static` `abstract` methods (which looks like code small anyway). Am I missing a point? Happy to jump on a call to clarify

sebhmg: and even if  it could be instantiated, python is likely to throw hat exception since it the function was declared abstract. I suspect an explicit `NotImplementedError` is useful only when not using the abstract decorator. Or when planning some features that are not supported yet. Possibly as well for a sub-class which does not implement fully the parent interface. E.g. Human::speak(), implemented on GrownUp, but not on Baby

benk-mira: Looks like my test passed because the parent wasn't a pure abstract class.  You're right, if the parent is abstract, you can't instantiate without all the abstract methods, so NotImplementedError is redundant here.

benk-mira: Removed

geoh5py/ui_json/forms.py
benk-mira: done.

geoh5py/ui_json/forms.py
benk-mira: done.

geoh5py/ui_json/enforcers.py
benk-mira: removed.

geoh5py/ui_json/parameters.py
```diff
@@ -33,31 +32,17 @@ class Parameter:
...
-        validation key value dictionary.
     """
 
     validations: dict[str, Any] = {}
 
-    def __init__(
-        self, name: str, value: Any = None, enforcers: EnforcerPool | None = None
-    ):
+    def __init__(self, name: str, value: Any = None):
         self.name: str = name
```
domfournier, line 40:
```suggestion
        self.name: str = name
        self._value: Any | None = None
```

geoh5py/ui_json/parameters.py
```diff
@@ -33,31 +32,17 @@ class Parameter:
...
-        self._enforcers: EnforcerPool = self._get_enforcer_pool(enforcers)
+        self._enforcers: EnforcerPool = EnforcerPool.from_validations(
+            self.name, self.validations
+        )
         setattr(self, "_value" if value is None else "value", value)
```
domfournier, line 44:
```suggestion
        if value is not None:
            self.value = value
```

domfournier (approved):
Small suggestion about the handling of parameter.value. Otherwise looks fantastic.

domfournier (approved):
Great - another milestone. Moving on.

PR #413, opened 2023-09-12
Simplify logic. Adjust tests to work on name AND type.name

domfournier:
This modification should handle both vertex and cell association. We might want to split the merge data into two methods (one for each), but the idea is there.

### DIFF
--- a/geoh5py/shared/merging/base.py
+++ b/geoh5py/shared/merging/base.py
@@ -60,7 +60,7 @@ class BaseMerger(ABC):
                 if label in data_dict:
                     values = data_dict[label].values
                     if isinstance(values, np.ndarray) and ~np.all(
-                        np.isnan(values[start:end])
+                        values[start:end] == data.nan_value
                     ):
                         label = (
                             data.name + f"({ind})",
--- a/geoh5py/shared/merging/base.py
+++ b/geoh5py/shared/merging/base.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from warnings import warn
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


MatthieuCMira:
There is no pytest on this pull request.
If this logic works, it would be great, but I'd rather be sure it's 100% safe before changing the code.

domfournier:
@sebhmg To address @MatthieuCMira comment, would you be against running the tests on PRs between branches?

tests/merger_test.py
```diff
@@ -81,7 +80,7 @@ def test_merge_point_data_unique_entity(tmp_path):
         data.append(
```
MatthieuCMira, line 80: The first test functio is not pertinent anymore if we force data to have same name and entity type. 
Remove this one

tests/merger_test.py
```diff
@@ -294,33 +269,16 @@ def test_merge_point_data_unique_entity_name_unique_name(tmp_path):
...
-                    }.items()
-                )
-            ).values()
-        )
-
-        np.testing.assert_almost_equal(
-            merged_data[0], np.hstack((nan_array, data[3].values))
-        )
-
         np.testing.assert_almost_equal(
```
MatthieuCMira, line 272: As we talk about, it's a bit of an assertion to consider the "first" data as the data to be merged. 
Would be safer to dispatch all the data in different data in this case no?

tests/merger_test.py
```diff
@@ -294,33 +269,16 @@ def test_merge_point_data_unique_entity_name_unique_name(tmp_path):
...
-            merged_data[1], np.hstack((data[0].values, nan_array))
```
MatthieuCMira, line 315: What if the data with similar entity_type and  name are in the second object (or the nth object?
)

sebhmg:
> @sebhmg To address @MatthieuCMira comment, would you be against running the tests on PRs between branches?

No problem. In practice PRs between intermediate branches do not happen often, thus running the tests will not cost to much against our quota. It might as well run the whole pre-commit checks. What do you think @domfournier ?

geoh5py/shared/merging/base.py
```diff
@@ -30,150 +29,73 @@ class BaseMerger(ABC):
...
-
-        # prepare the attributes
-        unique_entity_types = list(
-            dict.fromkeys(
-                [entity_type[:nb_keys] for entity_type in unique_entity_types]  # type: ignore
+                # Increment label in case of duplicate match
+                if label in data_dict:
+                    values = data_dict[label].values
+                    if isinstance(values, np.ndarray) and ~np.all(
+                        np.isnan(values[start:end])
```
MatthieuCMira, line 63: That's a good idea.
However, how to ensure the user want to merge his first data, not the second one.

I agree it's stupid, so we don't really have to care about it programmatically,  but that's a possible scenario. 
I would just pass a warning there "Just the first data encountered was added"

MatthieuCMira (commented):
I really like the overall logic, much more elegant. 

Just passing a warning in the case where data are "overlapping", precising the data merged might not be the one wanted.

MatthieuCMira:
merge dom's data merger solution.
Need to adapt some tests

PR #420, opened 2023-09-25
GEOPY-1093

benk-mira:
**GEOPY-1093 - Add RequiredWorkspaceObjectEnforcer to uijson-level validation pool.**


geoh5py/ui_json/enforcers.py
```diff
@@ -314,24 +315,16 @@ def enforce(self, name: str, value: Any):
...
+        return all(k in self.collection(value) for k in self.validations)
 
-    @validations.setter
-    def validations(self, val: str | list[str]):
-        if isinstance(val, str):
-            val = [val]
-        self._validations = val
+    def collection(self, value: Any) -> list[Any]:
+        """Returns the collection to check for required items."""
+        return value
```
domfournier, line 327: What's happening here, the `collection` doesn't seem to do anything?

geoh5py/ui_json/enforcers.py
```diff
@@ -314,24 +315,16 @@ def enforce(self, name: str, value: Any):
...
     def rule(self, value: Any) -> bool:
         """True if all required parameters are in 'value' collection."""
-        return all(k in value for k in self.validations)
-
-    @property
-    def validations(self) -> list[str]:
-        if self._validations is None:
-            self._validations = []
-        return self._validations
+        return all(k in self.collection(value) for k in self.validations)
```
domfournier, line 323: Can we get away without the collection? Seems backward compared to the other Enforcers, where the `rule` takes in the value to check. Here the value is used to validate? 

```suggestion
        return all(k in self.validations for k in value)

```

geoh5py/ui_json/enforcers.py
```diff
@@ -348,6 +341,14 @@ class RequiredWorkspaceObjectEnforcer(RequiredEnforcer):
     enforcer_type = "required_workspace_objects"
     validation_error = RequiredWorkspaceObjectValidationError
 
+    def rule(self, value: Any) -> bool:
+        """True if all objects are in the workspace."""
+        validations = [value[k].uid for k in self.validations]
+        return all(k in self.collection(value) for k in validations)
```
domfournier, line 347: Same comment as above, feels like we can do the reserve logic

```suggestion
        entities_list = self.validations[geoh5].list_entities_name
        return all(k in entities_list for k in value)

```

geoh5py/ui_json/forms.py
```diff
@@ -453,23 +493,34 @@ def __init__(self, name, value=None, **kwargs):
...
-        super().__init__(name, value=value, **kwargs)
+        super().__init__(name, value=value, data_type=data_type, **kwargs)
 
     @property
     def value(self):
         """Form value is value of property when is_value is False."""
-        val = self.property
-        if self.is_value:
+        val = self.property  # type: ignore # pylint: disable=no-member
+        if self.is_value:  # type: ignore # pylint: disable=no-member
```
domfournier, line 506: How does this work? `is_value` is a BoolParameter, so this would always pass.

geoh5py/ui_json/forms.py
```diff
@@ -453,23 +493,34 @@ def __init__(self, name, value=None, **kwargs):
...
         value = NumericParameter("value", value=value)
-        super().__init__(name, value=value, **kwargs)
+        super().__init__(name, value=value, data_type=data_type, **kwargs)
 
     @property
     def value(self):
         """Form value is value of property when is_value is False."""
-        val = self.property
-        if self.is_value:
+        val = self.property  # type: ignore # pylint: disable=no-member
```
domfournier, line 505: Indeed, this does seem to be defined on the class

geoh5py/ui_json/forms.py
```diff
@@ -453,23 +493,34 @@ def __init__(self, name, value=None, **kwargs):
...
     def value(self, val):
         if isinstance(val, (int, float)):
             self._value.value = val
-            self.is_value = True
         else:
-            self.property = val
-            self.is_value = False
+            self._property.value = val
+
+    def _data_type_string_to_type(self, data_type: str) -> type:
```
domfournier, line 517: This could be static, even better make it a constant. You will have to check that the key is valid wherever you are calling this.

geoh5py/ui_json/__init__.py
```diff
@@ -18,6 +18,67 @@
...
+    def update(self, data: dict[str, Any]):  # type: ignore
+        """Update keys and associated sets."""
+
+        for key, value in data.items():
+            if key in self.__dict__:
+                self.__dict__[key].update(self._make_set(value))
+            else:
+                self[key] = value
+
+    def _make_set(self, value) -> set:
```
domfournier, line 50: This looks like it could be static

geoh5py/ui_json/__init__.py
```diff
@@ -18,6 +18,67 @@
...
+from __future__ import annotations
+
+from typing import Any
+
 from .input_file import InputFile
 from .utils import monitored_directory_copy
 from .validation import InputValidation
+
+
+class SetDict(dict):
```
domfournier, line 30: Can we move this out of the __init__?

domfournier: Actually, when is it required to assert the unique set?

domfournier (commented):
Nice progress. See comments.

geoh5py/ui_json/enforcers.py
```diff
@@ -105,6 +110,8 @@ def _recruit_enforcer(enforcer_type: str, validation: Any) -> Enforcer:
             enforcer = ValueEnforcer(validation)  # type: ignore
         if enforcer_type == "uuid":
             enforcer = UUIDEnforcer(validation)  # type: ignore
+        if enforcer_type == "type_uid":
+            enforcer = TypeUIDEnforcer(validation)  # type: ignore
```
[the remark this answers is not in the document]
domfournier: Could maybe even just be a constant.

geoh5py/ui_json/__init__.py
```diff
@@ -18,6 +18,67 @@
...
+    def __iter__(self):
+        return iter(self.__dict__)
+
+    def keys(self):
+        return self.__dict__.keys()
+
+    def values(self):
+        return self.__dict__.values()
+
+    def items(self):
```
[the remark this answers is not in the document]
benk-mira: Looks like I took a wrong turn when I decided to push the overriden setitem values into the __dict__.  If i just update the dictionary and pass to super().__setitem__, I can simplify the class a lot!

geoh5py/ui_json/forms.py
benk-mira: The private attribute `self._is_value` is a BoolParameter, but I'm accessing the underlying value here with the public name through the ValueAccess descriptor.

geoh5py/ui_json/forms.py
benk-mira: Again, ValueAccess!

geoh5py/ui_json/forms.py
benk-mira: Done.  Note, I use .get with a None default to the lookup so I don't have to check for a valid key.

geoh5py/ui_json/__init__.py
benk-mira: Perhaps, but I wouldn't open this up unless there's a good reason.  This just invites coupling between classes.

benk-mira: ...If you think it's useful elsewhere I would pull it out as a utility instead

geoh5py/ui_json/__init__.py
benk-mira: I put it in shared/utils.py

benk-mira: > Actually, when is it required to assert the unique set?

Everywhere we set validations.  The goal is to not waste time generating multiple enforcers.

geoh5py/ui_json/enforcers.py
benk-mira: Just in the base class.  I use this in the children to transform the collection.  Eg RequiredWorkspaceObjectEnforcer takes the dictionary of param/value from uijson as it's value, but the collection that I'm interested in is the geoh5.list_entities_name.

geoh5py/ui_json/enforcers.py
benk-mira: The suggestion doesn't work since value is the full dictionary of uijson parameter/values.

geoh5py/ui_json/enforcers.py
benk-mira: self.validations doesn't contain a 'geoh5', in fact it's not even a dictionary, it's a set.

domfournier (approved):
Nice, thanks for the udpates.

PR #422, opened 2023-09-27
add a curve merge option and test.

MatthieuCMira:


geoh5py/shared/merging/curve.py
```diff
@@ -0,0 +1,57 @@
...
+from typing import cast
+
+import numpy as np
+
+from ...objects import Curve, ObjectBase
+from ...workspace import Workspace
+from .base import BaseMerger
+
+
+class CurveMerger(BaseMerger):
```
domfournier, line 29: Can make a CellMerger class, as both Curve and Surface can share the same mechanics.

geoh5py/shared/merging/curve.py
```diff
@@ -0,0 +1,57 @@
...
+    ) -> Curve:
+        # create the vertices
+        vertices = np.vstack([input_entity.vertices for input_entity in input_entities])
+
+        # merge the parts
+        parts: list = []
+        previous: int = 0
+
+        for entity in input_entities:
+            parts += (cast(Curve, entity).parts + previous).tolist()
```
domfournier, line 50: Good thoughts using parts, but I recommend using `cells` directly and incrementing the cell index by the n_vertices of the previous objects. 
For three reasons:

1- Parts need to be computed (never stored on geoh5), and it is not a super efficient process right now.
2- Parts is `n_vertices` long (assume a single connection), but it is possible for a vertex to belong to multiple cells (spyder web), in which case this implementation would lose those connections.
3- If we use `cells`, the same method can be used for Surfaces

geoh5py/shared/merging/curve.py
```diff
@@ -0,0 +1,57 @@
...
+    @classmethod
+    def validate_type(cls, input_entity):
+        # want to make sure that the input entities are Points, no subclasses
+        if type(input_entity) is not cls._type:  # pylint: disable=unidiomatic-typecheck
+            raise TypeError("The input entities must be a list of geoh5py Points.")
+
+    @classmethod
+    def create_object(
+        cls, workspace: Workspace, input_entities: list[ObjectBase], **kwargs
+    ) -> Curve:
```
domfournier, line 41: Add docstrings. You have the option to reference to docstrings of the parent class, but you should precise that the list are CellObject entities

geoh5py/ui_json/input_file.py
```diff
@@ -489,9 +489,9 @@ def numify(cls, ui_json: dict[str, Any]) -> dict[str, Any]:
                     cls.ui_validation(value)
                 except tuple(BaseValidationError.__subclasses__()) as error:
                     raise JSONParameterValidationError(key, error.args[0]) from error
-
                 value = cls.numify(value)
 
+            # todo: problem here: cannot get a path to a normal path
```
domfournier, line 494: Good idea. Created an issue to implement GEOPY-1103

tests/merger_curve_test.py
```diff
@@ -0,0 +1,113 @@
...
+import numpy as np
+
+from geoh5py.objects import Curve
+from geoh5py.shared.merging import CurveMerger
+from geoh5py.workspace import Workspace
+
+# import pytest
+
+
+def test_merge_curve(tmp_path):
```
domfournier, line 30: Even if you switch to `cells` merging, this test should still hold true.

domfournier (commented):
See comments

geoh5py/ui_json/input_file.py
MatthieuCMira: The problem was not about that, 
It was about having a ".geoh5" in my text but not at the end.

geoh5py/shared/merging/curve.py
MatthieuCMira: Did it in the cells

domfournier (approved):
Looks great.

PR #423, opened 2023-09-28
GEOPY-1106 - pathlib in ui.json utils

MatthieuCMira:
**GEOPY-1106 - path2workspace, use Pathlib Path suffix**
use pathlib to get suffix instead of strings in path2workspace in ui.json utils


domfournier (approved):
Great.

domfournier (approved):

PR #426, opened 2023-10-02
Geopy 1045

MatthieuCMira:
**GEOPY-1045 - Merge DrapeModel objects**
merge drape model


geoh5py/shared/merging/base.py
```diff
@@ -104,10 +109,14 @@ def merge_data(cls, out_entity, input_entities: list[ObjectBase]):
                     data_dict[label].values = values
 
             data_count["VERTEX"] += (
-                input_entity.n_vertices if input_entity.n_vertices is not None else 0
+                input_entity.n_vertices + cls._ghost
```
domfournier, line 112: I would vote that we keep the ghost attribute to the DrapeMerger only, as it is only used specifically there. Consider simply re-implementing merge_data under the class

```
def merge_data(cls, out_entity, input_entities: list[ObjectBase]):
   - call super().merge_data(out_entity, input_entities)
   - Loop over the children and shift the values to account for the ghost cells
```

geoh5py/shared/merging/drape_model.py
```diff
@@ -0,0 +1,145 @@
...
+        ghost_prism[4] = 1
+
+        ghost_layer[0] = previous_layer
+        ghost_layer[1] = 0
+        ghost_layer[2] = ghost_prism[2]
+
+        return np.expand_dims(ghost_prism, 0), np.expand_dims(ghost_layer, 0)
+
+    @classmethod
+    def validate_structure(cls, input_entity: ObjectBase):
```
domfournier, line 135: You could type it directly as DrapeModel and avoid casting below.

geoh5py/shared/merging/base.py
```diff
@@ -178,10 +187,17 @@ def validate_objects(cls, input_entities: list[ObjectBase]):
...
+            cls.validate_structure(input_entity)
+
+    @classmethod
+    def validate_structure(cls, input_entity: ObjectBase):
+        """
+        Validate the input entity structure and raises error if incompatible.
+        :param input_entity: the input entity to validate.
+        """
+        # verify if the input entity have vertices
+        if not isinstance(input_entity.vertices, np.ndarray):
```
domfournier, line 200: I guess this doesn't much sense anymore to be checking for vertices on the base class. Consider making validate_structure a `@abstractclass` to force the implementation on the sub class.

domfournier (commented):
See comments about methods.

geoh5py/shared/merging/base.py
MatthieuCMira: That's what I did in the first place.
.
I agree it's not useful in BaseMerger, but it's just a tiny +2 somewhere. 

option 1) It seems to me that calling super() and rerunning a second loop to insert nan values between the previous values is not efficient as it use a second loop and also introduces a lot of code. 

option 2) Just copying the full function and adding +2  introduces a lot of duplicated code too. 

option 3) But having a class attribute just added 4 lines of code in total, without changing the performances...


I'm letting you choose the option in the comment and correct the code the way you want.

geoh5py/shared/merging/drape_model.py
MatthieuCMira: If I change the type, pylint doesn't like it as it's not the same type as the parent method 'validate_structure'.

Or maybe I misunderstood your comment?

geoh5py/shared/merging/base.py
MatthieuCMira: Ok, I did that and inherited cells from points

geoh5py/shared/merging/base.py
domfournier: I am not too worried about the efficiency part, since the number of child will be small, and re-assigning values is pretty fast. I would rather the base class to be simple and not contain special cases

geoh5py/shared/merging/drape_model.py
domfournier: If validate_structure is abstract, then you can probably not type there, and only type the subclasses?

geoh5py/shared/merging/base.py
MatthieuCMira: Added a new loop using super to redefined merge_data() in DrapeModel

geoh5py/shared/merging/drape_model.py
```diff
@@ -0,0 +1,180 @@
...
+                n_cells = cast(int, input_entity.n_cells)
+                values[previous + twos : previous + twos + n_cells] = data.values[
+                    previous : previous + n_cells
+                ]
+
+                previous += n_cells
+                twos += 2
+
+            # update the values
+            data.values = values
```
domfournier, line 167: You could just create an index map once, and apply on all data. Will be faster than doing to nested loops.

geoh5py/shared/merging/drape_model.py
```diff
@@ -0,0 +1,171 @@
...
+
+class DrapeModelMerger(BaseMerger):
+    _type: type = DrapeModel
+
+    @classmethod
+    def create_object(
+        cls, workspace: Workspace, input_entities: list[ObjectBase], **kwargs
+    ) -> DrapeModel:
+        """
+        Create a new object of type cls._type from a list of input entities.
```
domfournier, line 38:
```suggestion
        Create a new DrapeModel from a list of input DrapeModels.
        
```

geoh5py/shared/merging/drape_model.py
```diff
@@ -0,0 +1,171 @@
...
+        ghost_layer: np.ndarray = np.array([])
+
+        for input_entity in input_entities:
+            # get the values of the entity
+            temp_prisms: np.ndarray = cast(
+                np.ndarray, cast(DrapeModel, input_entity).prisms
+            )
+            temp_layers: np.ndarray = cast(
+                np.ndarray, cast(DrapeModel, input_entity).layers
+            )
```
domfournier, line 62:
```suggestion
            temp_prisms: np.ndarray = input_entity.prisms
            temp_layers: np.ndarray = input_entity.layers
```

geoh5py/shared/merging/drape_model.py
```diff
@@ -0,0 +1,171 @@
...
+
+        ind_map: list = []
+        data_count: int = 0
+        n_values: int = out_entity.n_cells - (len(input_entities) - 1) * 2
+        for ghost, input_entity in enumerate(input_entities):
+            if input_entity.n_cells is None:
+                continue
+            ind_map += [
+                np.arange(data_count, data_count + input_entity.n_cells),
+                [n_values + ghost * 2, n_values + ghost * 2],
```
domfournier, line 150: Same value twice? Aren't we missing a +1 somewhere? I think it will end up being the same (all ndv anyway), but not really the parsing intended.

domfournier (commented):
A couple more edits and we are good to go.

geoh5py/shared/merging/drape_model.py
```diff
@@ -0,0 +1,171 @@
...
+from ...workspace import Workspace
+from .base import BaseMerger
+
+
+class DrapeModelMerger(BaseMerger):
+    _type: type = DrapeModel
+
+    @classmethod
+    def create_object(
+        cls, workspace: Workspace, input_entities: list[ObjectBase], **kwargs
```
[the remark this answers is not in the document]
MatthieuCMira: Pylint doesn't gonna like it...

geoh5py/shared/merging/drape_model.py
MatthieuCMira: I agree, but can I shut pylint when it acts stupid?

geoh5py/shared/merging/drape_model.py
domfournier: Very efficient - love it.

MatthieuCMira: you are right

geoh5py/shared/merging/drape_model.py
```diff
@@ -32,7 +32,7 @@ class DrapeModelMerger(BaseMerger):
 
     @classmethod
     def create_object(
-        cls, workspace: Workspace, input_entities: list[DrapeModel], **kwargs
+        cls, workspace: Workspace, input_entities: list[DrapeModel], **kwargs  # type: ignore
```
domfournier, line 35: Remove the type in the abstract instead.

geoh5py/shared/merging/drape_model.py
```diff
@@ -0,0 +1,171 @@
...
+                ghost_prism, ghost_layer = cls._ghost_point(
+                    temp_prisms[0],
+                    temp_prisms[1],
+                    previous_prism - 1,
+                    previous_layer - 1,
+                )
+
+                # append first ghost prism and layer
+                prisms.append(ghost_prism)
+                layers.append(ghost_layer)
```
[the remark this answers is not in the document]
domfournier: I see, it adds two backward... great.

domfournier (dismissed):

domfournier (approved):

PR #427, opened 2023-10-02
GEOPY-1092

MatthieuCMira:
**GEOPY-1092 - create a property_group.collect_values()**
add a property group values collector

### DIFF
--- a/geoh5py/groups/property_group.py
+++ b/geoh5py/groups/property_group.py
@@ -22,6 +22,8 @@ from abc import ABC
 from collections.abc import Iterable
 from typing import TYPE_CHECKING
 
+import numpy as np
+
 from geoh5py.data import Data, DataAssociationEnum
 
 if TYPE_CHECKING:
@@ -136,6 +138,19 @@ class PropertyGroup(ABC):
         :obj:`dict` Attribute names mapping between geoh5 and geoh5py
         """
         return self._attribute_map
+
+    @property
+    def values(self) -> np.ndarray | None:
+        """
+        The values of the properties in the group.
+        """
+
+        if self._properties is None:
+            return None
+
+        return np.r_[
+            [self._parent.get_data(data)[0].values for data in self._properties]
+        ]
 
     @property
     def name(self) -> str:
--- a/geoh5py/groups/property_group.py
+++ b/geoh5py/groups/property_group.py
@@ -140,7 +140,7 @@ class PropertyGroup(ABC):
         return self._attribute_map
 
     @property
-    def values(self) -> np.ndarray | None:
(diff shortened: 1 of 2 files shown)


geoh5py/groups/property_group.py
```diff
@@ -137,6 +139,19 @@ def attribute_map(self) -> dict:
...
+    def collect_values(self) -> np.ndarray | None:
+        """
+        The values of the properties in the group.
+        """
+
+        if self._properties is None:
+            return None
+
+        return np.r_[
+            [self._parent.get_data(data)[0].values for data in self._properties]
```
domfournier, line 152: I would vote that we keep it as a list here, and let the user decide how to parse the data. Otherwise, we are sort of losing the partition of each child data...

domfournier (commented):
Suggestion below.

geoh5py/groups/property_group.py
MatthieuCMira: We are not losing the partition.
The output shape is (len(children), len(nvertices))
Having a list doesn't change anything, even if we loop over it?

I'm okay passing it as a list anyway.

domfournier: `np.r_` should technically give you a long vector shape (n_child * n_vertices, ). You meant np.c_? Let's do a list anyway, since it will remain a pointer to the values of the each entity. User can decide if they want to duplicate the memory.

geoh5py/groups/property_group.py
```diff
@@ -137,6 +139,17 @@ def attribute_map(self) -> dict:
         """
         return self._attribute_map
 
+    @property
+    def collect_values(self) -> np.ndarray | None:
```
domfournier, line 143:
```suggestion
    def collect_values(self) -> list | None:
```

domfournier (dismissed):
Minor typing fix. Got to go otherwise

geoh5py/groups/property_group.py
```diff
@@ -22,6 +22,8 @@
 from collections.abc import Iterable
 from typing import TYPE_CHECKING
 
+import numpy as np
+
 from geoh5py.data import Data, DataAssociationEnum
```
domfournier, line 27:
```suggestion
```

domfournier (approved):

geoh5py/groups/property_group.py
```diff
@@ -22,6 +22,8 @@
 from collections.abc import Iterable
 from typing import TYPE_CHECKING
 
+import numpy as np
+
```
domfournier:
```suggestion
```

domfournier (approved):

PR #428, opened 2023-10-02
GEOPY-444: print replaced by logger

MatthieuCMira:
**GEOPY-444 - Use logging instead of print**
Should we change warning to logger.warning?

Change from sprint 63 to sprint 62. I don't have the access

### DIFF
--- a/geoh5py/shared/__init__.py
+++ b/geoh5py/shared/__init__.py
@@ -17,8 +17,6 @@
 
 # pylint: disable=unused-import
 # flake8: noqa
-import logging
-
 from . import weakref_utils
 from .entity import Entity
 from .entity_type import EntityType
@@ -26,5 +24,3 @@ from .utils import fetch_h5_handle, match_values, merge_arrays
 
 INTEGER_NDV = -2147483648
 FLOAT_NDV = 1.17549435e-38
-
-logger = logging.getLogger(__name__)
--- a/geoh5py/shared/__init__.py
+++ b/geoh5py/shared/__init__.py
@@ -17,6 +17,7 @@
 
 # pylint: disable=unused-import
 # flake8: noqa
+import logging
 
 from . import weakref_utils
 from .entity import Entity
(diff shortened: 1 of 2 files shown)


domfournier (approved):
Simple. Looks good.

domfournier:
@sebhmg We are not using `print` a lot at this point. Did you have something bigger in mind?

geoh5py/shared/__init__.py
```diff
@@ -25,3 +26,5 @@
 
 INTEGER_NDV = -2147483648
 FLOAT_NDV = 1.17549435e-38
+
+logger = logging.getLogger(__name__)
```
sebhmg, line 30: **suggestion:** create logger in each module

This will create a logger for this specific module "shared", used everywhere. this is not desirable.
One benefit of using logger is that it accept a global configuration where the behaviour can be tweaked for each module.

A client application can configure the logging, and turn off log from some modules, or change the logging level per module, or redirect, reformat, etc...

From [advanced-logging-tutorial](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial):

> A good convention to use when naming loggers is to use a module-level logger, in each module which uses logging, named as follows:
> ```python
> logger = logging.getLogger(__name__)
> ```
> This means that logger names track the package/module hierarchy, and it’s intuitively obvious where events are logged just from the logger name.

sebhmg (changes requested):
recommendation is to create a logger per module

geoh5py/shared/__init__.py
MatthieuCMira: Should you want to have a common logger for all the 'object'?
Or a logger for every specific module using 'print'?

sebhmg: the recommendation is to have a logger per module.
So each file will call its own: `logger = logging.getLogger(__name__)`

It is safe to call it multiple times in a file: it returns the same logger instance.
It can also be placed once the top of each file that needs it, after imports

domfournier (approved):
Grrreat 
image

domfournier (approved):

sebhmg (approved):

PR #429, opened 2023-10-03
GEOPY-1045b: Build index map

domfournier:
**GEOPY-1045 - Merge DrapeModel objects**

### DIFF
--- a/geoh5py/shared/merging/drape_model.py
+++ b/geoh5py/shared/merging/drape_model.py
@@ -139,32 +139,30 @@ class DrapeModelMerger(BaseMerger):
     ):
         super().merge_data(out_entity, input_entities)
 
-        # prepare an empty nan array
-        values: np.ndarray = np.empty(out_entity.n_cells)
-        values[:] = np.nan
+        ind_map = []
+        data_count = 0
+        ghost_count = 0
+        n_values = np.sum([input_entity.n_cells for input_entity in input_entities])
+        for input_entity in input_entities:
+            if input_entity.n_cells is None:
+                continue
 
+            n_cells = input_entity.n_cells
+            ind_map.append(np.arange(data_count, data_count + n_cells))
+
+            if input_entity != input_entities[-1]:
+                ind_map.append([n_values + ghost_count, n_values + ghost_count + 1])
+
+            ghost_count += 2
+            data_count += n_cells
+
+        ind_map = np.hstack(ind_map)
         # get all the values in the output entity
         for data in out_entity.children:
-            if (
-                not isinstance(data, NumericData)
-                or data.association is None
-                or data.n_values is None
-                or data.values is None
-            ):
+            if not isinstance(data, NumericData) or data.values is None:
                 continue
-            previous: int = 0
-            twos: int = 0
-            for input_entity in input_entities:
-                n_cells = cast(int, input_entity.n_cells)
-                values[previous + twos : previous + twos + n_cells] = data.values[
-                    previous : previous + n_cells
-                ]
 
-                previous += n_cells
-                twos += 2
-
-            # update the values
-            data.values = values
+            data.values = data.values[ind_map]
 
     @classmethod
     def validate_structure(cls, input_entity: DrapeModel):


geoh5py/shared/merging/drape_model.py
```diff
@@ -139,32 +139,30 @@ def merge_data(
...
 
-        # prepare an empty nan array
-        values: np.ndarray = np.empty(out_entity.n_cells)
-        values[:] = np.nan
+        ind_map = []
+        data_count = 0
+        ghost_count = 0
+        n_values = np.sum([input_entity.n_cells for input_entity in input_entities])
+        for input_entity in input_entities:
+            if input_entity.n_cells is None:
```
MatthieuCMira, line 147: Is this possible? We check for prisms and layers to exist for input entities, 
except for pylint, we don't need that, doesn't we? cast does the job.

geoh5py/shared/merging/drape_model.py
```diff
@@ -139,32 +139,30 @@ def merge_data(
     ):
         super().merge_data(out_entity, input_entities)
 
-        # prepare an empty nan array
-        values: np.ndarray = np.empty(out_entity.n_cells)
-        values[:] = np.nan
+        ind_map = []
+        data_count = 0
+        ghost_count = 0
+        n_values = np.sum([input_entity.n_cells for input_entity in input_entities])
```
MatthieuCMira, line 145: What about :

`n_values = output_entity.n_cells - (len(input_entities)-1)*2`?
more efficient than a loop no?

geoh5py/shared/merging/drape_model.py
```diff
@@ -139,32 +139,30 @@ def merge_data(
...
+        for input_entity in input_entities:
+            if input_entity.n_cells is None:
+                continue
+
+            n_cells = input_entity.n_cells
```
MatthieuCMira, line 150: Seems a lot of code lines . 
What about putting 

`ind_map.append([n_values + ghost_count, n_values + ghost_count + 1])`at the beginning of the loop.

initiate `ghost_count`to -2 or even better, use enumerate instead (ghost then `(i-1)*2`).

replace `np.hstack(ind_map)` by `np.hstack(ind_map[1:])`

saving 2 lines and an if condition. 

Sure it's not a big deal but made the code prettier.

MatthieuCMira (commented):
very minor comments to make the code lighter

geoh5py/shared/merging/drape_model.py
domfournier: It is possible that one of the input object doesn't have cell defined.

geoh5py/shared/merging/drape_model.py
domfournier: I am not getting the math here. `n_cells` is all the combined cells

geoh5py/shared/merging/drape_model.py
MatthieuCMira: If so, it raises an error at the beginning no?
It checks for every object to have prisms and layers, which are associated to cells
Except if prisms and layers are empty array? I don't even know if it's possible.

geoh5py/shared/merging/drape_model.py
domfournier: Maybe? Fine by me. Maybe better would be to do it on your PR once this is merged in.

geoh5py/shared/merging/drape_model.py
MatthieuCMira: yep, and `output.n_cells = sum(*inputs.n_cells) + (len(inputs)-1)*2

domfournier: ok sure yep, both do the same. I think it's a bit harder to read, but if you prefer that.

geoh5py/shared/merging/drape_model.py
domfournier: Correct, technically it is checked in during the super(). And yes, it is raised by pylint since n_cells can be None.

PR #436, opened 2023-10-11
GEOPY-1102 corrections

MatthieuCMira:
**GEOPY-1102 - merge curves**
add a workspace setter in entity_type

change workspace of the entity_type

close the initial workspace in the tests

### DIFF
--- a/geoh5py/shared/entity_type.py
+++ b/geoh5py/shared/entity_type.py
@@ -132,3 +132,9 @@ class EntityType(ABC):
         # which means workspace has been deleted.
         assert workspace is not None
         return workspace
+
+    @workspace.setter
+    def workspace(self, workspace: ws.Workspace):
+        assert workspace is not None
+        self._workspace = weakref.ref(workspace)
+        self.workspace.update_attribute(self, "attributes")
--- a/geoh5py/shared/merging/base.py
+++ b/geoh5py/shared/merging/base.py
@@ -91,12 +91,17 @@ class BaseMerger(ABC):
                         if association == "VERTEX"
                         else out_entity.n_cells
                     )
+
+                    # update the entity type
+                    entity_type = data.entity_type
+                    entity_type.workspace = out_entity.workspace
+
                     data_dict[label] = out_entity.add_data(
                         {
                             data.name: {
                                 "values": np.ones(shape) * data.nan_value,
                                 "association": association,
-                                "entity_type": data.entity_type,
+                                "entity_type": entity_type,
                             }
                         }
                     )
--- a/tests/merger_drape_model_test.py
+++ b/tests/merger_drape_model_test.py
@@ -74,6 +74,8 @@ def test_merge_drape_model(tmp_path):  # pylint: disable=too-many-locals
             test_prisms.append(drape_model.prisms)
             test_layers.append(drape_model.layers)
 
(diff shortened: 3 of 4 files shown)


geoh5py/shared/entity_type.py
```diff
@@ -132,3 +132,9 @@ def workspace(self) -> ws.Workspace:
         # which means workspace has been deleted.
         assert workspace is not None
         return workspace
+
+    @workspace.setter
```
domfournier, line 136: We don't want a setter on the workspace as it should not be changed by the user. The responsibility is on the workspace to manage the unique entity_types

tests/merger_drape_model_test.py
```diff
@@ -74,6 +74,8 @@ def test_merge_drape_model(tmp_path):  # pylint: disable=too-many-locals
             test_prisms.append(drape_model.prisms)
             test_layers.append(drape_model.layers)
 
+    h5file_path_2 = tmp_path / "drapedmodel2.geoh5"
+    with workspace.create(h5file_path_2) as workspace:
         drape_model_merged = DrapeModelMerger.merge_objects(
             workspace, drape_models, name="merged", children=[Points(workspace)]
```
domfournier, line 80: I just noticed this. Points is a child of drape model? Doens't make too much sense.

domfournier (commented):
We need to rethink the fix

geoh5py/shared/merging/base.py
```diff
@@ -91,12 +91,17 @@ def merge_data(
                         if association == "VERTEX"
                         else out_entity.n_cells
                     )
+
+                    # update the entity type
+                    entity_type = data.entity_type
+                    entity_type.workspace = out_entity.workspace
```
domfournier, line 97: We can't do that as it will change the workspace of the original data. 
Consider making a copy the entity_type in the new workspace instead.

tests/merger_drape_model_test.py
MatthieuCMira: That's just a trivial test to augment coverage.

When iterating over children => get the data 
else continue <= covering this part of the code.

geoh5py/shared/entity_type.py
MatthieuCMira: But as workspace is weakref, once it's closed, it's None.
We need to copy the entity to the new workspace! 
at least create a copy function for entity_type?

geoh5py/shared/merging/base.py
MatthieuCMira: Refer to my first answer: what about creating a copy function then?

tests/merger_drape_model_test.py
domfournier: gotcha

geoh5py/shared/merging/base.py
domfournier: Agree, a copy method on entity_type would work.

geoh5py/shared/entity_type.py
domfournier: Yep that would be nicer. I would also need a check in the data creation to make sure that the workspace of entity_type is the same as parent object, else - call the copy method.

MatthieuCMira: Created a copy function in entity_type

PR #437, opened 2023-10-11
Geopy 1102 corr2

MatthieuCMira:
**GEOPY-1102 - merge curves**
add one object merger

### DIFF
--- a/geoh5py/shared/entity_type.py
+++ b/geoh5py/shared/entity_type.py
@@ -133,8 +133,17 @@ class EntityType(ABC):
         assert workspace is not None
         return workspace
 
-    @workspace.setter
-    def workspace(self, workspace: ws.Workspace):
-        assert workspace is not None
-        self._workspace = weakref.ref(workspace)
-        self.workspace.update_attribute(self, "attributes")
+    def copy(self, workspace: ws.Workspace) -> EntityType:
+        """Copy this entity type to another workspace."""
+
+        attributes = {
+            prop: getattr(self, prop)
+            for prop in dir(self)
+            if isinstance(getattr(self.__class__, prop, None), property)
+            and prop not in ["workspace", "uid"]
+            and getattr(self, prop) is not None
+        }
+
+        attributes.update({"workspace": workspace})
+
+        return self.__class__(**attributes)
--- a/geoh5py/shared/entity_type.py
+++ b/geoh5py/shared/entity_type.py
@@ -133,17 +133,21 @@ class EntityType(ABC):
         assert workspace is not None
         return workspace
 
-    def copy(self, workspace: ws.Workspace) -> EntityType:
+    def copy(self, **kwargs) -> EntityType:
         """Copy this entity type to another workspace."""
 
         attributes = {
             prop: getattr(self, prop)
             for prop in dir(self)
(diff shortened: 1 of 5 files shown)


geoh5py/shared/merging/base.py
```diff
@@ -91,12 +91,15 @@ def merge_data(
...
                     )
+
                     data_dict[label] = out_entity.add_data(
                         {
                             data.name: {
                                 "values": np.ones(shape) * data.nan_value,
                                 "association": association,
-                                "entity_type": data.entity_type,
+                                "entity_type": data.entity_type.copy(
+                                    out_entity.workspace
```
domfournier, line 101: Any chance we can move this to the `object_base.validate_data_type`. That way we can catch more possible issues with people adding data and sharing entity_type. Just need to add a little logic to check if the workspace differ.

geoh5py/shared/entity_type.py
```diff
@@ -132,3 +132,18 @@ def workspace(self) -> ws.Workspace:
...
         return workspace
+
+    def copy(self, workspace: ws.Workspace) -> EntityType:
+        """Copy this entity type to another workspace."""
+
+        attributes = {
+            prop: getattr(self, prop)
+            for prop in dir(self)
+            if isinstance(getattr(self.__class__, prop, None), property)
+            and prop not in ["workspace", "uid"]
```
domfournier, line 143: Should only ignore the uid if it's in the same workspace, otherwise keep it.
```suggestion
```

domfournier (commented):
Almost there.

geoh5py/shared/entity_type.py
MatthieuCMira: If we copy an object, don't we want to let the uid being automatically defined?
We do not need to have the same uid, because it's a new object anyway.

geoh5py/shared/merging/base.py
MatthieuCMira: We can't check `entity_type.workspace` 
If we try to access it with the property, we'll raise an assertion error (see my previous comment).
So: 
- or we remove this assertion error in `entity_type.workspace`
- or we access with `getattr()`

Could be another issue, as this option is not concerning the merging but a more global issue?

geoh5py/shared/entity_type.py
```diff
@@ -132,3 +132,18 @@ def workspace(self) -> ws.Workspace:
...
+            and getattr(self, prop) is not None
+        }
+
+        attributes.update({"workspace": workspace})
+
```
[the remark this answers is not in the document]
MatthieuCMira: see previous answer

geoh5py/shared/entity_type.py
domfournier: Not really if the copy is going to a different workspace. Then GA would be able to recognize they are the same

geoh5py/shared/merging/base.py
domfournier: Don't see how that's possible. Maybe we should talk

domfournier (approved):
Will need to make a public access to `_types` at some point. But beyond this issue.

Thanks for the changes.

PR #441, opened 2023-10-24
GEOPY-1148: bug with grid.object.copy()

MatthieuCMira:
**GEOPY-1148 - Geoh5py. GridObject.copy() => if PropertyGroup in children it raises an error**
In grid_object

Bug when copying grid_object containing a property group. Since propertygroups are contained in the children.

=> catch property groups with continue

### DIFF
--- a/geoh5py/objects/grid_object.py
+++ b/geoh5py/objects/grid_object.py
@@ -56,7 +56,6 @@ class GridObject(ObjectBase, ABC):
         copy_children: bool = True,
         clear_cache: bool = False,
         mask: np.ndarray | None = None,
-        copy_property_groups: bool = True,
         **kwargs,
     ):
         """
@@ -66,7 +65,6 @@ class GridObject(ObjectBase, ABC):
         :param copy_children: Copy children entities.
         :param clear_cache: Clear cache of data values.
         :param mask: Array of indices to sub-sample the input entity.
-        :param copy_property_groups: Copy property groups.
         :param kwargs: Additional keyword arguments.
 
         :return: New copy of the input entity.
@@ -110,7 +108,7 @@ class GridObject(ObjectBase, ABC):
                 )
                 children_map[child.uid] = child_copy.uid
 
-            if copy_property_groups and self.property_groups:
+            if self.property_groups:
                 self.workspace.copy_property_groups(
                     new_entity, self.property_groups, children_map
                 )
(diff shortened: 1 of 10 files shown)


MatthieuCMira:
monitoring_directory_copy() now copy the property group.
However, unlike data, it's been copied on the workspace, which leads to duplicate the propertygroups

MatthieuCMira:
I introduce an option in grid_object copy() which is copy_property_group.
It's an option now in monitoring directoy_copy.
If call "False", the porperty groups are not copied

geoh5py/objects/grid_object.py
```diff
@@ -55,6 +56,7 @@ def copy(
         copy_children: bool = True,
         clear_cache: bool = False,
         mask: np.ndarray | None = None,
+        copy_property_groups: bool = True,
```
domfournier, line 59: Let's not just modify this class. This would technically have to be pushed to the base class.

geoh5py/ui_json/utils.py
```diff
@@ -341,7 +345,11 @@ def monitored_directory_copy(
 
     with fetch_active_workspace(entity.workspace, mode="r"):
         with Workspace.create(working_path / temp_geoh5) as w_s:
-            entity.copy(parent=w_s, copy_children=copy_children)
+            entity.copy(
+                parent=w_s,
+                copy_children=copy_children,
+                copy_property_groups=copy_property_groups,
```
domfournier, line 351: Changing here will break the copy for all other entities that don't have the copy_property_groups, or don't 
have property_groups. Let's not do that now and focus on fixing the grid2d instead.

domfournier (changes requested):
Try to focus on the issue. Adding an option to a core method like `copy` is a much larger endeavour.

domfournier:
It is likely that we need to also check the mechanics in grid2d. Since `property_groups` don't have `values`, it will likely crash here
image

MatthieuCMira:
> e need to also check the mechanics in grid2d. Since `property_groups` don't have `values`, it will likely crash her

there is no copy in grid2d, it's in grid object.

MatthieuCMira:
The problem is not about the copy. It's working.

The problem is about copying property_groups.
When I "monitored_directory_copy"
It duplicates the property_group in the workspace.

We just have to correct this

domfournier:
> > e need to also check the mechanics in grid2d. Since `property_groups` don't have `values`, it will likely crash her
> 
> there is no copy in grid2d, it's in grid object.

Correct, but the `copy_from_extent` will likely break as well. Good time to fix this bug since related to the copy method. We can augment the test you wrote to use the `copy_from_extent`

domfournier (approved):
Fantastic, thanks for this.

MatthieuCMira:
pre-commit mypy is annoying about typing with subclass/superclass.

domfournier (approved):
Go for testintg.

PR #442, opened 2023-10-26
GEOPY-1127

domfournier:
**GEOPY-1127 - Fix warnings about group optional**


benk-mira (changes requested):
tests failing

benk-mira (approved):
Great!  Seems like the tests show the behaviour remains the same.

PR #453, opened 2023-11-08
GEOPY-1094

benk-mira:
**GEOPY-1094 - Add RequiredObjectDataEnforcer to uijson-level enforcer pool.**


geoh5py/ui_json/input_file.py
```diff
@@ -303,6 +304,7 @@ def validation_options(self):
             return {
                 "update_enabled": True,
                 "ignore_list": (),
+                "disable_promotion": False,
```
domfournier, line 307: Since we are going to turn off validations on input file pretty soon, can we somehow make this an attribute of the class?

domfournier (commented):
Looks reasonable. See suggestion about future implementation

geoh5py/ui_json/input_file.py
benk-mira: done

domfournier (approved):
Looks good. Moving on.

PR #457, opened 2023-11-09
GEOPY-1144

benk-mira:
**GEOPY-1144 - Create property group incrementer for add_data calls.**


geoh5py/shared/concatenation.py
```diff
@@ -1087,26 +1085,29 @@ def validate_depth_data(
...
-                    name=property_group,
-                    association="DEPTH",
-                    property_group_type="Depth table",
+            name_id = 0
+            while True:
+                out_group: ConcatenatedPropertyGroup = (
+                    self.find_or_create_property_group(  # type: ignore
+                        name=property_group
+                        if name_id == 0
+                        else f"{property_group} ({name_id})",
```
domfournier, line 1094: Good idea, but this won't work if the function gets called multiple time. Would yield "name (1) (1)" and so on. Will probably need a slightly more complex regex function...

domfournier (commented):
Good idea on loopinp the `find_create_prop_group`. See comment about name change

geoh5py/shared/concatenation.py
benk-mira: Only if they pass the string "name (1)" instead of "name" the second time..  I added another add_data call with the same property string to the unit test to confirm: If you keep passing "name" into it you'll get "name (1)", "name (2)", etc..  Which is what our importer will do since we give the name as a string in the ui.json and stick with it for all the imports.

domfournier: Yep, you are right. Thanks.

domfournier (approved):
Looking great.

PR #460, opened 2023-11-20
GEOPY-1217

MatthieuCMira:
Drillhole Text. management

Problem with pylint and mypy. Just shut them?

### DIFF
--- a/geoh5py/objects/drillhole.py
+++ b/geoh5py/objects/drillhole.py
@@ -408,7 +408,10 @@ class Drillhole(Points):
                 )
 
             attributes, new_property_group = self.validate_data(
-                attributes, property_group, collocation_distance=collocation_distance
+                attributes,
+                property_group,
+                collocation_distance=collocation_distance,
+                type_=attributes.get("type"),
             )
 
             entity_type = self.validate_data_type(attributes)
@@ -495,7 +498,8 @@ class Drillhole(Points):
         from_to: np.ndarray | list,
         values: np.ndarray,
         collocation_distance: float = 1e-4,
-    ):
+        type_: str | None = None,
+    ):  # pylint: disable=too-many-locals
         """
         Compare new and current depth values, append new vertices if necessary and return
         an augmented values vector that matches the vertices indexing.
@@ -558,9 +562,15 @@ class Drillhole(Points):
                 from_to.flatten()[ind_new], return_inverse=True
             )
 
+            # check if its text data, and defined nan array if so
+            if type_ == "TEXT":
+                nan_values = np.array([""] * self.n_cells)  # type: ignore
+            else:
+                nan_values = np.ones(self.n_cells) * np.nan
+
             # Append values
             values = merge_arrays(
-                np.ones(self.n_cells) * np.nan,
+                nan_values,
                 np.r_[values],
                 replace="B->A",
                 mapping=cell_map,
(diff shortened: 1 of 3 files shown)


geoh5py/objects/drillhole.py
```diff
@@ -408,7 +408,10 @@ def add_data(
                 )
 
             attributes, new_property_group = self.validate_data(
-                attributes, property_group, collocation_distance=collocation_distance
+                attributes,
+                property_group,
+                collocation_distance=collocation_distance,
+                type_=attributes.get("type"),
```
domfournier, line 414: Can't we check the type directly in the collocation_distance method?

geoh5py/shared/concatenation.py
```diff
@@ -960,7 +960,11 @@ def to_(self) -> list[Data]:
         return obj_list
 
     def validate_data(
-        self, attributes: dict, property_group=None, collocation_distance=None
+        self,
+        attributes: dict,
+        property_group=None,
+        collocation_distance=None,
+        **_,
```
domfournier, line 967: Let's not do that.

geoh5py/shared/concatenation.py
```diff
@@ -988,6 +992,12 @@ def validate_data(
                 )
             attributes["from-to"] = None
 
+        # set a specific nan value if text
+        if attributes.get("type") == "TEXT":
```
domfournier, line 996: Yep

domfournier (commented):
See comment.

geoh5py/objects/drillhole.py
MatthieuCMira: collocation_distance isn't a float?
type is a string....
What if both are passed?

geoh5py/data/text_data.py
```diff
@@ -39,20 +39,32 @@ def values(self) -> np.ndarray | str | None:
...
-        self._values = values
+        if isinstance(values, bytes):
+            values = values.decode()
 
-        if not isinstance(values, (np.ndarray, str, type(None))):
+        if isinstance(values, np.ndarray) and values.dtype == object:
+            values = values.astype(str)
+
+        if (not isinstance(values, (str, type(None), np.ndarray))) or (
+            isinstance(values, np.ndarray) and values.dtype.kind not in ["U", "S"]
```
domfournier, line 60: I had never seen this before, but makes sense. When/where are you encountering bad kinds?

geoh5py/objects/drillhole.py
```diff
@@ -490,15 +490,23 @@ def add_vertices(self, xyz):
...
         return indices.astype("uint32")
 
-    def validate_interval_data(
+    def validate_interval_data(  # pylint: disable=too-many-locals
         self,
         from_to: np.ndarray | list,
         values: np.ndarray,
         collocation_distance: float = 1e-4,
-    ):
+        type_: str | None = None,
```
domfournier, line 498: Would be great if we could avoid another parameter. Any chance you can figure this out by checking the dtype of values?

domfournier (commented):
Great new checks. See comment.

geoh5py/data/text_data.py
MatthieuCMira: it ensure the array contains string values.
If not it raises an error. 

In the scenario someone want to push numerics into text, he should convert them before

geoh5py/objects/drillhole.py
MatthieuCMira: Good idea. So simple I don't even thought about it ;-)

domfournier (approved):
Great.

PR #464, opened 2024-01-04
GEOPY-1198

benk-mira:
**GEOPY-1198 - Misspelled 'depth' key in concatenated drillhole add_data not handled correctly**

### DIFF
--- a/geoh5py/objects/surveys/electromagnetics/tipper.py
+++ b/geoh5py/objects/surveys/electromagnetics/tipper.py
@@ -25,7 +25,8 @@ import numpy as np
 from geoh5py.objects.curve import Curve
 from geoh5py.objects.object_type import ObjectType
 from geoh5py.objects.points import Points
-from geoh5py.objects.surveys.electromagnetics.base import AirborneEMSurvey, FEMSurvey
+
+from .base import AirborneEMSurvey, FEMSurvey
 
 # pylint: disable=too-many-ancestors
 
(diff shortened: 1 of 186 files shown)


geoh5py/objects/surveys/electromagnetics/tipper.py
```diff
@@ -25,8 +25,7 @@
 from geoh5py.objects.curve import Curve
 from geoh5py.objects.object_type import ObjectType
 from geoh5py.objects.points import Points
-
-from .base import AirborneEMSurvey, FEMSurvey
+from geoh5py.objects.surveys.electromagnetics.base import AirborneEMSurvey, FEMSurvey
```
domfournier, line 28: You can probably revert this now.

tests/drillhole_v4_0_test.py
```diff
@@ -16,6 +16,7 @@
 #  along with geoh5py.  If not, see <https://www.gnu.org/licenses/>.
 
 # pylint: disable=R0914
+# pylint: disable=R0401
```
domfournier, line 19: Can you put both disables on the same line, and use the word style instead of the code?

domfournier (commented):
Small edits

tests/drillhole_v4_0_test.py
benk-mira: yep, removed R0401 since the circular import problem is fixed..

geoh5py/objects/surveys/electromagnetics/tipper.py
benk-mira: done

domfournier (approved):
Great. Moving on.

PR #469, opened 2024-01-10
Geopy 1287

MatthieuCMira:
**GEOPY-1229 - correct drillhole concatenated creation**
Change text bytes import into geoh5py


domfournier (commented):
Looks good, but would be nice to have a unittest that shows this at work.

MatthieuCMira:
It's covered by unitest no?

just the multi-text data is not covered.

MatthieuCMira:
I added a asssert for values property and setter

domfournier (approved):
Fantastic.

PR #470, opened 2024-01-10
GEOPY-1147

benk-mira:
**GEOPY-1147 - Increment property group for data with new depth sampling**


geoh5py/groups/property_group.py
```diff
@@ -61,7 +61,7 @@ def __init__(
 
         self._parent: ObjectBase = parent
         self._properties: list[uuid.UUID] | None = None
-        self._property_group_type = "Multi-element"
+        self._property_group_type = kwargs.get("property_group_type", "Multi-element")
```
domfournier, line 64: It would be better to add it to the list of optional arguments, so we can document it.

geoh5py/objects/drillhole.py
```diff
@@ -401,11 +402,12 @@ def add_data(
             attributes["name"] = name
 
             if attributes["name"] in self.get_data_list():
-                raise UserWarning(
+                warnings.warn(
```
domfournier, line 405: Why the change to warning?

domfournier (approved):
Great.

PR #487, opened 2024-01-22
GEOPY-1230: reorder Concatenation object

MatthieuCMira:
**GEOPY-1230 - reorder Concatenation object**
Convert the "Concatenation" Python file to a module. 

This file was unworkable: 
Tons of intricate classes calling each other. 

The idea is to do a basement for future work and refactoring. 
But the code is so big that we have to proceed step by step. So the first step is to split it up. 

I dispatched those classes in a module.
However, the quality of the previous code is still the same. 

Some straps to make it work: 
 - some type: ignore
 - an "utils" file in the module to suppress the "isinstance" creating circular imports (this file should be removed with the code improvement)


geoh5py/shared/concatenation/property_group.py
```diff
@@ -0,0 +1,102 @@
...
+    @parent.setter
+    def parent(self, parent):
+        if self._parent is not None:
+            raise AttributeError("Cannot change parent of a property group.")
+
+        if not is_concatenated_object(parent):
+            raise AttributeError(
+                "The 'parent' of a concatenated Data must be of type 'Concatenated'."
+            )
+        parent.add_children([self])
```
domfournier, line 100: I think we could avoid circular import and reduce coupling if we let the `parent` do the check for the `isinstance(child, Concatened)` on the `add_children` call, and only do `TYPE_CHECKING` imports for the type hints.

The hierarchy of real imports would be

Concatenated, ConcatenatedPropertyGroup, ConcatenatedData -> ConcatenatedObject -> Concanator

domfournier: Would require to add a `add_children` method on the `ConcatenatedObject` do to the check.

domfournier: See PR #489

MatthieuCMira: The problem is not from the parent add_children function but from the object: 
it ensure the parent the user want to set is a concatenated_object.

PR #489, opened 2024-01-23
GEOPY-1230: Change class dependencies

domfournier:
**GEOPY-1230 - reorder Concatenation object**


geoh5py/shared/concatenation/object.py
```diff
@@ -38,14 +36,39 @@ class ConcatenatedObject(Concatenated, ObjectBase, ABC):
...
+        children,
+    ):
+        """
+        :param children: Add a list of entities as
+            :obj:`~geoh5py.shared.entity.Entity.children`
+        """
+        property_groups = self._property_groups or []
+
+        for child in children:
+            if child not in self._children and isinstance(
```
MatthieuCMira, line 58: a property group is both in children and property_groups?

geoh5py/shared/concatenation/object.py
```diff
@@ -38,14 +36,39 @@ class ConcatenatedObject(Concatenated, ObjectBase, ABC):
...
+            ):
+                self._children.append(child)
+
+            if (
+                isinstance(child, ConcatenatedPropertyGroup)
+                and child not in property_groups
+            ):
+                property_groups.append(child)
+
+            if property_groups:
```
MatthieuCMira, line 69: But then if the type is not the one we want, it just pass, not raising an error or anything?

geoh5py/ui_json/validation.py
```diff
@@ -133,8 +135,11 @@ def _validations_from_uijson(ui_json: dict[str, Any]) -> dict[str, dict]:
                 validations[key] = {
                     "types": [str, UUID, int, float, Entity],
                 }
-                validations[key]["association"] = item["parent"]
-                validations[key]["uuid"] = None
+                try:
+                    validations[key]["association"] = item["parent"]
+                    validations[key]["uuid"] = None
+                except KeyError:
```
MatthieuCMira, line 141: The try except is a bit patchy. 
Why do you include it? Should we raise a specifi warning?

MatthieuCMira (commented):
Well done!
See comments/questions

geoh5py/shared/concatenation/object.py
domfournier: Correct, we did that change together last year, so we can access prop_groups from the top

geoh5py/shared/concatenation/object.py
domfournier: Currently, it won't get appended to the list. We never raised the error before in the parent class `ObjectBase`, but we could.

geoh5py/ui_json/validation.py
domfournier: That's unrelated to this issue, but pylint just caught it. It's possible that the validation just doesn't exist, so we skip over. To be removed soon anyway, just need the linters to pass.

geoh5py/shared/concatenation/object.py
MatthieuCMira: The add_data of entity accepts everything (which is strange for a public method but out of the scope of this project). 

the one of object passes, I agree, but a warning could be nice. :-)

geoh5py/shared/concatenation/object.py
MatthieuCMira: You're right

geoh5py/ui_json/validation.py
MatthieuCMira: For pylint whims, should we do this kind of code or just shut them off at the corresponding line?

domfournier: Agree. Feel free to revert it on your PR, and just add the disable instead. I am cool with that.

geoh5py/shared/concatenation/object.py
domfournier: I agree, Entity probably shouldn't have the add_children, since Data is also a sub-class but can't have children (as far as I know). Feel free to raise a specific issue for this.

geoh5py/ui_json/validation.py
MatthieuCMira: So I accept and do the changes on my PR?

domfournier:
I ll let you merge if you accept the changes,.

geoh5py/ui_json/validation.py
domfournier: Ideally yes. So we know who did the changes.

PR #490, opened 2024-01-24
GEOPY-1297: Get Tabled Data-Depth from DrillholeGroup

MatthieuCMira:
**GEOPY-1297 - Get Tabled Data-Depth from DrillholeGroup**
Add a function in Concatenator to get a table of the data (vectorial form)

With drillholes uuid, depth and val;ues associated. 

Can be usefull for in the future for:
- add_data for concatenator group
- drillhole data composite

Should we add such function on Data too?


geoh5py/shared/concatenation/concatenator.py
```diff
@@ -616,3 +617,186 @@ def update_array_attribute(
...
+
+        :param property_group: The property group to extract.
+
+        :return: the association table of the property group.
+        """
+        # get the associations
+        associations: np.ndarray = np.array([])
+        # todo: if association is DEPTH,. it just can be "Interval Table" or "Depth Table"
+        if property_group.property_group_type == "Interval table":
+            associations = np.array(
```
domfournier, line 680: No sure we need to make this an array. Putting a lot of different object types together and it makes it hard to understand what is passed around.

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -616,3 +617,186 @@ def update_array_attribute(
...
+                    [
+                        property_group.from_.name,
+                        np.sort(
+                            self.index[property_group.from_.name], order="Start index"
+                        ),
+                        self.data[property_group.from_.name],
+                    ],
+                    [
+                        property_group.to_.name,
+                        np.sort(
```
domfournier, line 691: Looks like you are assuming that from and to will be sorted the same? They might not always line up with the same drillholes

geoh5py/shared/concatenation/concatenator.py
domfournier: Also, this array call fails locally on my machine. Not sure why it works on gitaction, but I get a shape mismatch.

domfournier (commented):
We need to rethink this a bit. There are assumptions in here about sorting order that may not always work. I think we can't probably simplify this by just making a mapper between indices and drillholes to query the arrays. Let's talk

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -616,3 +617,186 @@ def update_array_attribute(
...
+                            self.index[property_group.depth_.name], order="Start index"
+                        ),
+                        self.data[property_group.depth_.name],
+                    ]
+                ]
+            )
+
+        return associations
+
+    def _pad_data_table(
```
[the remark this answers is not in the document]
MatthieuCMira: Or the values are padded, or the depths are modified.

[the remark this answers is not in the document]
MatthieuCMira: I create two simple staticmethod for each

geoh5py/shared/concatenation/concatenator.py
MatthieuCMira: I create another function o create this array.
I actually start those issues to create it...

geoh5py/shared/concatenation/concatenator.py
MatthieuCMira: Logic cahnged

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -616,3 +617,186 @@ def update_array_attribute(
...
+
+        # get the index of the data and sort it
+        data_index = np.sort(self.index[data_name], order="Start index")
+
+        # get the first object of the know association
+        object_: ConcatenatedObject = self.workspace.get_entity(  # type: ignore
+            UUID(data_index[0][-2].decode("utf-8").strip("{}"))
+        )[0]
+
+        data: Data = object_.get_data(
```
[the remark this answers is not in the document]
MatthieuCMira: The problem with that is 
"ConcatenatedData" has no "association".

and it's not False, as the fucntion get a ConcatenatedTextData, which is a concatenation of Data and concatenatedData

geoh5py/shared/concatenation/object.py
```diff
@@ -97,7 +98,7 @@ def create_property_group(
 
         return prop_group
 
-    def get_data(self, name: str | uuid.UUID):
+    def get_data(self, name: str | uuid.UUID) -> list[Data]:
```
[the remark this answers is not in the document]
domfournier: Yea that's fine. A bit annoying that the linter doesn't get the inheritance. Moving on.

domfournier (commented):
Getting there. Let's have another quick chat

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -616,3 +618,237 @@ def update_array_attribute(
...
+        data_info = {}
+        for name in data_name:
+            # ensure data_name is in the data
+            if name not in self.data:
+                raise KeyError(f"Data '{name}' not found in concatenated data.")
+
+            # get the first data of the know association
+            data: Data = self.workspace.get_entity(  # type: ignore
+                UUID(self.index[name][0][-2].decode("utf-8").strip("{}"))
+            )[0].get_data(UUID(self.index[name][0][-1].decode("utf-8").strip("{}")))[0]
```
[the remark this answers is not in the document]
MatthieuCMira: Yes, I tried without and it doesn't worked

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -616,3 +618,237 @@ def update_array_attribute(
...
+        self, data_name: str | list[str]
+    ) -> tuple[dict, ConcatenatedPropertyGroup]:
+        """
+        Extract the data no data values and the common property group for a given data name.
+
+        :param data_name: The name of the data to extract.
+
+        :return: The data with the no data value and the property group
+        """
+        if isinstance(data_name, str):
```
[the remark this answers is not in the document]
MatthieuCMira: data_info dictionary is nice because it supports both depth and from-to in the same mechanic. 

We chat about selecting several data. I think it's a good idea, and we can select unique data as str.

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -616,3 +618,237 @@ def update_array_attribute(
...
+        Constructs a dictionary with the objects as keys.
+        The values are (start index, n count and no data value) tuple.
+        The order of the dictionary depends on the first key of 'data_info'.
+
+        :param data_info: the dictionary containing the name of the data and the no data values.
+
+        :return: The object dictionary {object: {Column: (start_index, n_count, ndv)}}.
+        """
+        object_index_dictionary: dict = {}
+        for object_ in self.index[list(data_info.keys())[0]]["Object ID"]:
```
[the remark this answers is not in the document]
MatthieuCMira: data_info[0] is the 'From' or 'Depth' index. 

Select one or several name should be a conversation to have before going further.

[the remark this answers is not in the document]
MatthieuCMira: I used the same logic but with a simple function with obvious name.

geoh5py/shared/concatenation/concatenator.py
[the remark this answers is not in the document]
MatthieuCMira: I create a sup-function to pass a str. Logics are separated

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -616,3 +618,237 @@ def update_array_attribute(
...
+
+        :param: data_info: The data with the no data value.
+        :param property_group: The property group to extract.
+
+        return: the same dict with the depth name and no data value.
+        """
+        if property_group.property_group_type == "Interval table":
+            # add form and to at the beginning of
+            data_info = {
+                **{property_group.from_.name: np.nan, property_group.to_.name: np.nan},
```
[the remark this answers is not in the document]
MatthieuCMira: accessing nan_value from data (changes in data)

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -616,3 +618,237 @@ def update_array_attribute(
...
+        """
+        Constructs a dictionary with the objects as keys.
+        The dictionary contains the data name as keys
+            and the values are (start index, n count and no data value) tuple.
+
+        :param data_name: The name of the data to extract.
+
+        :return: The object dictionary {object: {Column: (start_index, n_count, ndv)}}.
+        """
+
```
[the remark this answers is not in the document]
MatthieuCMira: Change with this logic. 
I just keep the list append for rec.array, as discussed yesterday

geoh5py/shared/concatenation/drillholes_concatenator.py
```diff
@@ -0,0 +1,271 @@
...
+                    self.index[name][self.index[name]["Object ID"] == drillhole][0]
+                )[:2]
+                for name in names
+            }
+            for drillhole in self.index[names[0]]["Object ID"]
+        }
+
+        return object_index
+
+    def depth_name_association(self, name: str) -> dict:
```
domfournier, line 103: This is getting a bit confusing with `depth_names_association`. Feels like they could be combined, or use this method to just validate the depth association?

geoh5py/shared/concatenation/drillholes_concatenator.py
```diff
@@ -0,0 +1,271 @@
...
+        if property_group.property_group_type == "Interval table":
+            return property_group.from_.name, property_group.to_.name
+        if property_group.property_group_type == "Depth table":
+            return (property_group.depth_.name,)
+
+        return None
+
+    def association_by_drillhole(
+        self,
+        names: tuple,
```
domfournier, line 77: Let's be explicit about the inputs, and only take it one data name. If I understand correctly it is only used one at a time right now anyway.

```suggestion
        depth_name: str,
        data_name: str
```

geoh5py/shared/concatenation/drillholes_concatenator.py
```diff
@@ -0,0 +1,271 @@
...
+        # ensure all the names are in the index
+        if not all(name in self.index for name in names):
+            raise KeyError(f"Data '{names}' not found in concatenated data.")
+
+        object_index = {
+            drillhole: {
+                name: list(
+                    self.index[name][self.index[name]["Object ID"] == drillhole][0]
+                )[:2]
+                for name in names
```
domfournier, line 96: We can avoid the nested loop if we assume single element.

domfournier (commented):
To be discussed on new PR to here

geoh5py/shared/concatenation/drillholes_concatenator.py
MatthieuCMira: Sure but it's nice to have this function as a standalone

geoh5py/shared/concatenation/drillholes_concatenator.py
MatthieuCMira: Well, depends on the function. this is why I add a function for list and tuple, to make it clearer

geoh5py/shared/concatenation/drillholes_concatenator.py
MatthieuCMira: Well, the nested loop is taking account To if from-to to avoid different process for depth and from to

For from to, only from is considerated
For depth, only depth

geoh5py/shared/concatenation/drillholes_concatenator.py
```diff
@@ -0,0 +1,271 @@
...
+        :return: A dictionary with the object uuid and the index of all the data.
+        """
+        # ensure all the names are in the index
+        if not all(name in self.index for name in names):
+            raise KeyError(f"Data '{names}' not found in concatenated data.")
+
+        object_index = {
+            drillhole: {
+                name: list(
+                    self.index[name][self.index[name]["Object ID"] == drillhole][0]
```
domfournier, line 94: Get only the first index? I am not sure if this is going to work if you have more than one drillhole. Can you augment your test to check that?

domfournier: Also, it assumes that all data tables have all the drillholes in them, which is not necessarily the case. Some drillholes may not have certain data on them but share the same property group name.

geoh5py/shared/concatenation/drillholes_concatenator.py
```diff
@@ -0,0 +1,271 @@
...
+                )[:2]
```
domfournier, line 95: This assumes that you will always find the ID drillhole on all data tables, which is not necessarily the case. Some drillholes may not have certain data on them but share the same property group name.

tests/drillhole_v4_0_test.py
```diff
@@ -732,3 +732,201 @@ def test_is_collocated(tmp_path):
...
+        value = np.random.randn(8)
+
+        well.add_data(
+            {
+                "Depth Data": {
+                    "depth": depth,
+                    "values": value,
+                },
+            }
+        )
```
domfournier, line 805: In order to cover the case where some data are not present on certain drillhole, can you add another drillhole and another data to your test? Only one of the data should be shared by both drillholes.

domfournier (commented):
So I couldn't come up with anything better. Let's go with you implementation, but make sure that we cover the case mentioned below. This can happen very often that not all drillholes contain all data.

geoh5py/shared/concatenation/drillholes_concatenator.py
MatthieuCMira: Put a if in the loop

domfournier (commented):
Thanks for the changes. Still need to work on the test a bit though

1- Make sure it writes out something valid
2- Test the case of partial overlap between the data tables. We want to test the concatenation of data with 0 length on certain drillholes.

PR #491, opened 2024-01-24
GEOPY-1296: geoh5py: export data in vectorial form in Drillholes PropertyGroup

MatthieuCMira:
**GEOPY-1296 - geoh5py: export data in vectorial form in Drillholes PropertyGroup**
add a function "add_data_from_template" in concatenator

wait for 1297 to be review first


geoh5py/shared/concatenation/drillholes_group_table.py
```diff
@@ -0,0 +1,101 @@
...
+                f"The name '{name}' is not in the parent's unique_property_group_names."
+            )
+
+        self._name: str = name
+
+    # get all the property groups in the concatenator
+    def association_by_drillhole(
+        self,
+        names: tuple,
+    ) -> dict:
```
domfournier, line 54:
```suggestion
    ) -> dict[uuid.UUID, list]:
```

geoh5py/shared/concatenation/drillholes_group_table.py
```diff
@@ -0,0 +1,101 @@
...
+            object_index[drillhole] = {}
+            for name in names:
+                if drillhole in self._parent.index[name]["Object ID"]:
+                    object_index[drillhole][name] = list(
+                        self._parent.index[name][
+                            self._parent.index[name]["Object ID"] == drillhole
+                        ][0]
+                    )[:2]
+                else:
+                    object_index[drillhole][name] = [0, 0]
```
domfournier, line 81: Make this chunk a method of Concatenator, something like ?

`def get_data_interval(self, drillhole: uuid.UUID, name: str) -> tuple:`

geoh5py/shared/concatenation/drillholes_concatenator.py
```diff
@@ -251,30 +161,123 @@ def _truncate_arrays_to_smallest(arrays: list[np.ndarray]) -> list[np.ndarray]:
...
+                            indices[template][0] : indices[template][0]
+                            + indices[template][1]
+                        ],
+                        association: np.array(drillhole_association).T,
+                    },
+                },
+                property_group=property_group,
+            )
+
+    def add_values_on_property_group(
```
domfournier, line 239: Suggest we change this to `add_values_on_data_table` and pass a DrillholeGroupTable as argument.

geoh5py/shared/concatenation/drillholes_concatenator.py
```diff
@@ -251,30 +161,123 @@ def _truncate_arrays_to_smallest(arrays: list[np.ndarray]) -> list[np.ndarray]:
...
+        drillholes_group_table = DrillholesGroupTable(self, property_group)
+
+        # if template is None, get the first association for template
+        if template is None:
+            template_: tuple = drillholes_group_table.association_names
+            template_ = (template_[0],) + template_
+        elif template not in self.unique_property_group_names[property_group][1]:
+            raise KeyError(f"The template '{template}' is not in the property group.")
+        else:
+            template_ = (template,) + drillholes_group_table.association_names
```
domfournier, line 267: This is a bit too complicated. I think the Table should know about its own "depth association" and shouldn't have to be supplied in a specific order.

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -615,3 +618,131 @@ def update_array_attribute(
             self.data[alias] = values
 
         self.save_attribute(field)
+
+    @staticmethod
+    def get_depth_association(
+        property_group: ConcatenatedPropertyGroup,
```
domfournier, line 624: Suggest we move this to Table class

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -615,3 +618,131 @@ def update_array_attribute(
...
+        """
+        if property_group.properties is None:
+            return (None,)
+        return tuple(
+            property_group.parent.get_data(property_)[0].name
+            for property_ in property_group.properties
+        )
+
+    @property
+    def unique_property_group_names(self) -> dict:
```
domfournier, line 656: I think it would be nicer to ask for all Tables, and let Tables figure out what is under them.

geoh5py/shared/concatenation/drillholes_concatenator.py
```diff
@@ -251,30 +161,123 @@ def _truncate_arrays_to_smallest(arrays: list[np.ndarray]) -> list[np.ndarray]:
...
+        for idx, data_name in enumerate(names):
+            type_temp = np.array([output[0, idx]]).dtype
             if type_temp.kind in ["S", "U"]:
                 dtype.append((data_name, "O"))
             else:
                 dtype.append((data_name, type_temp))
 
         return np.core.records.fromarrays(output.T, dtype=dtype)
+
+    def _push_values_to_drillhole(
```
domfournier, line 184: Same here, we can make this a method of Tables

domfournier (commented):
I think Tables should do more of the heavy lifting, and reduce the amount of new code under Concatenator.

geoh5py/shared/concatenation/drillholes_group_table.py
```diff
@@ -0,0 +1,377 @@
...
+                        "values": values[
+                            indices[self.association[0]][0] : indices[
+                                self.association[0]
+                            ][0]
+                            + indices[self.association[0]][1]
+                        ],
+                        self.association_type: np.array(drillhole_association).T,
+                    },
+                },
+                property_group=self.name,
```
domfournier, line 199: I think this would create a new property group?
Simpler is just append to the current one like this

```suggestion
                property_group=self._property_groups[drillhole.uid],
```

geoh5py/shared/concatenation/drillholes_group_table.py
```diff
@@ -0,0 +1,377 @@
...
+            drillhole.add_data(
+                {
+                    name: {
+                        "values": values[
+                            indices[self.association[0]][0] : indices[
+                                self.association[0]
+                            ][0]
+                            + indices[self.association[0]][1]
+                        ],
+                        self.association_type: np.array(drillhole_association).T,
```
domfournier, line 196: Don't need to define depth if you know which property_group it belongs to

```suggestion
```

geoh5py/shared/concatenation/drillholes_group_table.py
```diff
@@ -0,0 +1,377 @@
...
+        return self._association_type
+
+    @property
+    def depth_table(
+        self,
+    ) -> np.ndarray:
+        """
+        Get a table with all the data associated with depth for every drillhole object.
+        The Drillhole name is added at the beginning of the table for every row.
+        The table is based on the association and contains nan values if no data is found.
```
domfournier, line 223:
```suggestion
        Get a table with all the data associated with depth for every drillhole object.
        
        The drillhole name is added at the beginning of the table for every row.
        The table is based on the association and contains no-data-values if no data is found.
```

geoh5py/shared/concatenation/drillholes_group_table.py
```diff
@@ -0,0 +1,377 @@
...
+
+    :param parent: The parent DrillholesConcatenator.
+    :param name: The name of the PropertyGroup to extract.
+    """
+
+    def __init__(self, parent: Concatenator, name: str):
+        self._association: tuple = ()
+        self._association_type: str = ""
+        self._depth_table: np.ndarray = np.array([])
+        self._index_by_drillhole: dict = {}
```
domfournier, line 45: Better to instantiate to None instead of mutable

domfournier (commented):
Fantastic, getting very clean. See one more comment about handling of add_data with property_group instead of association depth. Also, we will probably need some mechanism to update/reset the table if data is added on the drillhole?

geoh5py/shared/concatenation/drillholes_group_table.py
MatthieuCMira: ok but it makes a lot of boring pylint and type gestion. 
With an empty iterable, when it has to be iterate, the loop pass and everyone's happy.

geoh5py/shared/concatenation/drillholes_group_table.py
MatthieuCMira: No, as data in drillhole Concatenator, property Groups are based on names. 
GA recognize it

domfournier (approved):
Looks great!

PR #501, opened 2024-02-06
GEOPY-1319: add_children consistency

MatthieuCMira:
**GEOPY-1319 - add_children consistency**
Split entity into 2 classes:

Entity and EntityConcatenator

The first one is used for all object
The second deal with children that are needed for group and object

### DIFF
--- a/geoh5py/data/data.py
+++ b/geoh5py/data/data.py
@@ -218,6 +218,12 @@ class Data(Entity):
     def primitive_type(cls) -> PrimitiveTypeEnum:
         ...
 
+    def add_children(self, children: list):
+        """
+        Alias not implemented from base Entity class.
+        """
+        raise NotImplementedError("Data entity cannot contain children.")
+
     def add_file(self, file: str):
         """
         Alias not implemented from base Entity class.
--- a/geoh5py/data/data.py
+++ b/geoh5py/data/data.py
@@ -150,16 +150,6 @@ class Data(Entity):
         )
 
     @property
-    def extent(self) -> np.ndarray | None:
-        """
-        Geography bounding box of the parent object.
-
-        :return: shape(2, 3) Bounding box defined by the bottom South-West and
-            top North-East coordinates.
-        """
-        return None
-
-    @property
     def n_values(self) -> int | None:
         """
         :obj:`int`: Number of expected data values based on
@@ -246,7 +236,9 @@ class Data(Entity):
     @classmethod
     @abstractmethod
     def primitive_type(cls) -> PrimitiveTypeEnum:
-        ...
+        """
+        Return the primitive type of the data.
+        """
 
     def mask_by_extent(
         self, extent: np.ndarray, inverse: bool = False
--- a/geoh5py/data/data.py
+++ b/geoh5py/data/data.py
@@ -67,7 +67,6 @@ class Data(Entity):
     def copy(
         self,
         parent=None,
-        copy_children: bool = True,
         clear_cache: bool = False,
         mask: np.ndarray | None = None,
         **kwargs,
@@ -77,7 +76,6 @@ class Data(Entity):
 
         :param parent: Target parent to copy the entity under. Copied to current
             :obj:`~geoh5py.shared.entity.Entity.parent` if None.
-        :param copy_children: (Optional) Create copies of all children entities along with it.
         :param clear_cache: Clear array attributes after copy.
         :param mask: Array of bool defining the values to keep.
         :param kwargs: Additional keyword arguments to pass to the copy constructor.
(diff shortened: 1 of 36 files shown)


geoh5py/data/data.py
```diff
@@ -119,6 +117,38 @@ def copy(
 
         return new_entity
 
+    def copy_from_extent(
```
domfournier, line 120: Can we add a unitest for this? You can probably just modify the copy from extent test on the object, and check that the call on the data yields the same outcome.

domfournier (commented):
Looks great. Just need to augment tests a bit.

geoh5py/data/data.py
```diff
@@ -62,12 +62,11 @@ def __init__(
         if self.entity_type.name == "Entity":
             self.entity_type.name = self.name
 
-        data_type.workspace._register_data(self)
+        getattr(data_type.workspace, "_register_data")(self)
```
domfournier, line 65: We've been warned by @sebhmg to avoid sweeping the private methods under the rug with `getattr`.

@sebhmg Would there be an issue with making the registering public? Else, we could add a `register_entity` method that calls the right private method depending on class type...

domfournier (commented):
Minor comment. Let's here from Seb's comment.

geoh5py/data/data.py
domfournier: Ok @MatthieuCMira , let's create a public `register` method on the workspace. Then workspace does the `weakref_utils.insert_once` based on the type of the incoming argument. We can simplify the inits of all other class types later, but do the change here now.

sebhmg: right, since each object registers upon its instantiation, it probably make sense to expose it as a public method.
Discussing with @domfournier, he suggest a single public `register()` method that would call the other private ones base on the object type. It seems to make sense

domfournier: Should look here like

```suggestion
        data_type.workspace.register(self)
```

sebhmg: does it need to be in each Entity class then?
Could it be in Entity __init__ ?

domfournier: Exactly what I had in mind. We can move the registering all the way down the base Entity class + property_groups (which are special kind). Same for entity_type

MatthieuCMira: I created a register function and integrated it in entity,
Removed it everywhere else

MatthieuCMira: Resolving this conversation

geoh5py/workspace/workspace.py
```diff
@@ -1228,6 +1224,25 @@ def open(self, mode: str | None = None) -> Workspace:
...
+        if isinstance(entity, EntityType):
+            self._register_type(entity)
+        elif isinstance(entity, Group):
+            self._register_group(entity)
+        elif isinstance(entity, Data):
+            self._register_data(entity)
+        elif isinstance(entity, ObjectBase):
+            self._register_object(entity)
+        elif isinstance(entity, PropertyGroup):
+            self.register_property_group(entity)
```
domfournier, line 1242: Should have been accessing the private method, but let's just do the operation here.

domfournier (commented):
Excellent. Some minor changes below.

geoh5py/workspace/workspace.py
sebhmg, line 1242: **thought**: I see `PropertyGroup` does not inherit from `Entity` and calls directly `self.parent.workspace.register_property_group(self)`

Thus it will not go through this `register()` function. For consistency, `PropertyGroup __init__()` could call `workspace.register()`. and possibly the code of `register_property_group()` is moved to `_register_property_group()`

I am also curious: why this special case with PropertyGroup?
```
        if not property_group.on_file:
            self.add_or_update_property_group(property_group)
```

Would it make sense that `PropertyGroup` inherits from `Entity`?

sebhmg (commented):
looks good. thanks !

I am curious though about the special case for `PropertyGroup`. See comment.

geoh5py/workspace/workspace.py
MatthieuCMira: A bunch of methods are not shared between entity and PropertyGroup.
Still, it might be possible to create some superclass over entity. 

Dom, what do you think?

domfournier: Property groups are different since they are only registered under the parent object, and not under specific h5 groups. I think it was just a typo here with accessing a non-private method? This issue is already getting way out of scope, I would avoid touching property_group at this stage.

geoh5py/workspace/workspace.py
```diff
@@ -1228,22 +1208,29 @@ def open(self, mode: str | None = None) -> Workspace:
...
+        """
+        Register an entity to the workspace based on its type.
 
-    def _register_object(self, obj: ObjectBase):
-        weakref_utils.insert_once(self._objects, obj.uid, obj)
+        :param entity: The entity to be registered.
+        """
+        dict_entity: dict[uuid.UUID, ReferenceType]
+        if isinstance(entity, EntityType):
+            dict_entity = self._types
```
domfournier, line 1219: Wowzer, nope. This is not doing anything right now. We still need the `weak_ref.insert_once` for every case.

geoh5py/workspace/workspace.py
```diff
@@ -1228,22 +1208,29 @@ def open(self, mode: str | None = None) -> Workspace:
...
+        elif isinstance(entity, Group):
+            dict_entity = self._groups
+        elif isinstance(entity, Data):
+            dict_entity = self._data
+        elif isinstance(entity, ObjectBase):
+            dict_entity = self._objects
+        elif isinstance(entity, PropertyGroup):
+            dict_entity = self._property_groups
+            if not entity.on_file:
+                self.add_or_update_property_group(entity)
```
domfournier, line 1229: Same here, we still want to register the weakref of the property_group.

domfournier (commented):
Not quite

domfournier (approved):
Never mind! I just noticed the weakref at the bottom ... good.

geoh5py/workspace/workspace.py
sebhmg: I agree changing intheritence of PropertyGroup is out of scope. I just want ed to raise the question.
Thanks @domfournier for the clarification.

MatthieuCMira:
Ready to merge?

geoh5py/workspace/workspace.py
```diff
@@ -1230,35 +1214,23 @@ def register(self, entity: Entity | EntityType | PropertyGroup):
...
-            self._register_data(entity)
+            dict_entity = self._data
         elif isinstance(entity, ObjectBase):
-            self._register_object(entity)
+            dict_entity = self._objects
         elif isinstance(entity, PropertyGroup):
-            self.register_property_group(entity)
+            dict_entity = self._property_groups
+            if not entity.on_file:
+                self.add_or_update_property_group(entity)
```
sebhmg, line 1229: **question:** no consequence of calling `self.add_or_update_property_group(entity)` before registering it?
It used to register first.

PR #503, opened 2024-02-08
Geopy 1341b

MatthieuCMira:
add a conveniency function to access property_group from data_name

### DIFF
--- a/geoh5py/shared/concatenation/drillholes_group_table.py
+++ b/geoh5py/shared/concatenation/drillholes_group_table.py
@@ -265,7 +265,7 @@ class DrillholesGroupTable(ABC):
 
     def depth_table_by_name(
         self,
-        names: tuple[str],
+        names: tuple[str] | str,
         spatial_index: bool = False,
     ) -> np.ndarray:
         """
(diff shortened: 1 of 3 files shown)


MatthieuCMira:
I had a small line concerning a small bug

geoh5py/shared/concatenation/drillholes_group_table.py
```diff
@@ -151,19 +153,22 @@ def _get_property_groups(parent, name) -> dict[UUID, ConcatenatedPropertyGroup]:
         return property_groups
 
     @staticmethod
-    def _pad_arrays_to_first(arrays: list[np.ndarray], ndv: Any) -> list[np.ndarray]:
+    def _pad_arrays_to_first(
```
domfournier, line 156: Feels like this shouldn't be static, and be called something more explicit about what it does.
Maybe more like `def pad_array_to_association(self)`? The table should know about its association lenght.

domfournier (commented):
See comment for clarity.

geoh5py/shared/concatenation/drillholes_group_table.py
MatthieuCMira: I agree for the name. 

For the static vs non static, it doesn't import much. 
I'll call index by drillhole if you consider it important.

domfournier: Well if it's not used externally, and relies on the state of local attributes (i.e. `association`) then it makes more sense to be a regular method.

domfournier (approved):
Thanks, looks good.

PR #504, opened 2024-02-13
GEOPY-1351: Implement referenced data in depth_tables add_data

MatthieuCMira:
**GEOPY-1351 - Implement referenced data in depth_tables add_data**
add referenced values and test


geoh5py/shared/concatenation/drillholes_group_table.py
```diff
@@ -206,6 +206,17 @@ def add_values_to_property_group(
...
+            add_data_dict = {
+                "values": values[
+                    indices[self.association[0]][0] : indices[self.association[0]][0]
+                    + indices[self.association[0]][1]
+                ],
+            }
+
+            if isinstance(value_map, dict):
+                add_data_dict["type"] = "referenced"
+                add_data_dict["value_map"] = value_map
```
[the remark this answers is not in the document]
domfournier: where `same_type` is created before the loop using

```
new_type = DataType(workspace, primitive_type = "REFERENCED", value_map = value_map)
```

We should also do it for all data types while we are at it.

geoh5py/workspace/workspace.py
```diff
@@ -1214,24 +1214,21 @@ def register(self, entity: Entity | EntityType | PropertyGroup):
...
+            weakref_utils.insert_once(self._groups, entity.uid, entity)
         elif isinstance(entity, Data):
-            dict_entity = self._data
+            weakref_utils.insert_once(self._data, entity.uid, entity)
         elif isinstance(entity, ObjectBase):
-            dict_entity = self._objects
+            weakref_utils.insert_once(self._objects, entity.uid, entity)
         elif isinstance(entity, PropertyGroup):
-            dict_entity = self._property_groups
+            weakref_utils.insert_once(self._property_groups, entity.uid, entity)
```
domfournier, line 1226: Ok that' good. Unrelated but will fix the comment raised after GEOPY-1319

tests/drillhole_v4_0_test.py
```diff
@@ -911,9 +912,17 @@ def test_add_data_to_property(tmp_path):
             "interval_values_a", spatial_index=True
         )
 
+        verification_map_value = (verification["interval_values_a"] * 10).astype(
+            np.int32
+        )
+        verification_map_value -= np.min(verification_map_value)
+
+        value_map = {idx: f"{idx}" for idx in np.unique(verification_map_value)}
```
domfournier, line 920: This map would actually not be valid for GA since it always needs the `0: "Unknown"` entry at the beginning. Maybe it's a good time to add a validation on the `value_map` setter.

domfournier (commented):
That will be helpful, thanks for this. See comments.

geoh5py/shared/concatenation/drillholes_group_table.py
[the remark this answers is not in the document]
MatthieuCMira: Arn't other types automatically detected?

geoh5py/workspace/workspace.py
MatthieuCMira: yes, and not changing anything. Marking it as resolve

tests/drillhole_v4_0_test.py
MatthieuCMira: Shouldn't we check keys and add an "unknown"?
Or just check if present?

MatthieuCMira: Improved setter method for value map.

geoh5py/data/reference_value_map.py
```diff
@@ -19,19 +19,30 @@
...
     def __setitem__(self, key, value):
+        if not isinstance(key, (int, np.int16, np.int32, np.int64)) or key < 0:
+            raise KeyError("Key must be an positive integer")
+        if key == 0:
+            if value != "Unknown" and not (
+                len(self) <= 2 and value == "False" and value == "True"
+            ):
+                raise ValueError("Value for key 0 must be 'Unknown'")
+        if not isinstance(value, str):
+            raise TypeError("Value must be a string")
```
domfournier, line 43: Can we turn this into a small static method to reduce code duplication with the map.setter?

geoh5py/data/reference_value_map.py
```diff
@@ -40,13 +51,31 @@ def __call__(self):
...
+            for k in value.keys()
+        ):
+            raise KeyError("Map keys must be positive integers")
+
+        if 0 in value.keys():
+            if value[0] != "Unknown" and not (
+                len(value) <= 2 and value[0] == "False" and value[1] == "True"
+            ):
+                raise ValueError("Map value for 0 must be 'Unknown'")
+        else:
```
domfournier, line 78: Could make an explicit loop and call the static method to check each key-value pairs. Only need to check/append the 0:"Unknown" entry beforehand.

geoh5py/shared/concatenation/drillholes_group_table.py
```diff
@@ -205,6 +206,18 @@ def add_values_to_property_group(
...
+            new_type = DataType(
+                self.parent.workspace, primitive_type="REFERENCED", value_map=value_map
+            )
+        else:
+            new_type = DataType.find_or_create(
+                self.parent.workspace,
+                **DataType.validate_data_type(
+                    self.parent.workspace, {"name": name, "values": values}
+                ),
+            )
```
domfournier, line 219: That's great - one step closer to have those methods consolidated. Feels like eventually the two may be combined, but would require a deeper look at where they are used.

geoh5py/data/reference_value_map.py
```diff
@@ -19,19 +19,30 @@
...
+import numpy as np
+
 
 class ReferenceValueMap(ABC):
     """Maps from reference index to reference value of ReferencedData."""
 
-    def __init__(self, color_map: dict[int, str] | None = None):
-        self._map = {} if color_map is None else color_map
+    def __init__(self, color_map: dict[int, str]):
+        self.map: dict[int, str] = color_map
```
domfournier, line 29:
```suggestion
    def __init__(self, value_map: dict[int, str]):
        self.map: dict[int, str] = value_map
```

domfournier (commented):
Looks much better. See small comments

geoh5py/data/reference_value_map.py
domfournier: That was my bad naming here - has nothing to do with colors...

MatthieuCMira:
Wait before merging. 
I have a problem in testing. It creates empty objects.

image

geoh5py/shared/concatenation/drillholes_group_table.py
MatthieuCMira: I combined the two.
Tests were passing on geoh5py, but when I tested with GA it was not able to recognize value_map.
I went back to dic attribute assignations.

domfournier (approved):
To be continued on GEOPY-1375

PR #505, opened 2024-02-14
GEOPY-1364: facilate metadata assignation to object

MatthieuCMira:
**GEOPY-1364 - facilate metadata assignation to object**
update metadata to not rewrite everything and push to existing dictionary.

Change usage and inheritance of this function in geoh5py

Test everything

PS: my-py doesn't like super() functions, I ignore it.

### DIFF
--- a/geoh5py/objects/surveys/direct_current.py
+++ b/geoh5py/objects/surveys/direct_current.py
@@ -14,6 +14,9 @@
 #
 #  You should have received a copy of the GNU Lesser General Public License
 #  along with geoh5py.  If not, see <https://www.gnu.org/licenses/>.
+
+# pylint: disable=too-many-ancestors
+
 from __future__ import annotations
 
 import logging
@@ -35,7 +38,6 @@ class BaseElectrode(Curve, ABC):
     _current_electrodes: CurrentElectrode | None = None
 
     def __init__(self, object_type: ObjectType, **kwargs):
-        self._metadata: dict | None = None
         self._ab_cell_id: ReferencedData | None = None
 
         super().__init__(object_type, **kwargs)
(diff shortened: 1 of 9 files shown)


geoh5py/shared/entity.py
```diff
@@ -237,10 +233,23 @@ def metadata(self) -> dict | None:
...
+                )
+
+            # ensure metadata is loaded
+            if getattr(self, "_metadata", None) is None:
+                self._metadata = self.workspace.fetch_metadata(self.uid)
+
+            if isinstance(self._metadata, dict):
+                self._metadata.update(value)
+            else:
+                self._metadata = value
```
domfournier, line 249: Go straight through the getter to check/load

domfournier, line 249:
```suggestion
            if isinstance(self.metadata, dict):
                self._metadata.update(value)
            else:
                self._metadata = value
```

geoh5py/shared/entity.py
```diff
@@ -237,10 +233,23 @@ def metadata(self) -> dict | None:
     @metadata.setter
     def metadata(self, value: dict | None):
         if value is not None:
-            assert isinstance(
-                value, (dict, str)
-            ), f"Input metadata must be of type {dict} or None"
-        self._metadata = value
+            if not isinstance(value, dict):
```
domfournier, line 236: You can remove one level of indentation if you treat each case individually. It will read better too

```suggestion
        if isinstance(value, dict):
            do the dict operations
            
        elif value is None:
            self._metadata = None
            
        else:
            raise TypeError(
                    "Input metadata must be of type dict or None"
                    f" find '{type(value)}'."
                )
        
        self.workspace.update_attribute(self, "metadata")
        
```

geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -337,19 +337,18 @@ def edit_metadata(self, entries: dict[str, Any]):
 
         :param entries: Metadata key value pairs.
         """
-        metadata = self.metadata.copy()
+        em_metadata = self.metadata.copy().pop("EM Dataset", {})
```
domfournier, line 340: We would almost need to change the name of this method `edit_em_metadata` to reflect what it does (i.e. only modify the 'EM Dataset' field of the metadata.)

geoh5py/objects/surveys/direct_current.py
```diff
@@ -229,36 +231,18 @@ def current_electrodes(self):
...
 
-        if self.workspace.get_entity(values["Potential Electrodes"])[0] is None:
-            raise IndexError("Input Potential Electrodes uuid not present in Workspace")
+            for key in default_keys:
+                if self.workspace.get_entity(values[key])[0] is None:
+                    raise IndexError(f"Input {key} uuid not present in Workspace")
 
-        self._metadata = values
-        self.workspace.update_attribute(self, "metadata")
+        super(Curve, Curve).metadata.fset(self, values)  # type: ignore
```
domfournier, line 245: Not sure this will work generally with more metadata then just the `default_keys`. Can you augment the tests under `tests/survey_dcip_test.py` to check?

domfournier (commented):
Mostly looks good. See comments.

geoh5py/objects/surveys/electromagnetics/base.py
MatthieuCMira: I will duplicate the function to have a deprecation warning for 0.10 for this method.

geoh5py/objects/surveys/direct_current.py
MatthieuCMira: Now, metadata must have those two parameters for direct_current. 
Tell me what you think about the implementation.
It's tested

domfournier: Yep, logic looks good.

geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -330,26 +331,40 @@ def default_units(self) -> list[str]:
...
+
+        :param entries:
+        :return:
+        """
+        warn(
+            "DEPRECATION WARNING"
+            "The method 'edit_metadata' will be deprecated in 0.10 version. "
+            "It will be replaced by 'edit_em_metadata'"
+        )
+        self.edit_em_metadata(entries)
```
domfournier, line 367: That's great.

domfournier (approved):
Super.

PR #507, opened 2024-02-20
GEOPY-1375: improve DataType in geoh5py 

MatthieuCMira:
Changing/Correcting the 'Type' class 

Main changes in EntityType, DataType, GroupType, ObjectType

The tests were not implemented for most Type options. Implementing tests for types can be done in this issue or in another issue? 

A first review of the mechanics could be done before looking at the tests.

The branch is based on 1351. Can we merge it before on develop?


geoh5py/data/data_type.py
```diff
@@ -29,13 +29,29 @@
 from .reference_value_map import ReferenceValueMap
 
 if TYPE_CHECKING:
-    from .. import workspace
-    from . import data  # noqa: F401
+    from ..workspace import Workspace
+    from .data import Data  # noqa: F401
+
+COLOR_MAPPING: list[str] = ["linear", "equal_area", "logarithmic", "cdf", "missing"]
```
domfournier, line 35: Sounds like the right place for Literal

```suggestion
from typing import get_args, Literal

ColorMapping = Literal["linear", "equal_area", "logarithmic", "cdf", "missing"]
```

geoh5py/data/data_type.py
```diff
@@ -29,13 +29,29 @@
...
 
 
 class DataType(EntityType):
+    # pylint: disable=too-many-arguments
     """
-    DataType class
+    DataType class. It controls all the attributes of the data type for
+    displays in Geoscience Analyst.
```
domfournier, line 42:
```suggestion
    DataType class. 
    
    Controls all the attributes of the data type for displays in Geoscience ANALYST.
```

geoh5py/data/data_type.py
```diff
@@ -49,30 +65,65 @@ class DataType(EntityType):
...
-
-    def __init__(self, workspace: workspace.Workspace, **kwargs):
-        assert workspace is not None
+    def __init__(
+        self,
+        workspace: Workspace,
+        primitive_type: type[Data] | PrimitiveTypeEnum | str | None = None,
+        color_map: ColorMap | None = None,
+        hidden: bool = False,
+        mapping: str = "equal_area",
```
domfournier, line 74:
```suggestion
        mapping: ColorMapping = "equal_area",
```

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,209 @@
...
+    :param description: The description of the entity type.
+    :param name: The name of the entity type.
+    :param kwargs: Additional keyword arguments to set as attributes.
+    """
 
-        assert uid is None or isinstance(uid, uuid.UUID)
-        self._description: str | None = "Entity"
-        self._name: str | None = "Entity"
-        self._on_file = False
-        self._uid: uuid.UUID = uid if uid is not None else uuid.uuid4()
```
domfournier, line 43: No, we need to define the private attributes.

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,209 @@
...
+    # todo: This mechanic feels quite hacky
```
domfournier, line 42:
```suggestion

```

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,209 @@
...
-                setattr(self, attr, item)
-            except AttributeError:
-                continue
+    def __init__(
+        self,
+        workspace: Workspace,
+        uid: uuid.UUID | None = None,
+        description: str | None = "Entity",
+        name: str | None = "Entity",
+        **kwargs,
```
domfournier, line 51:
```suggestion
```
Let's try to remove kwargs, we should technically know exactly what's coming in.

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,209 @@
...
+    ):
+
+        self._workspace: Workspace = self._set_workspace(workspace)
+        self.uid = self._modify_attribute("ID", uid, **kwargs)
+        self.description = self._modify_attribute("Description", description, **kwargs)
+        self.name = self._modify_attribute("Name", name, **kwargs)
```
domfournier, line 57: It would probably make more sense to do the translation directly on the reader, and only deal with formatted arguments here, and send it straight to the setters

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,209 @@
...
-                    attr = self._attribute_map[attr]
-                setattr(self, attr, item)
-            except AttributeError:
-                continue
+    def __init__(
+        self,
+        workspace: Workspace,
+        uid: uuid.UUID | None = None,
+        description: str | None = "Entity",
+        name: str | None = "Entity",
```
domfournier, line 50:
```suggestion
        uid: uuid.UUID = uuid.UUID4,
        description: str = "Entity",
        name: str = "Entity",
```

I don't think these should ever be None

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,209 @@
...
+        :param kwargs: The kwargs to modify the attribute.
+
+        :return: The modified attribute.
+        """
+        if cls._attribute_map.get(attribute_name, None):
+            if attribute_name in kwargs:
+                return kwargs[attribute_name]
+        return attribute
+
+    def _set_workspace(self, workspace: Workspace) -> Workspace:
```
domfournier, line 78: Why not just a setter on the workspace property?

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,209 @@
...
+        self._workspace: Workspace = self._set_workspace(workspace)
+        self.uid = self._modify_attribute("ID", uid, **kwargs)
+        self.description = self._modify_attribute("Description", description, **kwargs)
+        self.name = self._modify_attribute("Name", name, **kwargs)
+        self._on_file = False
 
         self.workspace.register(self)
 
+    @classmethod
+    def _modify_attribute(cls, attribute_name: str, attribute: Any, **kwargs):
```
domfournier, line 63: Don't think we need this if we handle the string conversion upstream.

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,209 @@
...
     @property
     def description(self) -> str | None:
+        """
+        The description of the entity type.
+        """
         return self._description
 
     @description.setter
-    def description(self, description: str):
+    def description(self, description: str | None):
```
domfournier, line 132: Don't think it can be None.

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,209 @@
...
+        if not isinstance(description, (str, type(None))):
+            raise TypeError(
+                f"Description must be a string or None, find {type(description)}"
+            )
+
         self._description = description
-        self.workspace.update_attribute(self, "attributes")
+
+        if hasattr(self, "_on_file"):
```
domfournier, line 140:
```suggestion
        if self._on_file:
```

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,209 @@
...
+        The unique identifier of an entity, either as stored
         in geoh5 or generated in :func:`~uuid.UUID.uuid4` format.
         """
         return self._uid
 
     @uid.setter
-    def uid(self, uid: str | uuid.UUID):
+    def uid(self, uid: str | uuid.UUID | None):
+        if uid is None:
+            uid = uuid.uuid4()
```
domfournier, line 209:
```suggestion
    def uid(self, uid: str | uuid.UUID):
```

Should never be None

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,209 @@
...
         if isinstance(uid, str):
             uid = uuid.UUID(uid)
+        if not isinstance(uid, uuid.UUID):
+            raise TypeError(f"uid must be a string or uuid.UUID, not {type(uid)}")
 
         self._uid = uid
-        self.workspace.update_attribute(self, "attributes")
+
+        if hasattr(self, "_on_file"):
```
domfournier, line 217:
```suggestion
        if self._on_file:
```

domfournier (commented):
Great, let's just make sure we define the private attributes of the class though.  I think at some point we could make those attributes a Dataclass to avoid cluttering the classes with so many inputs. But that can come later.

geoh5py/shared/entity_type.py
MatthieuCMira: pylint/mypy doesn't dislike setting init with setter. 
It get the logic. 
Are you sure about this practice?

geoh5py/shared/entity_type.py
MatthieuCMira: The problem with this part is that ID and UID are the same parameter, Description is the same as description, Name is the same as name. 

Then, the same mechanisms are present in data_type, group_type, object_type. 

So it would mean passing 2 times the same argument with different names. 

We have to choose a mechanic here.

geoh5py/shared/entity_type.py
MatthieuCMira: agree

geoh5py/shared/entity_type.py
MatthieuCMira: It won't. In the setter, if the uid is None, then it is created. 
But we could pass it as defualt argument for sure.

geoh5py/shared/entity_type.py
MatthieuCMira: I though changing a workspace might induce a lot of mechanism. 
This is why we have a copy method, to change an object to another workspace. 

If we can do i by directly using a setter, it might be dangerous

geoh5py/shared/entity_type.py
MatthieuCMira: HAve to check. 
Can't we remove a description? 
Should we set it to ""?

geoh5py/shared/entity_type.py
MatthieuCMira: Should we have a uid setter?
I don't think the uid of an object should change in a process no?

geoh5py/shared/entity_type.py
MatthieuCMira: This mechanic is quite hacky. 
Maybe pydantic might be a solution

geoh5py/shared/entity_type.py
MatthieuCMira: I will put that in find or create

geoh5py/shared/entity_type.py
MatthieuCMira: this mechanism is now defined by find_or_create

MatthieuCMira: I can remove kwargs in entity_type. 
Should we redefine  kwargs in other init?
It would create a lot of arguments in data init for example

geoh5py/shared/entity_type.py
MatthieuCMira: This method quite hacky. 
maybe us8ing pydantic will solve the issue

geoh5py/shared/entity_type.py
MatthieuCMira: an object can have no workspace

geoh5py/shared/entity_type.py
MatthieuCMira: set register in workspace setter

geoh5py/shared/entity_type.py
MatthieuCMira: use utility function to set uid type

geoh5py/shared/concatenation/drillholes_group_table.py
```diff
@@ -205,6 +206,16 @@ def add_values_to_property_group(
...
 
+        if isinstance(value_map, dict):
+            data_type = DataType(
+                self.parent.workspace, "REFERENCED", name=name, value_map=value_map
+            )
+        else:
+            data_type_ = DataType.validate_data_type(
+                self.parent.workspace, {"values": values}
+            )["primitive_type"]
+            data_type = DataType(self.parent.workspace, data_type_, name=name)
```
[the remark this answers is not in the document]
MatthieuCMira: pass DataType

tests/drillhole_v4_0_test.py
```diff
@@ -911,12 +912,32 @@ def test_add_data_to_property(tmp_path):
...
+        )
+        verification_map_value -= np.min(verification_map_value) - 1
+
+        value_map = {idx: f"{idx}" for idx in np.unique(verification_map_value)}
+
         drillholes_table.add_values_to_property_group(
             "new value",
-            verification["interval_values_a"],
+            verification_map_value,
+            value_map=value_map,
```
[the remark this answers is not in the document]
MatthieuCMira: as DataType is now an attribute, we don't have to care about this

geoh5py/shared/concatenation/drillholes_group_table.py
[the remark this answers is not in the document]
MatthieuCMira: I agree, but petro-lingo is now using value_amp.
Can we do another issue after this one is resolved to correct it both in geoh5py and petro-lingo to prevent bugs?

geoh5py/shared/entity_type.py
MatthieuCMira: seems to work pushing to geoh5py with None. 
tested it and see no difference nor log error

geoh5py/groups/group_type.py
```diff
@@ -69,44 +71,45 @@ def allow_delete_content(self) -> bool:
...
-        else:
-            for key, val in kwargs.items():
-                if key.lower() in ["id", "uid"]:
-                    uid = uuid.UUID(val)
-
-        entity_type = cls.find(workspace, uid)
-        if entity_type is not None:
-            return entity_type
+        uid = kwargs.get("uid", None)
+        if uid is not None:
```
domfournier, line 110: We can probably reverse the order of operation to make it simpler.

- first check that uid is provided - if not replace with the entity_class default.
- call the `find` method to see if it exists already and return
- otherwise create and return

geoh5py/objects/object_base.py
```diff
@@ -344,7 +346,8 @@ def find_or_create_type(cls, workspace: Workspace, **kwargs) -> ObjectType:
 
         :return: The ObjectType instance for the given object class.
         """
-        return ObjectType.find_or_create(workspace, cls, **kwargs)
+        kwargs["entity_class"] = cls
```
domfournier, line 349: Same comment as above for the Groups, we should leave it as an explicit input since it is used directly in the `find_or_create` method

geoh5py/objects/object_type.py
```diff
@@ -49,34 +41,38 @@ def create_custom(workspace: workspace.Workspace) -> ObjectType:
...
+            kwargs["uid"] = uid
         else:
-            for key, val in kwargs.items():
-                if key.lower() in ["id", "uid"]:
-                    uid = uuid.UUID(val)
-
-        entity_type = cls.find(workspace, uid)
-        if entity_type is not None:
-            return entity_type
+            uid = kwargs.get("uid", None)
```
domfournier, line 67: Same logic as above.

geoh5py/shared/entity_type.py
```diff
@@ -20,136 +20,210 @@
...
+            and getattr(self, prop) is not None
+            and prop != "attribute_map"
+        }
+
+        attributes.update(kwargs)
+
+        if attributes.get("uid") in getattr(
+            attributes.get("workspace", self.workspace), "_types"
+        ):
+            del attributes["uid"]
```
domfournier, line 104: When would we copy the entity_type by itself? The copy should happen with the object and/or data entity no?

domfournier (commented):
I like it. See comments on explicit arguments for `find_or_create` methods.

geoh5py/groups/group_type.py
```diff
@@ -69,44 +71,45 @@ def allow_delete_content(self) -> bool:
...
+        new auto-generated UUID.
+        """
+        return GroupType(workspace, **kwargs)
+
     @classmethod
-    def find_or_create(
-        cls, workspace: workspace.Workspace, entity_class, **kwargs
-    ) -> GroupType:
-        """Find or creates an EntityType with given UUID that matches the given
+    def find_or_create(cls, workspace: Workspace, **kwargs) -> GroupType:
```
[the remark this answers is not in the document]
MatthieuCMira: I changed as this method is shared by data_type, group_type, and object_type

geoh5py/shared/entity_type.py
MatthieuCMira: I did not write this logic in the first place. 
It seems to me it's a shared function:

data_type can be copy in the case of a referenced data type. 
group_type and object_type can be copied too aren't they?

At last, during the copying, the object is copied except the changed attributes.
And the uid, of the object is not copied itself.

What is wrong with this logic?

geoh5py/data/data_type.py
```diff
@@ -49,30 +68,57 @@ class DataType(EntityType):
...
+    ):
         super().__init__(workspace, **kwargs)
+        self.color_map = color_map
+        self.hidden = hidden
+        self.mapping = mapping
+        self.number_of_bins = number_of_bins
+        self.primitive_type = primitive_type  # type: ignore
+        self.transparent_no_data = transparent_no_data
+        self.units = units
+        self.value_map = value_map  # type: ignore
```
MatthieuCMira, line 92: Is this still usefull?

geoh5py/data/data_type.py
```diff
@@ -82,237 +128,214 @@ def color_map(self) -> ColorMap | None:
...
-        :obj:`~geoh5py.data.reference_value_map.ReferenceValueMap`:
-        Reference value map for :obj:`~geoh5py.data.reference_data.ReferenceData`
+        self._color_map: ColorMap | None = color_map
 
-        The value_map can be set from a :obj:`dict` of sorted values with
-        corresponding :obj:`str` description.
+        self.workspace.update_attribute(self, "color_map")
 
-        .. code-block:: python
+    @classmethod
```
MatthieuCMira, line 155: Should we mark the creates as deprecated?

geoh5py/groups/group_type.py
```diff
@@ -69,44 +71,45 @@ def allow_delete_content(self) -> bool:
 
     @allow_delete_content.setter
     def allow_delete_content(self, allow: bool):
+        if not isinstance(allow, bool) and allow != 1 and allow != 0:
+            raise TypeError("'allow_delete_content must be a boolean.")
         self._allow_delete_content = bool(allow)
 
+    @staticmethod
```
MatthieuCMira, line 77: is this usefull?

geoh5py/objects/object_type.py
```diff
@@ -23,24 +23,16 @@
 from ..shared import EntityType
 
 if TYPE_CHECKING:
-    from .. import workspace
+    from ..workspace import Workspace
 
 
 class ObjectType(EntityType):
```
MatthieuCMira, line 28: Is this usefull? It does not have methods anymore.

geoh5py/objects/object_type.py
```diff
@@ -23,24 +23,16 @@
...
 
-    def __init__(self, workspace: workspace.Workspace, **kwargs):
-        assert workspace is not None
-        super().__init__(workspace, **kwargs)
-
-    @staticmethod
-    def _is_abstract() -> bool:
-        return False
-
     @staticmethod
```
MatthieuCMira, line 33: same as group_type. 
What create custom use for?

MatthieuCMira (commented):
comments following my push. S

geoh5py/shared/entity_type.py
domfournier: Yea, I am not sure. I guess let's leave it until something comes up.

geoh5py/data/data_type.py
MatthieuCMira: all create method mark as deprecated in entity_type

geoh5py/objects/object_type.py
MatthieuCMira: add an empty class with a __ini__ super

geoh5py/objects/object_type.py
MatthieuCMira: mark as deprecated

geoh5py/shared/entity_type.py
MatthieuCMira: not touching it

domfournier (approved):

PR #514, opened 2024-03-07
GEOPY-1349: get_data() does not work consistently with concatenatedPropertyGroup

domfournier:
**GEOPY-1349 - get_data() does not work consistently with concatenatedPropertyGroup**

### DIFF
--- a/geoh5py/io/h5_reader.py
+++ b/geoh5py/io/h5_reader.py
@@ -192,6 +192,9 @@ class H5Reader:
                 if attribute.dtype in [float, "float64", "float32"]:
                     attribute[attribute == FLOAT_NDV] = np.nan
 
+                if attribute.dtype == object and isinstance(attribute[0], bytes):
+                    attribute = attribute.astype(str)
+
                 return attribute, group["Index"][label][:]
 
             except KeyError:
--- a/geoh5py/io/h5_writer.py
+++ b/geoh5py/io/h5_writer.py
@@ -282,14 +282,17 @@ class H5Writer:
             if channel in dict_values:
                 values = dict_values[channel].copy()
 
-                if isinstance(values, np.ndarray) and values.dtype in (
-                    np.float64,
-                    np.float32,
-                ):
-                    values[np.isnan(values)] = FLOAT_NDV
-                    values = values.astype(np.float32)
-                if np.issubdtype(values.dtype, np.str_):
-                    values = values.astype(h5py.special_dtype(vlen=str))
+                if isinstance(values, np.ndarray):
+
+                    if values.dtype in (
+                        np.float64,
+                        np.float32,
+                    ):
+                        values[np.isnan(values)] = FLOAT_NDV
+                        values = values.astype(np.float32)
+
+                    if np.issubdtype(values.dtype, np.str_):
+                        values = values.astype(h5py.special_dtype(vlen=str))
 
                 attr_handle.create_dataset(
                     name,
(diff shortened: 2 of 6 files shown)


geoh5py/io/h5_reader.py
```diff
@@ -192,6 +192,9 @@ def fetch_concatenated_values(
                 if attribute.dtype in [float, "float64", "float32"]:
                     attribute[attribute == FLOAT_NDV] = np.nan
 
+                if attribute.dtype == object and isinstance(attribute[0], bytes):
```
MatthieuCMira, line 195: I don't know if it's pertinent, but to ensure no future error, we could check the type here and raise an error?

geoh5py/io/h5_writer.py
```diff
@@ -539,7 +542,11 @@ def write_array_attribute(
             except KeyError:
                 pass
 
+            if isinstance(values, np.ndarray) and np.issubdtype(values.dtype, np.str_):
```
MatthieuCMira, line 545: same, shouldn't we test the types to ensure eventual error?

geoh5py/shared/concatenation/drillhole.py
```diff
@@ -254,7 +254,7 @@ def validate_depth_data(
         else:
             out_group = property_group
 
-        _ = getattr(self, "add_data")(
+        _ = self.add_data(
```
MatthieuCMira, line 257: Why having _ = before?
Just add data is enought

geoh5py/shared/concatenation/drillhole.py
```diff
@@ -350,7 +350,7 @@ def validate_interval_data(
                 )
             return out_group
 
-        _ = getattr(self, "add_data")(
+        _ = self.add_data(
```
MatthieuCMira, line 353: same

geoh5py/workspace/workspace.py
```diff
@@ -347,24 +347,26 @@ def create_from_concatenation(self, attributes):
...
                 Data,
                 save_on_creation=False,
                 **{
                     "entity": attributes,
                     "entity_type": self.fetch_type(
-                        uuid.UUID(attributes["Type ID"]), "Data"
+                        uuid.UUID(attributes.pop("Type ID")), "Data"
                     ),
                 },
             )
```
MatthieuCMira, line 372: else? 
No error?

MatthieuCMira (approved):
Seems good for me.

Sure the code could be reworked, but that's not the scope of the issue. You create a new one as you said earlier?

geoh5py/shared/concatenation/drillhole.py
domfournier: Nice catch. Removing.

geoh5py/shared/concatenation/drillhole.py
domfournier: Done

geoh5py/workspace/workspace.py
domfournier: It added "Type ID" on the class attributes, which not great, but would also make the `compare_entities` to crash in tests

geoh5py/io/h5_reader.py
domfournier: Well technically the format should be correct since it comes from geoh5. Not sure I am following.

MatthieuCMira: Ok then not type testing

geoh5py/io/h5_writer.py
MatthieuCMira: Guess the previous answer is the same here.

geoh5py/workspace/workspace.py
MatthieuCMira: perfect, mark this as resolved

geoh5py/io/h5_writer.py
domfournier: Good idea, but will be pretty complex to check since there are so many formats passing through. We would need serious validation mechanics

MatthieuCMira: ok, guess letting it as it is

MatthieuCMira (approved):

PR #515, opened 2024-03-07
GEOPY-1346: cannot removed concatenated data using object.remove() method

domfournier:
**GEOPY-1346 - cannot removed concatenated data using object.remove() method**

### DIFF
--- a/geoh5py/groups/property_group.py
+++ b/geoh5py/groups/property_group.py
@@ -236,18 +236,22 @@ class PropertyGroup(ABC):
         if self._properties is None:
             return
 
+        update = False
         for elem in data:
-            if isinstance(elem, uuid.UUID):
-                entity = self.parent.get_entity(elem)[0]
-            elif isinstance(elem, Data) and elem in self.parent.children:
-                entity = elem
+
+            if isinstance(elem, Data):
+                uid = elem.uid
             else:
-                continue
+                uid = elem
+
+            entity = self.parent.get_entity(uid)[0]
 
             if entity is not None and entity.uid in self._properties:
+                update = True
                 self._properties.remove(entity.uid)
 
-        self.parent.workspace.add_or_update_property_group(self)
+        if update:
+            self.parent.workspace.add_or_update_property_group(self)
 
     @property
     def uid(self) -> uuid.UUID:
--- a/geoh5py/groups/property_group.py
+++ b/geoh5py/groups/property_group.py
@@ -240,15 +240,11 @@ class PropertyGroup(ABC):
         for elem in data:
 
             if isinstance(elem, Data):
-                uid = elem.uid
-            else:
-                uid = elem
+                elem = elem.uid
 
-            entity = self.parent.get_entity(uid)[0]
-
-            if entity is not None and entity.uid in self._properties:
+            if elem in self._properties:
                 update = True
-                self._properties.remove(entity.uid)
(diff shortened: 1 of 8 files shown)


geoh5py/groups/property_group.py
```diff
@@ -236,18 +236,22 @@ def remove_properties(self, data: Data | list[Data | uuid.UUID] | uuid.UUID):
...
             return
 
+        update = False
         for elem in data:
-            if isinstance(elem, uuid.UUID):
-                entity = self.parent.get_entity(elem)[0]
-            elif isinstance(elem, Data) and elem in self.parent.children:
-                entity = elem
+
+            if isinstance(elem, Data):
```
MatthieuCMira, line 242: isn't a bit repetitive with line 247
get uid then get the data ?

tests/drillhole_v4_0_test.py
```diff
@@ -611,11 +612,19 @@ def test_remove_drillhole_data(tmp_path):
         well = workspace.get_entity("well")[0]
         well_b = workspace.get_entity("Number 2")[0]
         data = well.get_data("my_log_values/")[0]
-        workspace.remove_entity(data)
-        workspace.remove_entity(well_b)
 
+        well.remove_children(data)
+        del data
```
MatthieuCMira, line 617: shouldn't we del the data in the remove option?

MatthieuCMira (approved):
not a lot of changes. 
Just 2 small comments

geoh5py/groups/property_group.py
domfournier: I can simplify.

tests/drillhole_v4_0_test.py
domfournier: This is to remove the variable within the context of the test at runtime. Otherwise the test itself has a reference to the object, and the garbage collection won't clean it out, and the weakref with survive.

domfournier (commented):
Ready for round 2

MatthieuCMira (dismissed):
ok for me

MatthieuCMira (approved):
Those changes seems ok to me, going to test

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -488,8 +495,12 @@ def remove_entity(self, entity: Concatenated):
                 object_ids.remove(as_str_if_uuid(entity.uid).encode())
                 self.concatenated_object_ids = object_ids
 
-        if self.concatenated_attributes is not None:
+        if (
```
MatthieuCMira, line 498: It 's not for this review but I'm more and more sceptic about those "

def __init__(self):
   self._attr: mytype | None = None
   
and 

@property
def attr(self):
   if self._attr is None:
      # do something
    else return self._attr
    
It feels like it's not updating, introducing unecessary types management.

MatthieuCMira (approved):

geoh5py/shared/concatenation/concatenator.py
domfournier: That pattern can be nice to just force update by deleting the _attr. But I agree, can be dangerous. It's a case by case I guess.

MatthieuCMira: Then an updating function?
I don't know, maybe that's something to have to our "good practice" and decide once and for all* 

*can of course change in time ;-)

PR #516, opened 2024-03-08
support for wellpathpy for minimum curvature desurvey

collord:


domfournier:
Hi @collord!
Thank you for your contribution. We are looking into it, and brainstorming ideas on ways how to integrate it. We are likely going to make it an optional requirement, and let user chose which algo they want to use for desurveying.

We will get back to you shortly.

cardinalgeo:
I'm very excited about this added functionality and hope y'all accept it @domfournier! Good stuff @collord :)

PR #519, opened 2024-03-18
GEOPY-1112: b

MatthieuCMira:
**GEOPY-1112 - Implement geoh5py ui.json range slider form**
Use rangeslider with geoh5py

### DIFF
--- a/geoh5py/ui_json/input_file.py
+++ b/geoh5py/ui_json/input_file.py
@@ -274,7 +274,10 @@ class InputFile:
                 if (value is None) and (not self.ui_json[key].get("enabled", False)):
                     continue
 
-                self.ui_json[key][member] = value
+                if isinstance(value, dict):
+                    self.ui_json[key].update(value)
+                else:
+                    self.ui_json[key][member] = value
 
             else:
                 self.ui_json[key] = value
--- a/geoh5py/ui_json/input_file.py
+++ b/geoh5py/ui_json/input_file.py
@@ -274,11 +274,7 @@ class InputFile:
                 if (value is None) and (not self.ui_json[key].get("enabled", False)):
                     continue
 
-                if isinstance(value, dict):
-                    self.ui_json[key].update(value)
-                else:
-                    self.ui_json[key][member] = value
-
+                self.ui_json[key][member] = value
             else:
                 self.ui_json[key] = value
 
--- a/geoh5py/ui_json/input_file.py
+++ b/geoh5py/ui_json/input_file.py
@@ -107,6 +107,9 @@ class InputFile:
 
     @property
     def data(self) -> dict[str, Any] | None:
+        """
+        Dictionary representing the input data for the ui.json file.
+        """
         if self._data is None and self.ui_json is not None:
             self.data = flatten(self.ui_json)
 
--- a/geoh5py/ui_json/utils.py
+++ b/geoh5py/ui_json/utils.py
@@ -42,18 +42,7 @@ def flatten(ui_json: dict[str, dict]) -> dict[str, Any]:
                 if not truth(ui_json, name, "enabled"):
                     data[name] = None
                 else:
-                    temp_value = value[field]
-
-                    if value.get("groupValue"):
-                        temp_value = {"group": value["groupValue"], "value": temp_value}
-                    elif value.get("rangeLabel"):
-                        temp_value = {
-                            "property": value["property"],
-                            "value": value["value"],
-                            "isComplement": value["isComplement"],
-                        }
-
-                    data[name] = temp_value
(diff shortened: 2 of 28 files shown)


tests/ui_json_utils_test.py
```diff
@@ -298,6 +298,23 @@ def test_flatten_group_value():
     assert flat["test"] is None
 
 
+def test_range_label():
+    ui_json = deepcopy(default_ui_json)
```
benk-mira, line 302: I was trying to get familiar with the new template and decided to create a points object with data and write out an input file with the new type in there.  Seems to be something wrong with the format:
image
Here's the code that reproduces:
image
Strangely, Analyst seems to handle it ok, we just can't turn around and read back in.  Am I missing something?  Is this expected?

benk-mira (commented):
Maybe something fishy with the format on write out?

tests/ui_json_utils_test.py
MatthieuCMira: The problem doen't seems to come from the uijson but from "update_ui_values" in input_file (l.245)

It change the uijson.
Expected:
````
  "target": {
    "main": true,
    "label": "Threshold",
    "allowComplement": false,
    "isComplement": false,
    "parent": "grid_object",
    "property": "{e9567358-a3f9-4b46-8b5d-ce18ab4740b8}",
    "association": [
      "Cell",
      "Vertex"
    ],
    "dataType": [
      "Referenced",
      "Integer",
      "Float"
    ],
    "value": [
      0.49085091490850913,
      1.0
    ],
    "rangeLabel": "Values",
    "enabled": true
  },
````

obtained in the test:

````
    "test": {
        "main": true,
        "label": "Range",
        "allowComplement": false,
        "isComplement": false,
        "parent": "object",
        "property": "{bd70a785-7f51-48bc-9875-47afbcd879d6}",
        "association": "Vertex",
        "dataType": "Float",
        "value": {
            "property": "{bd70a785-7f51-48bc-9875-47afbcd879d6}",
            "value": [
                0.2,
                0.8
            ],
            "isComplement": false
        },
        "rangeLabel": "Values",
        "enabled": true
    }
````

Comment (MatthieuCMira) [reply tests/ui_json_utils_test.py:302]: it warps up the value in a sub-dictionary. 
But I don't know why...

Comment (MatthieuCMira) [reply tests/ui_json_utils_test.py:302]: I corrected it handling the exception for this widget.

File: geoh5py/ui_json/utils.py [hunk 2]
```diff
@@ -46,6 +46,12 @@ def flatten(ui_json: dict[str, dict]) -> dict[str, Any]:
 
                     if value.get("groupValue"):
                         temp_value = {"group": value["groupValue"], "value": temp_value}
+                    elif value.get("rangeLabel"):
```
benk-mira, line 49: I guess I didn't recognize the significance of this in the first pass.  The output or flatten should be a flat (not nested) dictionary.  So temp_value here should just be value["value"] and then any code needing to access the state (property and isComplement) of the parameter should refer to the ui_json dictionary attribute

benk-mira: If you fix this here, then you can also remove the dict handler in update_ui_values

geoh5py/ui_json/input_file.py
```diff
@@ -274,7 +274,10 @@ def update_ui_values(self, data: dict):
                 if (value is None) and (not self.ui_json[key].get("enabled", False)):
                     continue
 
-                self.ui_json[key][member] = value
+                if isinstance(value, dict):
```
benk-mira, line 277: See note below.  This function expects a flat dictionary, so there shouldn't be any need to handle a dictionary type here.

benk-mira (commented):
Hey, the fix you provided seems to point to a bigger problem.  See notes of flat dictionaries.  I also noticed that the range was not draggable in the ui.  This should be draggable, right:
image

geoh5py/ui_json/input_file.py
MatthieuCMira: Normally, the "value" is not a dictionary. 
It's only a dictionary for the range slider widget.

MatthieuCMira: It's a bit more complicated. 

The widgets were returning only one object: in "value"

But with more complex widgets, it returns more than one value...

MatthieuCMira:
I removed the dictionary handler. 

But now, data is obsolete with the new widgets. 
Something to consider in future works.

benk-mira (approved):
Great, thanks for this, hope it's not too much burden on the apps!

domfournier (approved):

PR #520, opened 2024-03-19
GEOPY-1423: Remove unnecessary line

domfournier:
**GEOPY-1423 - Issue with workspace on compute_trigger of edge detection app**


MatthieuCMira (approved):
going to test if we can pass with both

PR #522, opened 2024-03-27
GEOPY-1414: add property group suffix to las export files

domfournier:
**GEOPY-1414 - add property group suffix to las export files**


MatthieuCMira (approved):
Seems good for me, needs testing

PR #529, opened 2024-04-04
GEOPY-1441: Suppress unnecessary warnings triggered in set_enabled  

benk-mira:
**GEOPY-1441 - Suppress unnecessary warnings triggered in set_enabled  **


domfournier (approved):
Makes sense.

PR #532, opened 2024-04-12
GEOPY-1418: Clean out empty property groups on drillhole objects after data removal

domfournier:
**GEOPY-1418 - Clean out empty property groups on drillhole objects after data removal**

### DIFF
--- a/geoh5py/shared/concatenation/concatenator.py
+++ b/geoh5py/shared/concatenation/concatenator.py
@@ -476,18 +476,23 @@ class Concatenator(Group):  # pylint: disable=too-many-public-methods
     def remove_entity(self, entity: Concatenated | ConcatenatedPropertyGroup):
         """Remove a concatenated entity."""
 
+        parent = entity.parent
         if isinstance(entity, ConcatenatedData):
             # Remove the rows of data and index
             self.update_array_attribute(entity, entity.name, remove=True)
+            # Remove the data from the group
+
+            if entity.property_group is not None:
+                entity.property_group.remove_properties([entity])
+
             # Remove from the concatenated Attributes
-            parent_attr = self.get_concatenated_attributes(entity.parent.uid)
+            parent_attr = self.get_concatenated_attributes(parent.uid)
             name = entity.name
             del parent_attr[f"Property:{name}"]
+
         elif isinstance(entity, ConcatenatedObject):
-
-            for child in entity.children.copy():
-                self.remove_entity(child)
-
+            # First remove the children
+            entity.remove_children(entity.children.copy())
             object_ids = self.concatenated_object_ids
 
             if object_ids is not None:
@@ -495,7 +500,12 @@ class Concatenator(Group):  # pylint: disable=too-many-public-methods
                 self.concatenated_object_ids = object_ids
 
         elif isinstance(entity, ConcatenatedPropertyGroup):
-            self.update_array_attribute(entity.parent, "property_groups", remove=True)
+            # Remove all data within the group
+            if entity.properties is not None and len(entity.properties) > 0:
+                data = [entity.parent.get_entity(uid)[0] for uid in entity.properties]
+                entity.parent.remove_children(data)
+
+            self.update_array_attribute(parent, "property_groups", remove=True)
 
         if (
             self.concatenated_attributes is not None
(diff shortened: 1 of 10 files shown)


geoh5py/shared/concatenation/concatenator.py
```diff
@@ -472,30 +473,40 @@ def remove_children(self, children: list | Concatenated):
 
             self.remove_entity(child)
 
-    def remove_entity(self, entity: Concatenated):
+    def remove_entity(self, entity: Concatenated | ConcatenatedPropertyGroup):
```
MatthieuCMira, line 476: In a drillholes group, can we remove an entity (a drillhole), but keep the similar properties on the other drillholes? Then we don't want to drop the property group, but just drop one property to the property group?

Unless each drillhole has a property group, and the name of the property group is recognize as a key by GA. (I can't remember the mechanics precisely).

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -472,30 +473,40 @@ def remove_children(self, children: list | Concatenated):
...
-
-            for child in entity.children.copy():
-                self.remove_entity(child)
-
-            if entity.property_groups is not None:  # type: ignore
-                self.update_array_attribute(entity, "property_groups", remove=True)
 
+        elif isinstance(entity, ConcatenatedObject):
+            # First remove the children
+            entity.remove_children(entity.children.copy())
```
MatthieuCMira, line 495: ???

This line is strange. Don't we want to drop the children based on uid, not on "copy"?

geoh5py/shared/concatenation/object.py
```diff
@@ -162,7 +162,9 @@ def property_groups(self) -> list | None:
 
         return self._property_groups
 
-    def remove_children(self, children: list | Concatenated):
+    def remove_children(
```
MatthieuCMira, line 165: ok, we need the object, not the uids.

Still, why using "copy" above?

geoh5py/shared/concatenation/concatenator.py
```diff
@@ -472,30 +473,40 @@ def remove_children(self, children: list | Concatenated):
...
+            # Remove the data from the group
+
+            if entity.property_group is not None:
+                entity.property_group.remove_properties([entity])
+
             # Remove from the concatenated Attributes
-            parent_attr = self.get_concatenated_attributes(entity.parent.uid)
+            parent_attr = self.get_concatenated_attributes(parent.uid)
             name = entity.name
             del parent_attr[f"Property:{name}"]
```
MatthieuCMira, line 491: ok, we need the object, not the uids.

Still, why using "copy" above?

tests/drillhole_v4_0_test.py
```diff
@@ -630,6 +630,17 @@ def test_remove_drillhole_data(tmp_path):
...
+    with Workspace(h5file_path, version=2.0) as workspace:
+        well = workspace.get_entity("well")[0]
+        all_data = [well.get_entity(name)[0] for name in well.get_data_list()]
+        all_data = [data for data in all_data if data.allow_delete]
+        well.remove_children(all_data)
+
+        assert len(well.property_groups) == 0
+
+        well = workspace.get_entity("Number 3")[0]
+        well.remove_children(well.property_groups[0])
```
MatthieuCMira, line 642: don't you want to assert it has been removed?

MatthieuCMira (commented):
Small comments.
I probably not getting everything, just want to ensure the logic ;-)

geoh5py/shared/concatenation/concatenator.py
domfournier: That's correct, property groups are unique on each drillhole. Mechanics are more restrictive than on regular objects since we can't have data not belonging to a prop_group. So if remove the group, we also need to delete the data underneath

geoh5py/shared/concatenation/concatenator.py
domfournier: This is just a copy of the list. Issues arise otherwise in the next methods where data are removed from the same children list (list modified within the loop)

geoh5py/shared/concatenation/object.py
domfournier: see comment

tests/drillhole_v4_0_test.py
domfournier: yep, good point

geoh5py/shared/concatenation/concatenator.py
MatthieuCMira: ok with that.
Marking it as resolved

MatthieuCMira: ok with that.
Marking it as resolved

MatthieuCMira (approved):
ok for me.

PR #538, opened 2024-04-17
GEOPY-1418: Avoid removing empty arrays

domfournier:
**GEOPY-1418 - Clean out empty property groups on drillhole objects after data removal**


MatthieuCMira (changes requested):
problem with files on drillhole group.
Files get supressed

MatthieuCMira (approved):
ok.
I tested it on Petro-lingo and it works!
Well done!

PR #539, opened 2024-04-18
GEOPY-1450: Geoimage in geoh5py import tiff return black image

MatthieuCMira:
**GEOPY-1450 - Geoimage in geoh5py import tiff return black image**
Change the conversion from geoimage to grid, not converting data type anymore.
If the input image is a tiff, the data is automaticly duplicated as a grid

### DIFF
--- a/geoh5py/objects/geo_image.py
+++ b/geoh5py/objects/geo_image.py
@@ -351,6 +351,7 @@ class GeoImage(ObjectBase):
 
         :return: List of new Data objects.
         """
+        convert_to_grid = False
         if isinstance(image, np.ndarray) and image.ndim in [2, 3]:
             if image.ndim == 3 and image.shape[2] != 3:
                 raise ValueError(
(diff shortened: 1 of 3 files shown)


geoh5py/objects/geo_image.py
```diff
@@ -371,15 +372,16 @@ def image(self, image: str | np.ndarray | BytesIO | Image.Image):
                 raise ValueError(f"Input image file {image} does not exist.")
 
             image = Image.open(image)
-
             # if the image is a tiff save tag information
             if isinstance(image, TiffImageFile):
                 self.tag = image
+                convert_to_grid = True
```
domfournier, line 378: Combine duplicate logic. All of these could be just if statements that converts nd.array or str or bytes to an image. Then do the checks for is `Image` and is `TiffImageFile` once

tests/geo_image_test.py
```diff
@@ -586,3 +586,23 @@ def test_copy_from_extent_geoimage(tmp_path):
...
+    with Workspace.create(workspace_path) as workspace:
+        # load image
+        geoimage = GeoImage.create(workspace, name="test_area", image=str(image_path))
+
+        grid = workspace.get_entity("test_area_grid2d")[0]
+
+        assert all(
+            np.array(geoimage.image)[::-1].flatten()
+            == grid.get_data("band[0]")[0].values
+        )
```
domfournier, line 608: Can we add a check for multi-bands too?

domfournier (commented):
Simple, that's great. See comments

tests/geo_image_test.py
MatthieuCMira: PIL images can't handle >4 bands. 
That's why people are using Gdal

MatthieuCMira: Actually, I can play with more than 4 bands, but that's another level of complexity. 
I created an issue to refactor geoimage for tiff.

MatthieuCMira: I'll test for 4 bands here, and we will implement n bands later

domfournier: Ok doke

andrewg-mira (approved):

PR #542, opened 2024-04-24
GEOPY-1472: prepare for distribution files

sebhmg:
GEOPY-1472: prepare for distribution files

relates to https://github.com/MiraGeoscience/geopy-distrib/pull/1


pyproject.toml
```diff
@@ -20,21 +20,32 @@ classifiers = [
...
 
 
 [tool.poetry.group.dev.dependencies]
+Pygments = "*"
 lockfile = "^0.12.2"
 pylint = "*"
 pytest = "*"
 pytest-cov = "*"
-scipy = "^1.4.1"
+scipy = "~1.10.1"
```
sebhmg, line 48: @domfournier , tests are failing. Could you please give it a look?
Maybe due to the change of scipy version here?

Note that SciPy is only used in tests for this repo. I have changed its version, so that tests run with that same version as the one we ship in the runtime environment

sebhmg: thanks @domfournier for pointing at Pillow. I will constrain its version harder.

domfournier: That's actually a PIL issue. We need to pin it back for now.
To addressed on https://mirageoscience.atlassian.net/browse/GEOPY-1456

domfournier:
Need re-point autoflake

@benk-mira 
image

sebhmg:
> Need re-point autoflake
> 
> @benk-mira image

just removing it. Its scope was limited and likely covered by other tools.
See https://inventwithpython.com/blog/2022/11/19/python-linter-comparison-2022-pylint-vs-pyflakes-vs-flake8-vs-autopep8-vs-bandit-vs-prospector-vs-pylama-vs-pyroma-vs-black-vs-mypy-vs-radon-vs-mccabe/#:~:text=pass%20on%20eradicate.-,autoflake,-(Dead%20Code%20Linter

PR #549, opened 2024-05-02
GEOPY-1456: Update Pillow to ^10.1

MatthieuCMira:
**GEOPY-1456 - Update Pillow to ^10.1**
update for pillow 10.3


tests/coordinate_system_test.py
```diff
@@ -49,12 +48,13 @@ def test_coordinate_system(tmp_path):
...
     image = Image.fromarray(
-        np.random.randint(0, 255, (128, 128, 3)).astype("uint8"), "RGB"
+        np.random.randint(0, 255, (128, 128, 3)).astype("uint16"), "RGB"
     )
 
     for id_ in tag.items():
         image.getexif()[id_[0]] = id_[1]
-    image.save(tmp_path / r"testtif.tif", exif=image.getexif())
+
+    image.save(tmp_path / r"testtif.tif", exif=image.getexif())  # working with 10.1
```
domfournier, line 57:
```suggestion
    image.save(tmp_path / r"testtif.tif", exif=image.getexif())
```

domfournier (approved):

MatthieuCMira:
Merging?

domfournier:
@sebhmg This one never made it somehow. Guess we can add it to the last minute updates.

pyproject.toml
```diff
@@ -41,7 +41,7 @@ python = "^3.9,<3.11"
 
 h5py = "^3.2.1"
 numpy = "!=1.19.4, ~1.23.5"
-Pillow = "~10.1.0"
+Pillow = "^10.3.0"
```
sebhmg, line 44: **thought:** specifying `^10.3.0` allows any `10.x` version, while `~10.3.0` is any `10.3.x`

In the past, updating to a minor version broke the behavior at runtime. I would tend to limit to `~10.3.0`
Alos relocking  a repo with `^10.3.0` spec might lock on `10.5`, while a repo which has not been relocked yet will still use `10.3.0`. that create more discrepancy between repos.

it is quite depends on packages though. Some have more conservative policy on changing minor version.

sebhmg (changes requested):
see suggestion about using version spec with `~` instead of `^`

pyproject.toml
domfournier, line 44:
```suggestion
Pillow = "~10.3.0"
```

domfournier (approved):

sebhmg (approved):

PR #558, opened 2024-05-28
GEOPY-1530: Convert nan first

domfournier:
**GEOPY-1530 - Use UIJson group **


MatthieuCMira (dismissed):
it's allright

benk-mira (approved):
ok, makes sense. We should probably still protect the .options on plate-simulation with a deepcopy anywhere where we use it.

PR #560, opened 2024-06-04
GEOPY-1562: Drillhole does not import into existing drillhole group containing one or more drillholes

domfournier:
**GEOPY-1562 - Drillhole does not import into existing drillhole group containing one or more drillholes**


benk-mira (changes requested):
small suggestions.  Thanks for the fix!

benk-mira (dismissed):

sebhmg (changes requested):
better having some unit tests to cover the case that was failing

PR #551, opened 2024-05-03
Support for python 3.11 and 3.12 [GEOPY-1494]

SamuelBradley:
**GEOPY-1494 - Python 3.11 (and 3.12) Support**
Made the following changes
- Update Poetry file to allow python version <4 (Good practise to not arbitrarily restrict use with new minor python versions)
- Update lock file
- Run tests again python 3.12.1
- Update Github actions to run tests in 3.11 and 3.12

PR #561, opened 2024-06-05
GEOPY-1494: Python 3.11 (and 3.12) Support - drop Python 3.9

domfournier:
**GEOPY-1494 - Python 3.11 (and 3.12) Support**

Include changes from #551


domfournier:
@sebhmg Not to be merged in, but same PR will go on https://github.com/SamuelBradley/geoh5py/pull/1

tests/ui_json_test.py
```diff
@@ -540,7 +540,7 @@ def test_input_file(tmp_path: Path):
     ):
         InputFile.read_ui_json("somefile.json")
 
-    with pytest.raises(TypeError, match="expected str, bytes or os.PathLike object"):
+    with pytest.raises(TypeError):
```
sebhmg, line 543: **question:** why removing the test of the error message?

`match` can take a regex if it helps to have the test more resistant to minor changes in the message text

pyproject.toml
```diff
@@ -37,11 +37,13 @@ include = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.9,<3.11"
-
+python = "^3.9,<3.13"
+Pillow = "10.1.0"
```
sebhmg, line 41: **suggestion:** first merge release branch to dev, to have Pillow "~10.3.0"

sebhmg (approved):

pyproject.toml
sebhmg: or merge can happen after, but probably good to have it now and make sure tests are passing fine with newer python versions

tests/ui_json_test.py
domfournier: Looks like the error message returned by pathlib in version 3.12 has changed substantially.

sebhmg: ho, if the message is not coming from the code of the repo itself, checking for the error text is less relevant. As long as it captures a TypeError

sebhmg:
also another couple of changes please:
* 3.9 -> 3.10 in `.github/workflows/static-analysis.yml`
* `--py310-plus` for pyupgrade in `.pre-commit-config.yaml`

domfournier:
To be continued on #551

PR #568, opened 2024-06-26
GEOPY-1602: Refactor input arguments of class instantiation

domfournier:
**GEOPY-1602 - Refactor input arguments of class instantiation**

Main changes proposed on this PR.
- Explicit listing of attributes in the constructor of the class that are needed for the object to be valid
- Provide defaults for all attributes to construct a "unit" object if omitted.
- Remove the setter for attributes that should not be modified directly that would affect data (vertices, cells, *_count, octree_cells etc.)


sebhmg (commented):
I have skimmed through. Looks good!
I like the simplification of some parameter types

I will leave @benk-mira or @MatthieuCMira give it a more thorough review

geoh5py/objects/drape_model.py
```diff
@@ -113,37 +119,14 @@ def layers(self) -> np.ndarray | None:
...
-            np.core.records.fromarrays(
-                xyz.T.tolist(),
-                dtype=[("I", "<i4"), ("K", "<i4"), ("Bottom elevation", "<f8")],
-            )
-        )
-        self.workspace.update_attribute(self, "layers")
+        return np.asarray(self._layers.tolist())
 
     @property
     def n_cells(self):
```
MatthieuCMira, line 125: typing

geoh5py/objects/drape_model.py
```diff
@@ -164,29 +147,106 @@ def prisms(self) -> np.ndarray | None:
         if self._prisms is None and self.on_file:
             self._prisms = self.workspace.fetch_array_attribute(self, "prisms")
 
-        if self._prisms is not None:
-            return np.array(self._prisms.tolist())
+        return np.array(self._prisms.tolist())
 
+    @property
```
MatthieuCMira, line 152: you forgot types in the documents

geoh5py/objects/drape_model.py
```diff
@@ -164,29 +147,106 @@ def prisms(self) -> np.ndarray | None:
...
+    def rotation(self):
```
MatthieuCMira, line 153: I guess there is a good reason for these functions to exist?
As they are not used ghere, could be nice to say why they are here for programmers understanding

MatthieuCMira (changes requested):
PARTIAL Review

review for block model, cell object, curve and drape model.

I'll rest my eyes a bit before finishing the review

geoh5py/objects/block_model.py
```diff
@@ -92,7 +108,6 @@ def centroids(self) -> np.ndarray | None:
             xyz = np.c_[np.ravel(u_grid), np.ravel(v_grid), np.ravel(z_grid)]
 
             self._centroids = np.dot(rot, xyz.T).T
```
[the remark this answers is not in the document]
domfournier: correct, dip not implemented in GA for block model

geoh5py/objects/curve.py
```diff
@@ -152,28 +151,69 @@ def parts(self):
...
+
+        if len(indices) != self.n_vertices:
+            raise ValueError(
+                f"Provided parts must be of shape {self.vertices.shape[0]}"
+            )
+
+        return indices
+
+    @classmethod
+    def validate_vertices(cls, xyz: np.ndarray | list | tuple) -> np.ndarray:
```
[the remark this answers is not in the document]
domfournier: Yep, good call. I noticed the same with surface objects.

geoh5py/objects/block_model.py
```diff
@@ -115,162 +130,83 @@ def default_type_uid(cls) -> uuid.UUID:
         return cls.__TYPE_UID
```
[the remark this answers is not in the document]
domfournier: I would be tempted to deprecate the Grid2D attributes instead, but I can create a method here for now.

geoh5py/objects/block_model.py
```diff
@@ -67,13 +89,7 @@ def centroids(self) -> np.ndarray | None:
...
         """
-        if (
-            getattr(self, "_centroids", None) is None
-            and self.u_cells is not None
-            and self.v_cells is not None
-            and self.z_cells is not None
-            and self.origin is not None
-        ):
+        if getattr(self, "_centroids", None) is None:
             cell_center_u = np.cumsum(self.u_cells) - self.u_cells / 2.0
```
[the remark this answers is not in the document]
domfournier: I would be tempted to deprecate the Grid2D attributes instead, but I can create a method here for now.

geoh5py/objects/block_model.py
```diff
@@ -115,162 +130,83 @@ def default_type_uid(cls) -> uuid.UUID:
...
-            self.u_cells is not None
-            and self.v_cells is not None
-            and self.z_cells is not None
-        ):
-            return tuple(
-                [self.u_cells.shape[0], self.v_cells.shape[0], self.z_cells.shape[0]]
-            )
-        return None
+        return tuple(
+            [self.u_cells.shape[0], self.v_cells.shape[0], self.z_cells.shape[0]]
```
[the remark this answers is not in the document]
domfournier: shape is a stashed property, so only computed once, but can be used in several places. And you could also use it to reshape your data values like `values.reshape(mesh.shape)`

geoh5py/objects/block_model.py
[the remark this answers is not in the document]
domfournier: Done, refactored to use the base GridObject class

geoh5py/objects/block_model.py
```diff
@@ -115,162 +130,83 @@ def default_type_uid(cls) -> uuid.UUID:
...
+        )
 
     @property
-    def u_cell_delimiters(self) -> np.ndarray | None:
+    def u_cell_delimiters(self) -> np.ndarray:
         """
         :obj:`numpy.array` of :obj:`float`:
```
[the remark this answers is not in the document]
domfournier: Yep, tried to fix everywhere.

geoh5py/objects/block_model.py
```diff
@@ -115,162 +130,83 @@ def default_type_uid(cls) -> uuid.UUID:
...
         shape (:obj:`~geoh5py.objects.block_model.BlockModel.shape` [2], ):
         Cell size along the z-axis
         """
-        if self.z_cell_delimiters is not None:
-            return self.z_cell_delimiters[1:] - self.z_cell_delimiters[:-1]
-        return None
+        return self.z_cell_delimiters[1:] - self.z_cell_delimiters[:-1]
+
+    @staticmethod
+    def validate_cell_delimiters(value: np.ndarray, axis: str) -> np.ndarray:
```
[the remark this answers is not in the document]
domfournier: Done

geoh5py/objects/cell_object.py
```diff
@@ -20,36 +20,47 @@
...
 from ..shared.utils import box_intersect, mask_by_extent
+from .object_base import ObjectType
 from .points import Points
 
-if TYPE_CHECKING:
-    from geoh5py.objects import ObjectType
-
 
 class CellObject(Points, ABC):
     """
```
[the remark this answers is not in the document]
domfournier: Done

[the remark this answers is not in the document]
domfournier: Both surface and curves have cells, which define the connection between vertices. Main difference is that Curve connects pairs, while Surface connects triplets

geoh5py/objects/cell_object.py
```diff
@@ -20,36 +20,47 @@
...
 
     _attribute_map: dict = Points._attribute_map.copy()
+    _attribute_map.update(
+        {
+            "Cells": "cells",
+        }
+    )
+
+    def __init__(
+        self, object_type: ObjectType, cells: np.ndarray | None = None, **kwargs
```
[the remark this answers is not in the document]
domfournier: Cells can be defined with defaults (connected in order), or by parts. So it can be None on init, but the object will always return an array.

geoh5py/objects/curve.py
```diff
@@ -43,65 +43,53 @@ class Curve(CellObject):
         fields=(0x6A057FDC, 0xB355, 0x11E3, 0x95, 0xBE, 0xFD84A7FFCB88)
     )
 
-    def __init__(self, object_type: ObjectType, name="Curve", **kwargs):
+    def __init__(  # pylint: disable="too-many-arguments"
```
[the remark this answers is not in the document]
domfournier: Added

geoh5py/objects/cell_object.py
```diff
@@ -100,10 +110,10 @@ def remove_cells(self, indices: list[int] | np.ndarray, clear_cache: bool = Fals
...
 
         cells = np.delete(self.cells, indices, axis=0)
-        self._cells = None
-        setattr(self, "cells", cells)
 
+        self._cells = self.validate_cells(cells)
         self.remove_children_values(indices, "CELL", clear_cache=clear_cache)
+        self.workspace.update_attribute(self, "cells")
 
     def remove_vertices(
```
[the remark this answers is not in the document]
domfournier: Done

geoh5py/objects/cell_object.py
```diff
@@ -134,15 +144,17 @@ def remove_vertices(
         vert_index = np.ones(self.vertices.shape[0], dtype=bool)
         vert_index[indices] = False
         vertices = self.vertices[vert_index, :]
-
-        self._vertices = None
-        setattr(self, "vertices", vertices)
+        self._vertices = self.validate_vertices(vertices)
```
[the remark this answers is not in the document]
domfournier: Good point.

geoh5py/objects/block_model.py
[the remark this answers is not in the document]
domfournier: You got it

geoh5py/objects/block_model.py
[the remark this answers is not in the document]
domfournier: Yea I hear you. As discussed though, I would vote against making the objects pydantic classes as they are more procedural classes. Attributes should definitely become Dataclass at some point.

geoh5py/objects/cell_object.py
```diff
@@ -100,10 +105,10 @@ def remove_cells(self, indices: list[int] | np.ndarray, clear_cache: bool = Fals
             raise ValueError("Found indices larger than the number of cells.")
```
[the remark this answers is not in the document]
domfournier: Can never be none at this stage.

geoh5py/objects/cell_object.py
```diff
@@ -134,15 +139,17 @@ def remove_vertices(
         vert_index = np.ones(self.vertices.shape[0], dtype=bool)
```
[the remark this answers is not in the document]
domfournier: Also can't be None

geoh5py/objects/curve.py
```diff
@@ -43,65 +43,53 @@ class Curve(CellObject):
...
+        self,
+        object_type: ObjectType,
+        vertices: np.ndarray = ((0.0, 0.0, 0.0), (1.0, 1.0, 1.0)),
+        cells: np.ndarray | None = None,
```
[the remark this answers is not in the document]
domfournier: The validate_cells deals with it.

geoh5py/objects/curve.py
```diff
@@ -43,65 +43,53 @@ class Curve(CellObject):
...
+    ):
         self._current_line_id: uuid.UUID | None = None
-        self._parts: np.ndarray | None = None
-        super().__init__(object_type, name=name, **kwargs)
+        self._parts: np.ndarray | None
+        self._vertices: np.ndarray = self.validate_vertices(vertices)
+
+        if parts is not None:
+            if cells is not None:
+                raise ValueError("Parts can only be set if cells are not provided.")
```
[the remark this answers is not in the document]
domfournier: Trying to supply both at the same time. Either you make the cells from parts (on vertices), or cells directly. Ambiguous to know which one to use otherwise

geoh5py/objects/drape_model.py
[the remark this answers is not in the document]
domfournier: Changed base class

geoh5py/objects/octree.py
```diff
@@ -189,211 +196,121 @@ def octree_cells(self) -> np.ndarray | None:
...
-            value = np.r_[value]
-            assert len(value) == 1, "Rotation angle must be a float of shape (1,)"
-            self._centroids = None
-            self._rotation = value.astype(float).item()
-            self.workspace.update_attribute(self, "attributes")
+        return self._octree_cells
 
     @property
-    def shape(self) -> tuple | None:
+    def shape(self) -> tuple:
```
benk-mira, line 207:
```suggestion
    def shape(self) -> tuple[int, int, int]:
```

geoh5py/objects/points.py
```diff
@@ -47,60 +81,22 @@ def extent(self) -> np.ndarray | None:
...
-        :return: shape(2, 3) Bounding box defined by the bottom South-West and
-            top North-East coordinates.
+        :return: Bounding box defined by the bottom South-West and
+            top North-East coordinates,  shape(2, 3).
         """
-        if self.vertices is not None:
-            return np.c_[self.vertices.min(axis=0), self.vertices.max(axis=0)].T
-
-        return None
+        return np.c_[self.vertices.min(axis=0), self.vertices.max(axis=0)].T
```
benk-mira, line 87: remove | None from typing

geoh5py/objects/points.py
```diff
@@ -47,60 +81,22 @@ def extent(self) -> np.ndarray | None:
...
         self, extent: np.ndarray, inverse: bool = False
     ) -> np.ndarray | None:
         """
         Sub-class extension of :func:`~geoh5py.shared.entity.Entity.mask_by_extent`.
         """
-        if self.extent is None or not box_intersect(self.extent, extent):
+        if not box_intersect(self.extent, extent):
             return None
 
         return mask_by_extent(self.vertices, extent, inverse=inverse)
```
benk-mira, line 98: instead of returning None, can we just return a mask with all ones?

geoh5py/objects/surface.py
```diff
@@ -34,45 +34,53 @@ class Surface(CellObject):
...
     )
 
-    def __init__(self, object_type: ObjectType, **kwargs):
-        self._cells: np.ndarray | None = None
+    @classmethod
+    def default_type_uid(cls) -> uuid.UUID:
+        return cls.__TYPE_UID
 
-        super().__init__(object_type, **kwargs)
+    def validate_cells(self, indices: list | tuple | np.ndarray | None):
```
benk-mira, line 41:
```suggestion
    def validate_cells(self, indices: list | tuple | np.ndarray | None) -> np.ndarray:
```

benk-mira (changes requested):
Posting a review up to the surveys.  Will carry on in a bit.

geoh5py/shared/entity.py
```diff
@@ -220,24 +221,33 @@ def metadata(self) -> dict | None:
...
-            self._metadata = None
-        else:
+    def metadata(self, value: dict | np.ndarray | bytes | None):
+        self._metadata = self.validate_metadata(value)
+
+        if self.on_file:
+            self.workspace.update_attribute(self, "metadata")
+
+    @staticmethod
+    def validate_metadata(value):
```
benk-mira, line 237:
```suggestion
    def validate_metadata(value) -> dict:
```

geoh5py/shared/entity_container.py
```diff
@@ -42,27 +42,42 @@ class EntityContainer(Entity):
...
         self._uid = (
             str2uuid(uid) if isinstance(str2uuid(uid), uuid.UUID) else uuid.uuid4()
         )
         self._children: list = []
 
-        super().__init__(uid, name, **kwargs)
+        super().__init__(uid, **kwargs)
 
-    def add_file(self, file: str):
+    def add_file(self, file: str | Path | bytes, name: str = "filename.dat"):
```
benk-mira, line 53: return type

geoh5py/shared/utils.py
```diff
@@ -624,24 +704,64 @@ def dip_points(points: np.ndarray, dip: float, rotation: float = 0) -> np.ndarra
     return points.T
 
 
-def map_attributes(object_, **kwargs):
+def set_attributes(entity, **kwargs):
+    """
+    Loop over kwargs and set attributes to an entity.
+
+    TODO: Deprecate in favor of explicit attribute setting.
```
benk-mira, line 711: Maybe you can knock this off if the constructor refactoring I suggested above works out?

geoh5py/shared/utils.py
```diff
@@ -624,24 +704,64 @@ def dip_points(points: np.ndarray, dip: float, rotation: float = 0) -> np.ndarra
...
+    """
+    for key, value in kwargs.items():
+        try:
+            setattr(entity, key, value)
+        except AttributeError:
+            continue
+
+
+def map_name_attributes(object_, **kwargs: dict):
```
benk-mira, line 720:
```suggestion
def map_name_attributes(object_, **kwargs: dict) -> dict:
```

benk-mira (changes requested):
From surveys on..

geoh5py/objects/block_model.py
```diff
@@ -43,21 +51,42 @@ class BlockModel(GridObject):
...
-        self._origin: np.ndarray = np.zeros(3)
-        self._rotation: float = 0.0
-        self._u_cell_delimiters: np.ndarray | None = None
-        self._v_cell_delimiters: np.ndarray | None = None
-        self._z_cell_delimiters: np.ndarray | None = None
-
-        super().__init__(object_type, **kwargs)
+    def __init__(
+        self,
+        object_type: ObjectType,
```
MatthieuCMira, line 56: Just need an explanation for my understanding:

why object_type?
I understand Workspace, or parent (which disappear)
But object_type is strange. 

I guess, as user, we can pass a workspace or a parent, and it creates an Object type?

geoh5py/objects/object_base.py
MatthieuCMira, line 1: Maybe some update of this class regarding other comments might be usefull.

add some validaotrs for object creations for example?

geoh5py/objects/surface.py
MatthieuCMira, line 41: Add docstring to this function

General comments: Do we want to keep the validators as public function, or pass them as private?

geoh5py/objects/cell_object.py
```diff
@@ -228,3 +239,9 @@ def copy(  # pylint: disable=too-many-branches
                 )
 
         return new_object
+
+    @abstractmethod
+    def validate_cells(self, indices: tuple | list | np.ndarray | None) -> np.ndarray:
```
MatthieuCMira, line 244: Extend the docstring a bit ;-)_

geoh5py/objects/surface.py
```diff
@@ -34,45 +34,53 @@ class Surface(CellObject):
...
-        return cls.__TYPE_UID
+    def validate_vertices(cls, xyz: np.ndarray | list | tuple) -> np.ndarray:
+        """
+        Validate and format type of vertices array.
+
+        :param xyz: Array of vertices as defined by :obj:`~geoh5py.objects.points.Points.vertices`.
+        """
+        if xyz is None:
+            warnings.warn(
+                "No 'vertices' provided. Using a default point at the origin.",
```
MatthieuCMira, line 73: the point 0, 0, 0 as origin

geoh5py/objects/curve.py
```diff
@@ -30,6 +30,15 @@ class Curve(CellObject):
     """
     Curve object defined by a series of line segments (:obj:`~geoh5py.objects.curve.Curve.cells`)
     connecting :obj:`~geoh5py.objects.object_base.ObjectBase.vertices`.
+
+    Attributes
+    ----------
+    :attr cells: Array of integer shape(*, 2) defining the connection between pair of vertices.
```
MatthieuCMira, line 36: params?

geoh5py/objects/curve.py
```diff
@@ -43,65 +43,53 @@ class Curve(CellObject):
...
+        self,
+        object_type: ObjectType,
+        vertices: np.ndarray = ((0.0, 0.0, 0.0), (1.0, 1.0, 1.0)),
+        cells: np.ndarray | None = None,
+        current_line_id: uuid.UUID | None = None,
+        parts: np.ndarray | None = None,
+        name="Curve",
+        **kwargs,
+    ):
         self._current_line_id: uuid.UUID | None = None
```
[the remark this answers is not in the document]
MatthieuCMira: And again, uyop can use the setters in the inint to validate the values

geoh5py/objects/curve.py
```diff
@@ -43,65 +52,37 @@ class Curve(CellObject):
...
+            name=name,
+            **kwargs,
+        )
 
     @property
     def current_line_id(self) -> uuid.UUID | None:
-        if getattr(self, "_current_line_id", None) is None:
-            self._current_line_id = uuid.uuid4()
-
+        """
```
MatthieuCMira, line 83: That's redundant with the typing

geoh5py/objects/points.py
```diff
@@ -128,35 +119,54 @@ def remove_vertices(
             raise ValueError("Found indices larger than the number of vertices.")
 
         vertices = np.delete(self.vertices, indices, axis=0)
```
MatthieuCMira, line 121: you can remove the isinstance self.vertices

geoh5py/objects/drillhole.py
```diff
@@ -52,7 +54,9 @@ class Drillhole(Points):
         }
     )
 
-    def __init__(self, object_type: ObjectType, **kwargs):
+    def __init__(
```
MatthieuCMira, line 57: Some docstring for drillholes parameters

geoh5py/objects/drillhole.py
```diff
@@ -67,6 +71,11 @@ def __init__(self, object_type: ObjectType, **kwargs):
 
         super().__init__(object_type, **kwargs)
 
+        if vertices is None:
+            self._vertices = None
+        else:
+            self._vertices: np.ndarray = self.validate_vertices(vertices)
+
     @classmethod
```
MatthieuCMira, line 79: this is redundant with points that can be removed.
I would even suggest putting the default_type_uid mechanism in Object_Base, and return a trivial uid in this class, instead of repeating this function

geoh5py/objects/points.py
```diff
@@ -32,12 +32,46 @@ class Points(ObjectBase):
     """
 
     __TYPE_UID = uuid.UUID("{202C5DB1-A56D-4004-9CAD-BAAFD8899406}")
+    __VERTICES_DTYPE = np.dtype([("x", "<f8"), ("y", "<f8"), ("z", "<f8")])
 
-    def __init__(self, object_type: ObjectType, name="Points", **kwargs):
-        self._vertices: np.ndarray | None = None
+    def __init__(
```
MatthieuCMira, line 37: The extent function can be set in points for all the inherited objects

geoh5py/objects/drillhole.py
MatthieuCMira, line 1: I feel like a lot of functions could be in points (vertices, add_vertices, extent)

Maybe we could enjoy this "spring cleanup" to format typing and docstring in other functions?

geoh5py/objects/drillhole.py
```diff
@@ -719,11 +747,13 @@ def sort_depths(self):
...
+                    self.workspace.update_attribute(self, "vertices")
 
                 if self.cells is not None:
                     key_map = np.argsort(sort_ind)[self.cells.flatten()]
-                    self.cells = key_map.reshape((-1, 2)).astype("uint32")
+                    self._cells = key_map.reshape((-1, 2)).astype("uint32")
+                    self.workspace.update_attribute(self, "cells")
 
 
 def compute_deviation(surveys: np.ndarray) -> np.ndarray:
```
MatthieuCMira, line 759: why not create staticmethods?

Are those function used anywhere else?

If so, I t would be better to create a drillhole_utils file in share.

geoh5py/objects/drape_model.py
```diff
@@ -21,33 +21,54 @@
...
+        object_type: ObjectType,
+        layers: np.ndarray | list | tuple | None = None,
+        prisms: np.ndarray | list | tuple | None = None,
+        **kwargs,
+    ):
+        self._centroids: np.ndarray | None = None
+        self._layers: np.ndarray = self.validate_layers(layers)
+        self._prisms: np.ndarray = self.validate_prisms(prisms)
+
+        super().__init__(object_type, layers=layers, prisms=prisms, **kwargs)
```
MatthieuCMira, line 62: Ben missed this one ;-) 

ObjectBase doens't have layers nor prisms

geoh5py/objects/drape_model.py
```diff
@@ -21,33 +21,54 @@
...
+    )
+
+    def __init__(
+        self,
+        object_type: ObjectType,
+        layers: np.ndarray | list | tuple | None = None,
+        prisms: np.ndarray | list | tuple | None = None,
+        **kwargs,
+    ):
+        self._centroids: np.ndarray | None = None
```
MatthieuCMira, line 58: why not defining centroids after layers and prisms are set then? 
It will save a None

geoh5py/objects/drape_model.py
```diff
@@ -88,11 +99,21 @@ def centroids(self):
         return self._centroids
 
     @property
-    def layers(self) -> np.ndarray | None:
+    def extent(self) -> np.ndarray:
```
MatthieuCMira, line 102: I feel like extent is redundant.
centroids are vertices likes object.

and it will always be np.c_[self.centroids.min(axis=0), self.centroids.max(axis=0)].T 
or the same with vertices.

MatthieuCMira (changes requested):
PARTIAL REVIEW.

I review changes in objects. 
I did not dive into the mechanisms. so it's still quite superficial. 

I did not review the changes concerning the groups or the test yet

The test is going to be quite a challenge there ;-)

geoh5py/objects/curve.py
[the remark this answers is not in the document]
domfournier: Yip. Done

geoh5py/groups/uijson.py
```diff
@@ -16,48 +16,70 @@
...
+class UIJsonGroup(Group):
+    """Group for SimPEG inversions."""
 
-    _name = "SimPEG"
-    _description = "SimPEG"
-    _options = None
+    __TYPE_UID = uuid.UUID("{BB50AC61-A657-4926-9C82-067658E246A0}")
+    _default_name = "UIJson"
 
     def __init__(self, group_type: GroupType, **kwargs):
```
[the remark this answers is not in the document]
domfournier: yes

geoh5py/groups/uijson.py
```diff
@@ -16,48 +16,70 @@
...
+            if value is not None:
+                self.options = value
 
-        return self._options
+        return self._options or {}
 
     @options.setter
-    def options(self, value: dict | None):
+    def options(self, value: dict | np.ndarray | bytes | None):
+        if isinstance(value, np.ndarray):
```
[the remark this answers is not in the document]
domfournier: Coming out of geoh5 it is stored as a 1D array

geoh5py/objects/block_model.py
```diff
@@ -43,21 +51,42 @@ class BlockModel(GridObject):
...
+        u_cell_delimiters: np.ndarray = np.array([0.0, 1.0]),
+        v_cell_delimiters: np.ndarray = np.array([0.0, 1.0]),
+        z_cell_delimiters: np.ndarray = np.array([0.0, 1.0]),
+        **kwargs,
+    ):
+        self._u_cell_delimiters: np.ndarray = self.validate_cell_delimiters(
```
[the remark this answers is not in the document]
domfournier: True

geoh5py/objects/cell_object.py
```diff
@@ -40,35 +37,49 @@ class CellObject(Points, ABC):
...
-            cell_mask = np.all(vert_mask[self.cells], axis=1)
-            orphan_mask = np.zeros_like(vert_mask, dtype=bool)
-            orphan_mask[self.cells[cell_mask].flatten()] = True
-            vert_mask &= orphan_mask
+        cell_mask = np.all(vert_mask[self.cells], axis=1)
+        orphan_mask = np.zeros_like(vert_mask, dtype=bool)
+        orphan_mask[self.cells[cell_mask].flatten()] = True
+        vert_mask &= orphan_mask
 
         if ~np.any(vert_mask):
```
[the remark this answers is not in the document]
domfournier: To be tackled on GEOPY-1626

geoh5py/objects/grid2d.py
```diff
@@ -250,65 +245,16 @@ def dip(self, value):
...
+        Number of cells along the u and v-axis.
         """
-        if self.u_count is not None and self.v_count is not None:
-            return self.u_count, self.v_count
-        return None
+        return self.u_count, self.v_count
 
     @property
-    def u_cell_size(self) -> float | None:
+    def u_cell_size(self) -> float:
```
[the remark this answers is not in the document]
domfournier: ok

geoh5py/objects/drillhole.py
domfournier: To be addressed on a dedicated issue GEOPY-1627

geoh5py/objects/drillhole.py
domfournier: Sure, beyond scope of this PR. We can tackle on GEOPY-1627

geoh5py/objects/drape_model.py
domfournier: Done

geoh5py/objects/grid_object.py
```diff
@@ -142,10 +85,79 @@ def mask_by_extent(
...
+    @property
+    def rotation(self) -> float:
+        """
+        Clockwise rotation angle (degree) about the vertical axis.
+        """
+        return self._rotation
+
+    @rotation.setter
+    def rotation(self, value: np.ndarray | Number):
+        if isinstance(value, Number):
```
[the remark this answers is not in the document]
domfournier: Because it stored on geoh5 as an array

geoh5py/objects/grid_object.py
```diff
@@ -19,120 +19,63 @@
...
-    def __init__(self, object_type: ObjectType, **kwargs):
+    def __init__(
+        self,
+        object_type: ObjectType,
+        origin: np.ndarray | tuple = (0.0, 0.0, 0.0),
+        rotation: float = 0.0,
+        **kwargs,
+    ):
         self._centroids: np.ndarray | None = None
+        self._origin: np.ndarray = origin
```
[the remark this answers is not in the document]
domfournier: Done

geoh5py/objects/drape_model.py
domfournier: Does it matter? Centroids is just a stashed property to get cell centers

geoh5py/objects/drape_model.py
domfournier: The plan is to have faster lookups for the extent (or store). Beyond scope.

geoh5py/objects/grid2d.py
```diff
@@ -241,6 +235,7 @@ def dip(self) -> float:
     def dip(self, value):
```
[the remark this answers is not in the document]
domfournier: Done

geoh5py/objects/octree.py
```diff
@@ -52,32 +62,47 @@ class Octree(GridObject):
...
+        u_cell_size: float = 1.0,
+        v_cell_size: float = 1.0,
+        w_cell_size: float = 1.0,
+        octree_cells: np.ndarray | list | tuple | None = None,
+        **kwargs,
+    ):
+        self._u_count = self.validate_octree_count(u_count, "u")
+        self._v_count = self.validate_octree_count(v_count, "v")
+        self._w_count = self.validate_octree_count(w_count, "w")
+        self._u_cell_size: float
```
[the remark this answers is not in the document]
domfournier: Done

geoh5py/objects/octree.py
```diff
@@ -189,211 +196,121 @@ def octree_cells(self) -> np.ndarray | None:
...
-        :obj:`int`: Number of cells along v-axis.
+        Number of cells along v-axis.
         """
         return self._v_count
 
-    @v_count.setter
-    def v_count(self, value: int):
-        value = np.int32(value).item()
+    @staticmethod
+    def validate_octree_count(value: int, axis: str) -> np.int32:
```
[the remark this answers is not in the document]
domfournier: Done

geoh5py/objects/octree.py
```diff
@@ -189,211 +196,121 @@ def octree_cells(self) -> np.ndarray | None:
...
         """
-        :obj:`float`: Base cell size along the w-axis.
+        Base cell size along the w-axis.
         """
         return self._w_cell_size
 
     @w_cell_size.setter
-    def w_cell_size(self, value: float | np.ndarray):
-        if not isinstance(value, (float, np.ndarray)):
+    def w_cell_size(self, value: float):
```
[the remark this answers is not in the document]
domfournier: Done. Accept all real

geoh5py/objects/grid_object.py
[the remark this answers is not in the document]
MatthieuCMira: type(2.0) == type(np.r_[2].astype(float).item())

You are converting the array back to float when assigning to self._rotation.
I'm ok if it can take a numpy as input, it just seems overly complex for a float setter, even if the setter can accept an array of shape 1.

You can convert directly a 1d array with float(np.array([2]))

geoh5py/objects/drape_model.py
MatthieuCMira: just to simplify. 
Technicaly, centroids cannot be None as the object cannot be empty?
and as the object shape can't change, we can set them oncee and for all? 
So it's removing the None mechanic?

geoh5py/objects/block_model.py
domfournier: It's just the pattern that was originally used for all entities (groups, objects, data). First, the type is created and registered to the workspace, which validates that it is unique. For example data can share their types or create new ones, while objects and groups have fixed ones. Then the type is given to the entity as an attribute (already validated)

geoh5py/objects/surface.py
domfournier: I think it could be useful down the road to ask the object to validate an array ahead of time. Not sure making it private does anything more.,

geoh5py/objects/drillhole.py
```diff
@@ -52,7 +54,9 @@ class Drillhole(Points):
         }
     )
 
-    def __init__(self, object_type: ObjectType, **kwargs):
+    def __init__(
+        self, object_type: ObjectType, vertices: np.ndarray | None = None, **kwargs
+    ):
         self._cells: np.ndarray | None = None
         self._collar: np.ndarray | None = None
```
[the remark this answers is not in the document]
domfournier: Done

geoh5py/objects/grid2d.py
```diff
@@ -58,52 +58,58 @@ class Grid2D(GridObject):
...
+        object_type: ObjectType,
+        u_cell_size: float = 1.0,
+        v_cell_size: float = 1.0,
+        u_count: int = 1,
+        v_count: int = 1,
+        vertical: bool = False,
+        dip: float = 0.0,
+        **kwargs,
+    ):
+        self._u_cell_size: float
```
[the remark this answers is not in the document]
domfournier: Done

geoh5py/shared/utils.py
domfournier: Can't do until we refactor the Data classes

geoh5py/objects/curve.py
domfournier: Yes

geoh5py/objects/curve.py
domfournier: Yes

geoh5py/objects/drillhole.py
domfournier: Done

geoh5py/objects/drape_model.py
domfournier: True for DrapeModel yes. For other objects centroids can change if the object origin/rotation changes. I ll just keep it consistent with the other grid objects.

geoh5py/objects/points.py
domfournier: Yep, done.

geoh5py/objects/points.py
domfournier: True

benk-mira (approved):
Thanks for all the work in here, was a big one!

geoh5py/objects/curve.py
```diff
@@ -115,30 +83,29 @@ def current_line_id(self, value: uuid.UUID | None):
...
         property get modified by the setting of parts.
         """
-        if (
-            getattr(self, "_parts", None) is None
-            and self.cells is not None
-            and self.vertices is not None
-        ):
+        if getattr(self, "_parts", None) is None:
             cells = self.cells
             parts = np.zeros(self.vertices.shape[0], dtype="int")
```
MatthieuCMira, line 110:
```suggestion
            parts = np.hstack(([0], np.cumsum((cells[1:, 0] != cells[:-1, 1]).astype(int))))
```

geoh5py/objects/points.py
```diff
@@ -110,53 +102,64 @@ def remove_vertices(
...
-    def copy(
-        self,
-        parent=None,
-        copy_children: bool = True,
-        clear_cache: bool = False,
-        mask: np.ndarray | None = None,
-        **kwargs,
-    ):
+    @property
+    def vertices(self) -> np.ndarray:
```
MatthieuCMira, line 120: what about a n_vertices property too?
like n_cells in grid objects

geoh5py/objects/curve.py
```diff
@@ -115,30 +83,29 @@ def current_line_id(self, value: uuid.UUID | None):
...
+        if self.on_file:
+            self.workspace.update_attribute(self, "attributes")
+
+    def _make_cells_from_parts(self) -> np.ndarray:
         """
-        :return: Default unique identifier
+        Generate cells from parts.
         """
-        return cls.__TYPE_UID
+        cells = []
```
MatthieuCMira, line 94:
```suggestion

cells = np.c_[ np.arange(self.n_vertices-1),  np.arange(1, self.n_vertices)]
indices_to_suppress = np.flatnonzero(np.diff(self.parts)) + 1
cells = np.delete(cells, indices_to_suppress, axis=0)
return cells

Comment (MatthieuCMira) [reply geoh5py/objects/curve.py:94]: removing loops ;-)

File: geoh5py/objects/surface.py [hunk 58]
```diff
@@ -18,61 +18,74 @@
...
+                "Attribute 'cells' must be provided as type numpy.ndarray, list or tuple."
             )
 
-        if indices.ndim != 2 or indices.shape[1] != 3:
-            raise ValueError("Array of cells should be of shape (*, 3).")
+        if indices.ndim != 2 or indices.shape[-1] != 3:
+            raise ValueError("Array of 'cells' should be of shape (*, 3).")
 
         if not np.issubdtype(indices.dtype, np.integer):
             raise TypeError("Indices array must be of integer type")
```
MatthieuCMira, line 66: Same question as curve:

Is there more validation to do? Like a node can only be call 3 times, there can't be orphans nodes etc.

geoh5py/objects/surface.py
```diff
@@ -18,61 +18,74 @@
...
+            xyz = (0.0, 0.0, 0.0)
+            warnings.warn(
+                f"No 'vertices' provided. Using a default point {xyz} at the origin.",
+                UserWarning,
+            )
+
+        if isinstance(xyz, (list, tuple)):
+            xyz = np.array(xyz, ndmin=2)
+
+        if len(xyz) < 3:
```
MatthieuCMira, line 86: and if len(xyz) is 2?

geoh5py/objects/drape_model.py
```diff
@@ -21,33 +21,50 @@
...
 
 
-class DrapeModel(GridObject):
+class DrapeModel(ObjectBase):
     """
     Drape (curtain) model object made up of layers and prisms.
-    """
 
-    __TYPE_UID = uuid.UUID("{C94968EA-CF7D-11EB-B8BC-0242AC130003}")
+    :attr layers: Array of layers in the drape model organized into blocks
```
MatthieuCMira, line 32: use :params: to be consitent

geoh5py/objects/drape_model.py
MatthieuCMira: As you wish, I'm ok with that

geoh5py/objects/grid2d.py
```diff
@@ -58,52 +59,55 @@ class Grid2D(GridObject):
     )
     _converter: type[Grid2DConversion] = Grid2DConversion
 
-    def __init__(self, object_type: ObjectType, **kwargs):
-        self._origin: np.ndarray = np.asarray(
-            tuple(np.zeros(3)), dtype=[("x", float), ("y", float), ("z", float)]
+    def __init__(  # pylint: disable=too-many-arguments
```
MatthieuCMira, line 62: add the :params: in the docstring

geoh5py/objects/grid2d.py
```diff
@@ -58,52 +59,55 @@ class Grid2D(GridObject):
...
 
-        super().__init__(object_type, **kwargs)
+        self.u_cell_size = u_cell_size
+        self.v_cell_size = v_cell_size
+        self.dip = dip
+        self.vertical = vertical
 
     @property
-    def cell_center_u(self) -> np.ndarray | None:
+    def cell_center_u(self) -> np.ndarray:
```
MatthieuCMira, line 87: I agree with you, we could put cell_center_uv as deprecated, it's not very usefull as it's not considering origin nor rotation and dip

geoh5py/objects/grid2d.py
```diff
@@ -250,151 +236,93 @@ def dip(self, value):
...
-    def u_cell_size(self, value: float | np.ndarray):
-        if not isinstance(value, (float, np.ndarray)):
+    def u_cell_size(self, value: Real | np.ndarray):
+        if not isinstance(value, (Real, np.ndarray)):
             raise TypeError("Attribute 'u_cell_size' must be type(float).")
 
         self._centroids = None
+
         if isinstance(value, np.ndarray):
             assert len(value) == 1, "u_cell_size must be a float of shape (1,)"
```
MatthieuCMira, line 260: error rather than assert

geoh5py/objects/grid2d.py
```diff
@@ -250,151 +236,93 @@ def dip(self, value):
...
             self._u_cell_size = np.r_[value].astype(float).item()
         else:
-            self._u_cell_size = value
-        self.workspace.update_attribute(self, "attributes")
+            self._u_cell_size = float(value)
+
```
MatthieuCMira, line 264: so ucell size/vcell size can be negative?

We should have a validator to ensure positive

geoh5py/objects/grid2d.py
```diff
@@ -250,151 +236,93 @@ def dip(self, value):
...
     @v_cell_size.setter
-    def v_cell_size(self, value: float | np.ndarray):
-        if not isinstance(value, (float, np.ndarray)):
+    def v_cell_size(self, value: Real | np.ndarray):
+        if not isinstance(value, (Real, np.ndarray)):
             raise TypeError("Attribute 'v_cell_size' must be type(float).")
 
         self._centroids = None
         if isinstance(value, np.ndarray):
             assert len(value) == 1, "v_cell_size must be a float of shape (1,)"
```
MatthieuCMira, line 289: error instead of assert

geoh5py/objects/grid2d.py
```diff
@@ -410,3 +338,13 @@ def to_geoimage(
         :return: a new georeferenced :obj:`geoh5py.objects.geo_image.GeoImage`.
         """
         return self.converter.to_geoimage(self, keys, mode=mode, **geoimage_kwargs)
+
+    @staticmethod
+    def validate_count(value: int, axis: str) -> np.int32:
+        """
+        Validate and format type of count value.
+        """
+        if not isinstance(value, (np.integer, int)):
```
MatthieuCMira, line 347: we should validate if count is positive too

geoh5py/objects/block_model.py
```diff
@@ -67,210 +90,104 @@ def centroids(self) -> np.ndarray | None:
...
-            cell_center_u = np.cumsum(self.u_cells) - self.u_cells / 2.0
-            cell_center_v = np.cumsum(self.v_cells) - self.v_cells / 2.0
-            cell_center_z = np.cumsum(self.z_cells) - self.z_cells / 2.0
-
+        if getattr(self, "_centroids", None) is None:
+            cell_center_u = self.local_axis_centers("u")
+            cell_center_v = self.local_axis_centers("v")
+            cell_center_z = self.local_axis_centers("z")
             angle = np.deg2rad(self.rotation)
             rot = np.r_[
```
MatthieuCMira, line 98: use xy_rotation_matrix() function instead for clarity

geoh5py/objects/block_model.py
```diff
@@ -67,210 +90,104 @@ def centroids(self) -> np.ndarray | None:
...
-        shape (:obj:`~geoh5py.objects.block_model.BlockModel.shape` [2], ):
         Cell size along the z-axis
         """
-        if self.z_cell_delimiters is not None:
-            return self.z_cell_delimiters[1:] - self.z_cell_delimiters[:-1]
-        return None
+        return self.z_cell_delimiters[1:] - self.z_cell_delimiters[:-1]
+
+    @staticmethod
+    def validate_cell_delimiters(value: np.ndarray, axis: str) -> np.ndarray:
```
MatthieuCMira, line 175: is it enought in terms of validation?
Cells shouldn't be equals in x, y, z for example?

geoh5py/objects/octree.py
```diff
@@ -32,12 +33,22 @@ class Octree(GridObject):
     """
     Octree mesh class that uses a tree structure such that cells
     can be subdivided it into eight octants.
+
+    :attr u_count: Number of cells along the u-axis.
```
MatthieuCMira, line 37: use :params: for consitency

geoh5py/objects/octree.py
```diff
@@ -128,18 +142,12 @@ def centroids(self):
             ]
         """
         if getattr(self, "_centroids", None) is None:
-            assert self.octree_cells is not None, "octree_cells must be set"
-            assert self.u_cell_size is not None, "u_cell_size must be set"
-            assert self.v_cell_size is not None, "v_cell_size must be set"
-            assert self.w_cell_size is not None, "w_cell_size must be set"
-
             angle = np.deg2rad(self.rotation)
             rot = np.r_[
```
MatthieuCMira, line 146: use rot_xy function instead

geoh5py/objects/octree.py
```diff
@@ -189,212 +187,121 @@ def octree_cells(self) -> np.ndarray | None:
...
         """
-        :obj:`float`: Base cell size along the u-axis.
+        Base cell size along the u-axis.
         """
         return self._u_cell_size
 
     @u_cell_size.setter
-    def u_cell_size(self, value: float | np.ndarray):
-        if not isinstance(value, (float, np.ndarray)):
+    def u_cell_size(self, value: Real):
```
MatthieuCMira, line 212: cell_size must be bositive too

MatthieuCMira (changes requested):
I went a bit more into details there. 

General comment: in the validators, it seems it's not blocking all the bad usages. 
We can't be exhausive of course, but it would be nice to cover most obvious scenarios.
I commented most of the time, but of course, I might have miss one...

NOTA: I just did a review of the objects, the review of test and others will come latter.

geoh5py/objects/object_base.py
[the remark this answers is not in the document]
domfournier: Done

geoh5py/objects/object_base.py
```diff
@@ -336,10 +329,8 @@ def entity_type(self) -> ObjectType:
     @abstractmethod
```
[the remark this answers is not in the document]
domfournier: Good idea, we actually now have `locations` which does the same.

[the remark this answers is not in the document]
domfournier: Actually, harder to generalize. Would only simplify points and grids. I ll skip this one.

geoh5py/objects/points.py
```diff
@@ -31,76 +31,68 @@ class Points(ObjectBase):
...
+            clear_cache=clear_cache,
+            mask=mask,
+            **kwargs,
+        )
+
+        return new_entity
 
     @property
-    def extent(self) -> np.ndarray | None:
+    def extent(self) -> np.ndarray:
```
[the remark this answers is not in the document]
domfournier: Same here, it's a bit more complicated because not all objects have an extent, but all Points and Grids sub-classes do. I would rather keep those classes always returning an array rather than always checking for None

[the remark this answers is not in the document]
domfournier: Ok I managed to move some of it.

geoh5py/objects/octree.py
domfournier: Doesn not have to be no.

geoh5py/objects/block_model.py
domfournier: Nope, three independent vectors

geoh5py/objects/points.py
```diff
@@ -110,53 +102,64 @@ def remove_vertices(
...
 
         if not isinstance(indices, np.ndarray):
             raise TypeError("Indices must be a list or numpy array.")
 
-        if (
-            isinstance(self.vertices, np.ndarray)
-            and np.max(indices) > self.vertices.shape[0] - 1
-        ):
+        if np.max(indices) > self.vertices.shape[0] - 1:
             raise ValueError("Found indices larger than the number of vertices.")
```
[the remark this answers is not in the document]
domfournier: Good idea for the minimum_vertices. The validation method can handle it.

geoh5py/objects/cell_object.py
```diff
@@ -134,15 +136,17 @@ def remove_vertices(
         vert_index = np.ones(self.vertices.shape[0], dtype=bool)
```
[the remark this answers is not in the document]
domfournier: the mask is negative right now (indices taken out). So the bool array needs to take the complement.

geoh5py/objects/cell_object.py
```diff
@@ -100,10 +107,10 @@ def remove_cells(self, indices: list[int] | np.ndarray, clear_cache: bool = Fals
...
 
         cells = np.delete(self.cells, indices, axis=0)
-        self._cells = None
-        setattr(self, "cells", cells)
 
+        self._cells = self.validate_cells(cells)
         self.remove_children_values(indices, "CELL", clear_cache=clear_cache)
+        self.workspace.update_attribute(self, "cells")
 
     def remove_vertices(
```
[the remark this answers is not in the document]
domfournier: nice

geoh5py/objects/curve.py
```diff
@@ -152,28 +119,80 @@ def parts(self):
...
+            raise TypeError(
+                "Attribute 'cells' must be provided as type numpy.ndarray, list or tuple."
+            )
+
+        if indices.ndim != 2 or indices.shape[1] != 2:
+            raise ValueError("Array of cells should be of shape (*, 2).")
+
+        if not np.issubdtype(indices.dtype, np.integer):
+            raise TypeError("Indices array must be of integer type")
+
```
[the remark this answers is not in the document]
domfournier: Nope, good idea.

geoh5py/objects/curve.py
```diff
@@ -152,28 +119,80 @@ def parts(self):
...
 
-            assert indices.shape == (
-                self.vertices.shape[0],
-            ), f"Provided parts must be of shape {self.vertices.shape[0]}"
-            self._parts = indices
-            self._cells = None
-            self.workspace.update_attribute(self, "cells")
+        indices = indices.astype("int32")
+
+        if vertices is not None and len(indices) != vertices.shape[0]:
```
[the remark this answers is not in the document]
domfournier: Yes it can, although GA struggles with visualizing it.

geoh5py/objects/curve.py
domfournier: Love it

domfournier: Had to tweak it a bit, but works

geoh5py/objects/curve.py
domfournier: vertices might not be ordered, so we need to fetch each part sequentially.

geoh5py/objects/surface.py
domfournier: Already done on the base class

geoh5py/objects/grid2d.py
domfournier: Can be both positive and negative.

geoh5py/objects/grid2d.py
domfournier: Done

geoh5py/objects/grid2d.py
domfournier: Greater than 1, yes

geoh5py/objects/points.py
domfournier: Yep, already there.

geoh5py/objects/block_model.py
domfournier: nice

geoh5py/objects/octree.py
domfournier: Done

geoh5py/shared/entity.py
```diff
@@ -50,27 +67,41 @@ class Entity(ABC):
...
+        self.allow_move = allow_move
+        self.allow_rename = allow_rename
+        self.clipping_ids = clipping_ids
+        self.metadata = metadata
+        self.parent = parent or self.workspace.root
+        self.partially_hidden = partially_hidden
+        self.public = public
+        self.visible = visible
+
+        # TODO Deprecate in favor of explicit attribute setter
```
MatthieuCMira, line 103: do we still need the map_attributes then?

geoh5py/shared/utils.py
```diff
@@ -552,7 +632,7 @@ def mask_by_extent(
     return indices
 
 
-def get_attributes(entity, omit_list=(), attributes=None):
+def get_attributes(entity, omit_list=(), attributes=None) -> dict:
```
MatthieuCMira, line 635: typing Any in this case I guess?

MatthieuCMira (commented):
Great, it's definitely improved considerably objects.

Last review for me, some few last comments, but nothing major!
(I did not review the tests really, I guess you mostly change syntaxes)


We could create an issue for a last refactoring of object. Like ensuring the typing, error, functions are sorted alphabetically, and the good practices we put in place recently in general!

geoh5py/objects/points.py
```diff
@@ -22,84 +22,63 @@
...
-        Geography bounding box of the object.
-
-        :return: shape(2, 3) Bounding box defined by the bottom South-West and
-            top North-East coordinates.
+        Sub-class extension of :func:`~geoh5py.shared.entity.Entity.copy`.
         """
-        if self.vertices is not None:
-            return np.c_[self.vertices.min(axis=0), self.vertices.max(axis=0)].T
+        if mask is not None and self.vertices is not None:
+            if not isinstance(mask, np.ndarray) or mask.shape != (self.n_vertices,):
```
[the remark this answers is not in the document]
domfournier: true

geoh5py/objects/drillhole.py
```diff
@@ -90,52 +110,60 @@ def cells(self, indices):
...
-            )
-            self._collar = value
-            self.workspace.update_attribute(self, "attributes")
+        if len(value) != 3:
+            raise ValueError("Origin must be a list or numpy array of len (3,).")
+
+        value = np.asarray(tuple(value), dtype=self.__COLLAR_DTYPE)
+        self._collar = value
 
         self._locations = None
```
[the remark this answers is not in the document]
domfournier: Trace holds absolute coordinates of the well path, computed by GA, based on the collar and survey information. So if either changes, we need to let GA re-compute.

geoh5py/objects/drillhole.py
```diff
@@ -145,12 +173,14 @@ def end_of_hole(self) -> float | None:
         return self._end_of_hole
 
     @end_of_hole.setter
-    def end_of_hole(self, value: float | int | None):
+    def end_of_hole(self, value: Real | None):
         assert isinstance(
```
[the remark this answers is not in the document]
domfournier: To be tackled globally on GEOPY-1667

geoh5py/objects/surface.py
```diff
@@ -22,57 +22,46 @@
...
+            ].astype("uint32")
+
+        if not isinstance(indices, np.ndarray):
+            raise AttributeError(
+                "Attribute 'cells' must be provided as type numpy.ndarray, list or tuple."
             )
 
-        if indices.ndim != 2 or indices.shape[1] != 3:
-            raise ValueError("Array of cells should be of shape (*, 3).")
+        if indices.ndim != 2 or indices.shape[-1] != 3:
```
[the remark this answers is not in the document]
domfournier: Yea for sure, there are more advanced ways to create triangulation, but that's really on the user. This is just a way to have something for cells so it's never None. I don't think that's the responsibility of geoh5py

geoh5py/objects/octree.py
```diff
@@ -189,212 +178,121 @@ def octree_cells(self) -> np.ndarray | None:
...
-            value = np.r_[value]
-            assert len(value) == 1, "Rotation angle must be a float of shape (1,)"
-            self._centroids = None
-            self._rotation = value.astype(float).item()
-            self.workspace.update_attribute(self, "attributes")
+        return self._octree_cells
 
     @property
-    def shape(self) -> tuple | None:
+    def shape(self) -> tuple[np.int32, np.int32, np.int32]:
```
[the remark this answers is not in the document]
domfournier: That's the type enforced by geoh5.

geoh5py/objects/octree.py
```diff
@@ -189,212 +178,121 @@ def octree_cells(self) -> np.ndarray | None:
...
+            assert (
+                array.shape[1] == 4
+            ), "'octree_cells' requires an ndarray of shape (*, 4)"
+            array = np.asarray(
+                np.core.records.fromarrays(
+                    array.T.tolist(),
+                    dtype=cls.__OCTREE_DTYPE,
+                )
+            )
+        if array.dtype != cls.__OCTREE_DTYPE:
```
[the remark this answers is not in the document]
domfournier: I am not getting it. The counts are stored as int, `octree_cells` is an array defining the refinement and position of each cell...

geoh5py/data/data.py
```diff
@@ -154,13 +139,18 @@ def n_values(self) -> int | None:
...
-        if self.association is DataAssociationEnum.VERTEX:
-            return self.parent.n_vertices
-        if self.association is DataAssociationEnum.DEPTH:
+        if self.association in [
+            DataAssociationEnum.VERTEX,
+            DataAssociationEnum.DEPTH,
+        ] and hasattr(self.parent, "n_vertices"):
             return self.parent.n_vertices
-        if self.association is DataAssociationEnum.CELL:
+        if self.association is DataAssociationEnum.CELL and hasattr(
```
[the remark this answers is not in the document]
domfournier: From-to is a `CELL` association

geoh5py/data/data.py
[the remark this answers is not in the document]
domfournier: To be addressed on GEOPY-1668

geoh5py/shared/entity.py
domfournier: Will be removed once we tackle GEOPY-1668

geoh5py/objects/octree.py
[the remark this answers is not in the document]
MatthieuCMira: yes, but in a voxet, cells must be regulars arn't they?

MatthieuCMira (approved):
Great job! 

I created another issue (https://mirageoscience.atlassian.net/browse/GEOPY-1669) for all small syntaxes cleaning in the object classes. I'll take care of that ;-)

PR #583, opened 2024-07-22
GEOPY-1638: Convert mesh_type uuid to class type in ObjectForm

benk-mira:
**GEOPY-1638 - Convert mesh_type uuid to class type in ObjectForm**


geoh5py/shared/validators.py
```diff
@@ -41,6 +41,42 @@
 from geoh5py.shared.utils import iterable
 
 
+def to_list(value):
```
domfournier, line 44: Can you type those?

domfournier (commented):
Looks great. Can you just type the in/out of the new functions?

geoh5py/shared/validators.py
benk-mira: done.

domfournier (approved):
Looks great.

PR #584, opened 2024-07-22
GEOPY-1637: Allow default state (empty string) for all forms

benk-mira:
**GEOPY-1637 - Allow default state (empty string) for all forms**


geoh5py/ui_json/forms.py
```diff
@@ -101,7 +101,7 @@ class BaseForm(BaseModel):
     )
 
     label: str
-    value: Any
+    value: Any | None = None
```
domfournier, line 104: I don't None is valid for any value. I would revert

geoh5py/ui_json/forms.py
```diff
@@ -123,23 +123,23 @@ class StringForm(BaseForm):
     String valued uijson form.
     """
 
-    value: str = ""
+    value: str | None = ""
```
domfournier, line 126: Same here

domfournier (commented):
Let's chat

domfournier (approved):

PR #585, opened 2024-07-22
GEOPY-1674: Fail reading Geometric data

domfournier:
**GEOPY-1674 - Fail reading Geometric data**


geoh5py/data/data_type.py
```diff
@@ -153,43 +132,57 @@ def color_map(self, color_map: ColorMap | dict | np.ndarray | None):
...
+        :param workspace: An active Workspace class
+        :param uid: The unique identifier of the entity type.
+        :param entity_class: The class of the entity.
+        :param kwargs: The attributes of the entity type.
+
+        :return: EntityType
         """
-        Get the data type for y data.
+        primitive_type = kwargs.get("primitive_type", None)
+        if isinstance(primitive_type, str) and primitive_type.upper() == "GEOMETRIC":
```
MatthieuCMira, line 160: general question: why geometric type should be considered as a different datatype?

geoh5py/data/data_type.py
```diff
@@ -153,43 +132,57 @@ def color_map(self, color_map: ColorMap | dict | np.ndarray | None):
...
-    def for_z_data(cls, workspace: Workspace) -> DataType:
+    def _geometric_data(
+        cls, workspace: Workspace, uid: UUID | None, **kwargs
+    ) -> DataType:
         """
-        Get the data type for z data.
+        Get the data type for geometric data.
 
         :param workspace: An active Workspace.
+        :param uid: The uid of the existing data type to get.
```
MatthieuCMira, line 175: ad kwargs in the docstring too

geoh5py/data/data_type.py
```diff
@@ -153,43 +132,57 @@ def color_map(self, color_map: ColorMap | dict | np.ndarray | None):
...
         """
-        return cls._for_geometric_data(
-            workspace, GeometricDataConstants.z_datatype_uid()
+        module = __import__("geoh5py").data
+        geometric_class = getattr(
+            module.geometric_data,
+            f"GeometricDataConstants{kwargs.get('name', '')}",
+            None,
         )
+        return super().find_or_create(workspace, uid, geometric_class, **kwargs)
```
MatthieuCMira, line 185: this line is the same as  line 163. 

Wht don't we want to change entity_type by geometric class to pass to data by the same process. 

def find_or_create(...):

   if geometric_type:
       entity_class = geometric_class
   
   entity_type = super().find_ro_create
   
   return entity_type

geoh5py/data/data_type.py
```diff
@@ -153,43 +132,57 @@ def color_map(self, color_map: ColorMap | dict | np.ndarray | None):
...
+        Get the data type for geometric data.
 
         :param workspace: An active Workspace.
+        :param uid: The uid of the existing data type to get.
 
         :return: A new instance of DataType.
         """
-        return cls._for_geometric_data(
-            workspace, GeometricDataConstants.z_datatype_uid()
+        module = __import__("geoh5py").data
```
MatthieuCMira, line 179: These hacky lines are to avoid a circular import?

Is there a cleaner way to do that?

geoh5py/objects/geo_image.py
MatthieuCMira, line 1: Just a general comment for another issue:

creating multiband images with data is in the wishlist. 
Maybe creating an issue to deal with that, in the conversion method.
Could be an easy application in pro-geophysics.

MatthieuCMira (commented):
seems good overall. 

I am not so convinced in dealing with the circular import. 
It seems a level of complexity we don't want to have...

But if you find no other way, go for it.

geoh5py/data/data_type.py
domfournier: It's the only data type that has a fixed uuid

geoh5py/data/data_type.py
domfournier: I think so. Let's try again.

geoh5py/objects/geo_image.py
MatthieuCMira: more or less. 
Geologists want to create composite images.
for example, choose bands of data (like radiometrics) to create RGB, or CMYK

geoh5py/workspace/workspace.py
```diff
@@ -1182,13 +1182,18 @@ def load_entity(
...
+                base_classes[entity_type],
+                save_on_creation=False,
+                entity=entity_attrs,
+                entity_type=type_attrs,
+            )
+        except TypeError:
+            warnings.warn(
+                f"Could not create an entity from the given attributes {type_attrs}. Skipping over."
+            )
+            entity = None
```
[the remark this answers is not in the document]
MatthieuCMira: well, I was just thinking of 
```suggestion
            return None
```
very minor

geoh5py/objects/geo_image.py
domfournier: That's already handled by GA though.

geoh5py/data/geometric_data.py
```diff
@@ -19,31 +19,22 @@
...
-    __X_TYPE_UID = uuid.UUID(
-        fields=(0xE9E6B408, 0x4109, 0x4E42, 0xB6, 0xA8, 0x685C37A802EE)
-    )
-    __Y_TYPE_UID = uuid.UUID(
-        fields=(0xF55B07BD, 0xD8A0, 0x4DFF, 0xBA, 0xE5, 0xC975D490D71C)
-    )
-    __Z_TYPE_UID = uuid.UUID(
-        fields=(0xDBAFB885, 0x1531, 0x410C, 0xB1, 0x8E, 0x6AC9A40B4466)
-    )
+class GeometricDataConstants(Data):
```
MatthieuCMira, line 26: But then, where this class is used.
I only so its import in data __init__

MatthieuCMira (approved):
seems good to go for me :-)

I just wondering what is the usage of GeometricDataConstant now?

geoh5py/data/geometric_data.py
domfournier: The X, Y and Z are data of type `GeometricDataConstants`, of entity_type either GeometricDataX, Y or Z

geoh5py/objects/geo_image.py
MatthieuCMira: oh ok I did not knew

geoh5py/data/geometric_data.py
MatthieuCMira: But now, GeometricDataX, Y, Z are not inherited from GeometricDataConstants...

MatthieuCMira (approved):
perfect, good to test

MatthieuCMira (approved):
going to test

PR #590, opened 2024-08-06
DEVOPS-439: Switch from poetry to hatch

SophieCurinier:
**DEVOPS-439 - Switching geoh5py from poetry to hatch**
- Update `pyproject.toml` with hatch configuration
- Delete version test, given hatch manage the version it's not necessary anymore
- Update github workflows with hatch instead of poetry


pyproject.toml
```diff
@@ -27,40 +34,146 @@ classifiers = [
...
+]
 
-exclude = ["geoh5py/handlers/*", "geoh5py/interfaces/*"]
+[project.urls]
+Documentation = "https://mirageoscience-geoh5py.readthedocs-hosted.com/"
+Homepage = "https://www.mirageoscience.com/mining-industry-software/python-integration/"
+Repository = "https://github.com/MiraGeoscience/geoh5py"
 
+[tool.hatch.version]
+path = "geoh5py/__init__.py"
```
sebhmg, line 62: **suggestion:** configure for auto-versioning from tags

This config here assign version from parsing `__version__` in __init__.py, right? Better than before, for sure!
Just one step further ;) can also be tackled in another issue.
Note that version is likely used again in doc. Is there a way to get it automated in the doc as well?

sebhmg (changes requested):
very promising ! Happy to see it coming!

See a few questions, suggestions and fixes

pyproject.toml
```diff
@@ -1,21 +1,28 @@
...
+build-backend = "hatchling.build"
 
-license = "LGPL-3.0-or-later"
+[project]
+name = "geoh5py"
 description = "Python API for geoh5, an open file format for geoscientific data"
-authors = ["Mira Geoscience <support@mirageoscience.com>"]
-repository = "https://github.com/MiraGeoscience/geoh5py"
+readme = "package.rst"
+requires-python = "~=3.10"
```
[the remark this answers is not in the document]
SophieCurinier: `~=` mean compatible release clause, which matches any candidate version that is expected to be compatible with the specified version.
Eg: 
~=1.2 =>   >=1.2.0, <2.0.0
~=1.2.3 => >=1.2.3, <1.3.0

Ref: [hatch compatible release](https://hatch.pypa.io/1.6/config/dependency/#compatible-release)
[PEP 404 compatible release](https://peps.python.org/pep-0440/#compatible-release)

pyproject.toml
```diff
@@ -27,40 +34,146 @@ classifiers = [
...
+    "README.rst",
+    "THIRD_PARTY_SOFTWARE.rst",
+    "docs/**/THIRD_PARTY_SOFTWARE.rst",
+]
+exclude = [
+    "geoh5py/handlers/*",
+    "geoh5py/interfaces/*"
+]
+
+[tool.hatch.build.targets.wheel]
```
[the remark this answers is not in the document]
SophieCurinier: We could but it's not recommanded: 

> Although not recommended, you may define global configuration in the tool.hatch.build table. Keys may then be overridden by target config.

pyproject.toml
SophieCurinier: We could do it with the package [`versiongit`](https://github.com/jwodder/versioningit). It is supported by hatch. Good idea, thank you!

pyproject.toml
```diff
@@ -27,40 +34,146 @@ classifiers = [
...
+    "/geoh5py/handlers/*",
+    "/geoh5py/interfaces/*"
 ]
 
-[tool.poetry.dependencies]
-python = "^3.10"
+[tool.hatch.envs.default]
+dependencies = [
+    "pygments",
+    "scipy~=1.14.0",
```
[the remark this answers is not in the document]
SophieCurinier: Yes, it's only used for the test it's a better idea to keep it under `test` in `[project.optional-dependencies]`.

pyproject.toml
```diff
@@ -27,40 +34,146 @@ classifiers = [
...
+
+[tool.hatch.build.targets.wheel]
 include = [
-    { path = "COPYING", format = ["sdist", "wheel"] },
-    { path = "COPYING.LESSER", format = ["sdist", "wheel"] },
-    { path = "LICENSE", format = ["sdist", "wheel"] },
-    { path = "README.rst", format = ["sdist", "wheel"] },
-    { path = "THIRD_PARTY_SOFTWARE.rst", format = ["sdist", "wheel"] },
-    { path = "docs/**/THIRD_PARTY_SOFTWARE.rst", format = ["sdist", "wheel"] },
+    "/geoh5py",
```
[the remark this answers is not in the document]
SophieCurinier: It's a typing error. I'm gonna correct it.

pyproject.toml
```diff
@@ -27,40 +34,146 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Physics",
     "Topic :: Software Development :: Libraries :: Python Modules"
 ]
+dependencies = [
+    "pydantic~=2.5.2",
```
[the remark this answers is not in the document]
sebhmg: ok, that is confusing, but this one is correct.
I understand that poetry `pydantic = "^2.5.2"`, is equivalent to hatch "pydantic~=2.5.2"`
(see https://python-poetry.org/docs/dependency-specification/#caret-requirements)

pyproject.toml
```diff
@@ -27,40 +34,146 @@ classifiers = [
...
 
+[tool.hatch.version]
+path = "geoh5py/__init__.py"
+
+[tool.hatch.build.targets.sdist] # What to include given the target
+include = [
+    "geoh5py",
+    "COPYING",
+    "COPYING.LESSER",
+    "LICENSE",
```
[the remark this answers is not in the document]
SophieCurinier: Yes, it's needed. If we don't do it, the licenses files aren't shipped with the package.

pyproject.toml
```diff
@@ -27,40 +34,146 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Physics",
     "Topic :: Software Development :: Libraries :: Python Modules"
 ]
+dependencies = [
+    "pydantic~=2.5.2",
+    "h5py",
+    "numpy~=1.26.0",
+    "Pillow~=10.3.0",
```
sebhmg, line 41: **issue:** version constraint used to be stricter with poetry, using "= ~"

Likely means for Hatch:
```suggestion
    "numpy==1.26.*",
    "Pillow==10.3.*",
```

pyproject.toml
sebhmg, line 97:
```suggestion
    "scipy==1.14.*",
```

sebhmg (changes requested):
use stricter constraints on versions for some dependencies

pyproject.toml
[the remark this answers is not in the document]
SophieCurinier: `^` isn't recognize. 😞 

Only the following  versions operators can be used:
    `~=`: [Compatible release](https://peps.python.org/pep-0440/#compatible-release) clause
    `==`: [Version matching](https://peps.python.org/pep-0440/#version-matching) clause
    `!=`: [Version exclusion](https://peps.python.org/pep-0440/#version-exclusion) clause
    `<=`, >=: [Inclusive ordered comparison](https://peps.python.org/pep-0440/#inclusive-ordered-comparison) clause
    `<, >`: [Exclusive ordered comparison](https://peps.python.org/pep-0440/#exclusive-ordered-comparison) clause
    `===`: [Arbitrary equality](https://peps.python.org/pep-0440/#arbitrary-equality) clause.

pyproject.toml
sebhmg: also https://github.com/ofek/hatch-vcs
See https://github.com/ofek/hatch-showcase

.pre-commit-config.yaml
```diff
@@ -11,11 +11,6 @@ ci:
     autoupdate_schedule: monthly
 
 repos:
--   repo: https://github.com/python-poetry/poetry
```
sebhmg, line 14: is there a hook to validate pyproject.toml for hatch ? Or at least that it follows pyproject specifications?

In the past, we got a broken pyproject, not detected in the repo itself since the package itself is not installed in the github workflow, but it broke another repo that was depending on it.

It makes me think that it could be useful if the workflows were installing the package. Which might happen anyway when we move to pixi...
Ok good for now, we can revisit this aspect later

geoh5py/__init__.py
```diff
@@ -17,15 +17,32 @@
...
 import inspect
 
 from geoh5py.workspace import Workspace
 
 from . import groups, objects
 from .groups import CustomGroup
 
+# Define the variable '__version__' at the runtime
+try:
+    from setuptools_scm import get_version
```
sebhmg, line 29: **thought:**
I understand from the "Hatch VCS Footgun Example" that this whole block is it to provide an accurate version from the git tag in editable mode when moving between revisions.

However, will the attempt of setuptools_scm to find a git tag slow down the first import of geoh5?
And if applying the same mechanism across all our package, how much will it slow down the first final application which imports them all?

Trying to have the proper version in edit mode for developers might end up by as a slowdown for the end users. I am no convinced it is worth.

Another preferable approach could be a [post_checkout](https://git-scm.com/docs/githooks#_post_checkout) git hook, that would execute `setuptools_scm.get_version` to patch a possibly existing `_version` file.

The editable use case does not need to be solved now.

sebhmg (changes requested):
please, see comments about version constraints.

Cool auto-versioning with hatch-vcs, but I have some concern with the Hatch VCS Footgun approach for editable mode

.pre-commit-config.yaml
SophieCurinier: Not really. Hatch doesn't have a `check` command, like poetry to verify its pyproject.toml.

Maybe, we could create a command test pyproject we just run one basic test, to check if the project could be build.

SophieCurinier: We could also run `hatch build` to build the project, but it seems a lot of spent time before each commit.

pyproject.toml
```diff
@@ -27,40 +34,156 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Physics",
     "Topic :: Software Development :: Libraries :: Python Modules"
 ]
+dependencies = [
+    "pydantic==2.5.*",
+    "h5py==3.11.*",
```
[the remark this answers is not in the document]
SophieCurinier: I'm gonna exactly convert it with the meaning of `^`. But, for your information, `^` and `~=` are not equivalent. Indeed, `^3.2.1` is means `<4.0.0 >=3.2.1`, but `~=3.2.1` means `<3.3.0 >= 3.2.1`

pyproject.toml
```diff
@@ -27,40 +34,156 @@ classifiers = [
...
+quality = [
+    "pylint",
+    "pre-commit",
+]
+docs = [
+    # Sphinx dependencies
+    "sphinx",
+    "sphinx_issues",
+    "sphinx-autodoc-typehints",
+    "sphinx-rtd-theme",
```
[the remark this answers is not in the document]
SophieCurinier: I'm pretty sure, it's not used because in `readthedocs` used the following configuration:

```
    - sphinx
    - sphinx_issues
    - sphinx_rtd_theme
    - sphinxcontrib-bibtex
    - sphinxcontrib-googleanalytics
    - nbsphinx
    - nbstripout
    - numpydoc
    - jupyter_client
```

.pre-commit-config.yaml
SophieCurinier: If the doesn't have it, we need to delete the following `test_version_is_semver`. Indeed, it used `geoh5py.__version__` variable. However, this variable is not defined.

pyproject.toml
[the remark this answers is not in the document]
sebhmg: or, so it depends whether we give x.y.z or z.y

in that case, should be good with:
```
    "h5py~=3.2",
    "pydantic~=2.5",
```

.pre-commit-config.yaml
sebhmg: version should still be semver. And Analyst expects `__version__`

I think __init__.py needs:
```py
from geoh5py._version import __version__

__all__ = [
    __version__
]
```

Maybe a try catch around the import with that `RuntimeErrror("geoh5py package is not correctly installed. Please install through pip or Conda.")`

sebhmg (changes requested):
need __version__ to be defined

and Analyst expects it to be semver

.pre-commit-config.yaml
SophieCurinier: I updated it the `pyproject.toml` with `versioningit` instead of `hatch-vcs`. It offers us the possibility to format the `__version__` variable. Without it, we don't pass `test_version_is_semver` (at least locally).

geoh5py/__init__.py
```diff
@@ -24,6 +24,15 @@
 from . import groups, objects
 from .groups import CustomGroup
 
+try:
+    from ._version import __version__
+except ModuleNotFoundError:
+    raise (
+        ModuleNotFoundError(
+            "geoh5py is not correctly initialized. Please check if it is correctly built."
```
sebhmg, line 32: "built", or more "installed"?

I do not expect a user to be able to check is it built

sebhmg (approved):

geoh5py/__init__.py
SophieCurinier: The `_version.py` file is generated during build. This message is intended more for developers than for users. `_version.py` is automatically delivered, but is not tracked by git, since it is a generated file.
What do you think @domfournier?

sebhmg: Correct. User would have it installed anyway. Only a developer would have a non-installed local checkout.
Still, it is more about installing with pip (or pip -e) than about building

domfournier (dismissed):
Looks good to me.

PR #591, opened 2024-08-07
GEOPY-1656: Always check/closed EM loops and return warning

domfournier:
**GEOPY-1656 - Always check/closed EM loops and return warning**

### DIFF
--- a/geoh5py/data/color_map.py
+++ b/geoh5py/data/color_map.py
@@ -38,7 +38,7 @@ class ColorMap:
     def __init__(self, **kwargs):
         self._values = np.empty((0, 5))
         self._name = "geoh5py_custom.TBL"
-        self._parent = None
+        self._parent: DataType | None = None
 
         map_attributes(self, **kwargs)
 
@@ -85,10 +85,9 @@ class ColorMap:
             self._values = np.asarray(
                 values, dtype=list(zip(self._names, self._formats, strict=False))
             )
-        self.parent.modified_attributes = "color_map"
 
         if self.parent is not None:
-            self.parent.workspace.update_attribute(self, "color_map")
+            self.parent.workspace.update_attribute(self.parent, "color_map")  # pylint: disable=no-member
 
     @property
     def name(self) -> str:
@@ -100,15 +99,20 @@ class ColorMap:
     @name.setter
     def name(self, value: str):
         self._name = str(value)
-        self.parent.modified_attributes = "color_map"
+
+        if self.parent is not None:
+            self.parent.workspace.update_attribute(self.parent, "color_map")  # pylint: disable=no-member
 
     @property
-    def parent(self):
+    def parent(self) -> DataType | None:
         """Parent data type"""
         return self._parent
 
     @parent.setter
-    def parent(self, data_type: DataType):
+    def parent(self, data_type: DataType | None):
+        if not isinstance(data_type, (DataType, type(None))):
+            raise TypeError(f"Input 'data_type' must be of type {DataType}.")
+
         self._parent = data_type
 
     def __len__(self):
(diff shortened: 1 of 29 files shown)


geoh5py/objects/cell_object.py
```diff
@@ -51,6 +51,19 @@ def __init__(self, object_type: ObjectType, name="Object", **kwargs):
     def default_type_uid(cls) -> uuid.UUID:
         """Default type uid."""
 
+    @property
+    def cells(self) -> np.ndarray:
+        """
+        :obj:`numpy.array` of :obj:`int`: Array of indices
```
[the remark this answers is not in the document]
benk-mira: ..or if you want those `:obj:` references in there for autodoc or something, maybe keep the one-liner concise then drop a line to include in the 'more details' section

geoh5py/objects/object_base.py
```diff
@@ -406,9 +406,7 @@ def create_property_group(
 
         return prop_group
 
-    def find_or_create_property_group(
-        self, name=None, uid=None, **kwargs
-    ) -> PropertyGroup:
+    def fetch_property_group(self, name=None, uid=None, **kwargs) -> PropertyGroup:
```
benk-mira, line 409: Type the arguments, and the docstring doesn't really need  the `:obj:` thing since the return type is annotated.

geoh5py/data/color_map.py
```diff
@@ -85,10 +85,9 @@ def values(self, values: np.ndarray):
             self._values = np.asarray(
                 values, dtype=list(zip(self._names, self._formats, strict=False))
             )
-        self.parent.modified_attributes = "color_map"
 
         if self.parent is not None:
```
benk-mira, line 89: could you get rid of the pylint disable with `if hasattr(self, "parent", None) is not None`?

geoh5py/data/data.py
```diff
@@ -148,6 +149,11 @@ def copy_from_extent(
             **kwargs,
         )
 
+    @classmethod
+    def default_type_uid(cls) -> uuid.UUID | None:
```
benk-mira, line 153: Can this be None?

geoh5py/objects/object_base.py
```diff
@@ -430,6 +428,20 @@ def find_or_create_property_group(
 
         return prop_group
 
+    def find_or_create_property_group(
+        self, name=None, uid=None, **kwargs
+    ) -> PropertyGroup:
+        """
+        Find or create :obj:`~geoh5py.groups.property_group.PropertyGroup`
```
benk-mira, line 435: same comment as for fetch_property_group

geoh5py/objects/object_base.py
```diff
@@ -430,6 +428,20 @@ def find_or_create_property_group(
...
+        from given name and properties.
+        """
+        warnings.warn(
+            "The 'find_and_create_property_group' will be deprecated. "
```
benk-mira, line 439: haha, was just going to suggest this, then I refreshed to see you had read my mind!

geoh5py/shared/entity.py
```diff
@@ -181,6 +181,11 @@ def create(cls, workspace, **kwargs):
         )
         return new_object
 
+    @classmethod
+    @abstractmethod
+    def default_type_uid(cls) -> uuid.UUID | None:
```
benk-mira, line 186: Shouldn't all subclasses just have a default_type_uid?

benk-mira (changes requested):
Some small suggestions

geoh5py/data/color_map.py
domfournier: I tried, doesn't work somehow

geoh5py/data/data.py
domfournier: Yes, Data doesn't have default except for Geometry type

geoh5py/shared/entity.py
domfournier: No

benk-mira (approved):
Thanks

PR #592, opened 2024-08-12
GEOPY-1710: Find more efficient computation for DrapeModel centroids

domfournier:
**GEOPY-1710 - Find more efficient computation for DrapeModel centroids**


MatthieuCMira (approved):
yep, seems better

PR #593, opened 2024-08-12
GEOPY-1645: Update for geoh5py refactor of class instantiation

domfournier:
**GEOPY-1645 - Update for geoh5py refactor of class instantiation**

### DIFF
--- a/geoh5py/objects/curve.py
+++ b/geoh5py/objects/curve.py
@@ -173,7 +173,7 @@ class Curve(CellObject):
         if not np.issubdtype(indices.dtype, np.integer):
             raise TypeError("Indices array must be of integer type")
 
-        if len(indices) > 1 and np.max(indices) > (self.n_vertices - 1):
+        if len(indices) > 0 and np.max(indices) > (self.n_vertices - 1):
             raise ValueError("Found cell indices larger than the number of vertices.")
 
         return indices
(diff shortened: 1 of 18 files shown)


geoh5py/objects/curve.py
```diff
@@ -173,7 +173,7 @@ def validate_cells(self, indices: list | tuple | np.ndarray | None) -> np.ndarra
         if not np.issubdtype(indices.dtype, np.integer):
             raise TypeError("Indices array must be of integer type")
 
-        if np.max(indices) > self.n_vertices - 1:
+        if len(indices) > 1 and np.max(indices) > (self.n_vertices - 1):
```
MatthieuCMira, line 176: I don't get it, len(indices) will always be >1 no?
Don't you mean <=1?

MatthieuCMira (commented):
Seems good to me. 
Just one comment. I'm not sure I'm understanding a line

geoh5py/objects/curve.py
domfournier: It's possible that all vertices get removed, which crashes the np.max()

MatthieuCMira (dismissed):
ok my bad I did not understood

benk-mira (approved):
yup!

PR #594, opened 2024-08-12
GEOPY-1664: Property group type validation is broken

domfournier:
**GEOPY-1664 - Property group type validation is broken**


benk-mira (approved):
Thanks for this!  I'll try to set the two property_group_type options in my gempy-driver ui.json.

PR #595, opened 2024-08-13
GEOPY-1718: allow dependency on not bool data

MatthieuCMira:
**GEOPY-1718 - allow "enabled" on data that are not boolean**
allow ui.json dependency on any type of data


geoh5py/ui_json/utils.py
```diff
@@ -94,13 +94,21 @@ def dependency_requires_value(ui_json: dict[str, dict], parameter: str) -> bool:
     :param parameter: Name of parameter to check type.
     """
     dependency = ui_json[parameter]["dependency"]
-    key = "enabled" if ui_json[dependency].get("optional", False) else "value"
+
+    key = (
+        "enabled"
+        if (
+            ui_json[dependency].get("optional", False)
+            or "enabled" in ui_json[dependency]
```
domfournier, line 102: I am not sure this is going to work since "enabled" can be set to true, eventhough the groupOptional is set to False.

domfournier (changes requested):
I am not sure about this. Can you write a unitest to demonstrate why this is needed?

geoh5py/ui_json/utils.py
MatthieuCMira: I extended the tests

domfournier (approved):

PR #602, opened 2024-09-03
GEOPY-1738: Crash handling GI queries

domfournier:
**GEOPY-1738 - Crash handling GI queries**


markogauk (approved):
Good cleanups with try/excepts

Previous error with unhandled dtype no longer happens

markogauk:
Good cleanups with try/excepts

Previous error with unhandled dtype no longer happens

PR #603, opened 2024-09-05
GEOPY-1706: Add empty string to uid validator and flatten method in base class an…

benk-mira:
**GEOPY-1706 - Add flatten method to forms**
…d overloaded for DataForm to handle isValue logic

### DIFF
--- a/tests/ui_json/forms_test.py
+++ b/tests/ui_json/forms_test.py
@@ -256,6 +256,7 @@ def test_object_form():
         _ = ObjectForm(
             label="name", value=obj_uid, mesh_type=[TypeUID.POINTS, str(uuid.uuid4())]
         )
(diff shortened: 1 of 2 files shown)


tests/ui_json/forms_test.py
```diff
@@ -307,6 +308,44 @@ def test_data_form():
...
+
+    data_uid = str(uuid.uuid4())
+    form = DataForm(
+        label="name",
+        value=data_uid,
+        parent="my_param",
+        association="Vertex",
+        data_type="Float",
+    )
+    assert str(form.flatten()) == data_uid
```
domfournier, line 323: Should we not get it promoted to UUID at this point?

domfournier (approved):

PR #604, opened 2024-09-10
GEOPY-1749: Point simpeg-driver to geoh5py feature 0.10.0

domfournier:
**GEOPY-1749 - Point simpeg-driver to geoh5py feature 0.10.0**


benk-mira (dismissed):
Thanks for this

PR #607, opened 2024-09-12
GEOPY-1758: define dtytpe in DataClass as a class attribute and use this attribute where it's needed

MatthieuCMira:
**GEOPY-1758 - define dtytpe in DataClass as a class attribute and use this attribute where it's needed**
define dtype as class attribute


MatthieuCMira:
on ice

PR #615, opened 2024-09-13
GEOPY-1720: _r2

domfournier:
**GEOPY-1720 - Fix handling of geometric data from reference data table with user names**


MatthieuCMira (approved):
ok,

seems ok oin terms of logic.
Must be tested to see if it's ok.

PR #619, opened 2024-09-18
GEOPY-1755: create an object to access all the data of a property group in a recarray

MatthieuCMira:
**GEOPY-1755 - create an object to access all the data of a property group in a recarray**
add a table for property group.

I used the same way to name the tables as "drillholes_group_table"

Still, I think it could be benificial in the future to name the property "table", 
the "[...]_table" "full" (or even __call__) and the "[...]_table form name" => "from_name()" for simplification for both usages.

### DIFF
--- a/geoh5py/groups/property_group.py
+++ b/geoh5py/groups/property_group.py
@@ -17,18 +17,25 @@
 
 from __future__ import annotations
 
-import uuid
 from collections.abc import Iterable
 from enum import Enum
 from typing import TYPE_CHECKING, Literal
+from uuid import UUID, uuid4
+from warnings import warn
 
 from ..data import Data, DataAssociationEnum
-from ..shared.utils import map_attributes
+from ..shared.utils import (
+    find_unique_name,
+    map_attributes,
+    remove_duplicates_in_list,
+    str2uuid,
+)
 from .property_group_table import PropertyGroupTable
 
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: no cover
     from ..objects import ObjectBase
+    from ..shared import Entity
 
 
 class GroupTypeEnum(str, Enum):
@@ -48,6 +55,13 @@ class PropertyGroup:
     """
     Property group listing data children of an object.
     This group is not registered to the workspace and only visible to the parent object.
+
+    :param parent: Parent object.
+    :param name: Name of the group.
+    :param on_file: Property group is on file.
+    :param uid: Unique identifier.
+    :param property_group_type: Type of property group.
+    :param kwargs: Additional attributes to add to the group
     """
 
     _attribute_map = {
@@ -61,7 +75,6 @@ class PropertyGroup:
     def __init__(  # pylint: disable=too-many-arguments
         self,
         parent: ObjectBase,
-        association: DataAssociationEnum = DataAssociationEnum.VERTEX,
         name=None,
         on_file=False,
         uid=None,
@@ -69,55 +82,47 @@ class PropertyGroup:
         **kwargs,
     ):
         self.name = name or "property_group"
-        self.uid = uid or uuid.uuid4()
-        self._allow_delete = True
+        self.uid = uid or uuid4()
         self.on_file = on_file
-        self.association = association
-        self._property_table = PropertyGroupTable(self)
+        self.property_group_type = property_group_type
 
+        self._allow_delete = True
+        self._association: DataAssociationEnum | None = None
+        self._properties: list[UUID] | None = None
+
+        # define the parent
         if not hasattr(parent, "_property_groups"):
             raise AttributeError(
                 f"Parent {parent} must have a 'property_groups' attribute"
             )
-
         self._parent: ObjectBase = parent
-        self._properties: list[uuid.UUID] | None = None
-        self.property_group_type = property_group_type
-
         parent.add_children([self])
 
         map_attributes(self, **kwargs)
-
         self.parent.workspace.register(self)
 
-    def add_properties(self, data: Data | list[Data | uuid.UUID] | uuid.UUID):
+    def add_properties(self, data: str | Data | list[str | Data | UUID] | UUID):
         """
-        Remove data from the properties.
+        Add data to properties.
+
+        :param data: Data to add to the group.
+            It can be the name, the uuid or the data itself in a list or alone.
         """
-        if isinstance(data, (Data, uuid.UUID)):
+        if not isinstance(data, Iterable):
             data = [data]
 
         properties = self._properties or []
         for elem in data:
-            if isinstance(elem, uuid.UUID):
-                entity = self.parent.get_entity(elem)[0]
-            elif isinstance(elem, Data) and elem in self.parent.children:
-                entity = elem
-            else:
-                continue
-
-            if isinstance(entity, Data) and entity.uid not in properties:
-                properties.append(entity.uid)
+            properties.append(self.verify_data(elem))
 
         if properties:
-            self._properties = properties
+            self._properties = remove_duplicates_in_list(properties)
             self.parent.workspace.add_or_update_property_group(self)
-            self.property_table.update()
 
     @property
     def allow_delete(self) -> bool:
         """
-        :obj:`bool` Allow deleting the group
+        Allow deleting the group
         """
         return self._allow_delete
 
@@ -130,12 +135,21 @@ class PropertyGroup:
     @property
     def association(self) -> DataAssociationEnum:
         """
-        :obj:`~geoh5py.data.data_association_enum.DataAssociationEnum` Data association
+        The association of the data.
         """
+        if self._association is None:
+            return DataAssociationEnum.UNKNOWN
+
         return self._association
 
     @association.setter
     def association(self, value: str | DataAssociationEnum):
+        if self._association is not None:
+            raise UserWarning(
+                "Cannot modify association of an existing property group. "
+                "Consider creating a new property group."
+            )
+
         if isinstance(value, str):
             value = getattr(DataAssociationEnum, value.upper())
 
@@ -149,7 +163,7 @@ class PropertyGroup:
     @property
     def attribute_map(self) -> dict:
         """
-        :obj:`dict` Attribute names mapping between geoh5 and geoh5py
+        Attribute names mapping between geoh5 and geoh5py
         """
         return self._attribute_map
 
@@ -158,6 +172,10 @@ class PropertyGroup:
         """
         The values of the properties in the group.
         """
+        warn(
+            "PropertyGroup.collect_values is deprecated, use PropertyGroup.table instead.",
+            DeprecationWarning,
+        )
 
         if self._properties is None:
             return None
@@ -167,7 +185,7 @@ class PropertyGroup:
     @property
     def name(self) -> str:
         """
-        :obj:`str` Name of the group
+        Name of the group
         """
         return self._name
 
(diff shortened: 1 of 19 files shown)


domfournier (commented):
Sorry lots of comments. Generally a great idea, I would just vote to simplify things a bit.

geoh5py/groups/property_group_table.py
```diff
@@ -0,0 +1,232 @@
...
+
+        if self._property_group.properties is None:
+            return
+
+        for property_ in self._property_group.properties:
+            data = self.property_group.parent.get_data(property_)[0]
+
+            dtype = DataTypeEnum.from_primitive_type(data.entity_type.primitive_type)
+            if dtype not in [np.float32, np.int32, np.uint32, bool]:
+                dtype = "O"
```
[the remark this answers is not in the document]
MatthieuCMira: Those formats refers to "DataTypeEnum", no general types.

The problem with str is that recarray is allocating a size of the array automatically. So if we want to initiate with any string size, it should be an Object type

geoh5py/groups/property_group_table.py
```diff
@@ -0,0 +1,232 @@
...
+        if self.property_group.properties is None:
+            return None
+
+        return self.property_table_by_name(
+            self.property_group.properties,  # type: ignore
+            spatial_index=True,
+        )
+
+    def property_table_by_name(
+        self, names: list[str | UUID] | str | UUID, spatial_index: bool = False
```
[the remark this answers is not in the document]
MatthieuCMira: I replicated it from drillhole concatenate table.

The interest of this method if the property group is huge, it's saving memory just calling the right data

geoh5py/groups/property_group_table.py
```diff
@@ -0,0 +1,232 @@
...
+        The columns of the association table.
+
+        This function is needed as a data can be both associated to cell or
+        vertex in CellObjects.
+        """
+        if self.property_group.association == DataAssociationEnum.VERTEX:
+            return "vertices"
+        if self.property_group.association == DataAssociationEnum.CELL:
+            return "centroids"
+        return "locations"
```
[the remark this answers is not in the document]
MatthieuCMira: Sure, drillhole is a pain. 
Maybe we could raise an error if users are trying to use drillholes?

geoh5py/groups/property_group_table.py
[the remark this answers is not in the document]
domfournier: Ok dok

geoh5py/groups/property_group_table.py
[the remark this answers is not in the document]
domfournier: I understand that, but we can transfer the responsibility on the user to generate a new property group. Doesn't change anything else in terms of value access.

geoh5py/groups/property_group_table.py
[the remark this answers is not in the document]
domfournier: We could implement it later sure, and just raise a NotImplemented error for now if called from a drillhole.

geoh5py/groups/property_group_table.py
```diff
@@ -0,0 +1,162 @@
...
+        if self.property_group.association == DataAssociationEnum.VERTEX:
+            return self.property_group.parent.n_vertices
+
+        if self.property_group.association == DataAssociationEnum.CELL:
+            return self.property_group.parent.n_cells
+
+        raise ValueError(
+            f"The association {self.property_group.association} is not supported. "
+            f"Only VERTEX and CELL associations are supported."
+        )
```
domfournier, line 162:
```suggestion
        return self.locations.shape[0]
```

geoh5py/groups/property_group_table.py
```diff
@@ -0,0 +1,162 @@
...
+        The locations of the association table.
+
+        This function is needed as a data can be both associated to cell or
+        vertex in CellObjects.
+        """
+        if self.property_group.association == DataAssociationEnum.VERTEX:
+            return self.property_group.parent.vertices
+
+        if self.property_group.association == DataAssociationEnum.CELL:
+            return self.property_group.parent.centroids  # type: ignore
```
domfournier, line 134:
```suggestion
        locations = self.property_group.parent.locations
```

geoh5py/groups/property_group_table.py
```diff
@@ -0,0 +1,162 @@
...
+
+        raise ValueError(
+            f"The association {self.property_group.association} is not supported. "
+            f"Only VERTEX and CELL associations are supported."
+        )
```
domfournier, line 139:
```suggestion
    if locations is None:
            raise AttributeError(
            f"The parent object does not have 'locations' defined."
        )
```

geoh5py/groups/property_group_table.py
```diff
@@ -0,0 +1,162 @@
...
+    locations_keys = ("X", "Y", "Z")
+
+    def __init__(self, property_group: PropertyGroup):
+        if not hasattr(property_group, "property_group_type"):
+            raise TypeError("'property_group' must be a PropertyGroup object.")
+
+        if hasattr(property_group.parent, "collar"):
+            raise NotImplementedError(
+                "PropertyGroupTable is not supported for Drillhole objects."
+            )
```
domfournier, line 50: We should be able to support `drillholes` since `drillholes.locations` returns a 2D array of desurveyd xyz positions.

geoh5py/groups/property_group_table.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+    A class to store the information of a property group.
+
+    :param property_group: The property group to extract the data from.
+    """
+
+    locations_keys = ("X", "Y", "Z")
+
+    def __init__(self, property_group: PropertyGroup):
+        if not hasattr(property_group, "property_group_type"):
```
domfournier, line 44:
```suggestion
        if not isinstance(property_group, PropertyGroup):
```

geoh5py/groups/property_group_table.py
domfournier: Also check for assiociation type in `(DataAssociationEnum.CELL, DataAssociationEnum.VERTEX, DataAssociationEnum.DEPTH)`

geoh5py/objects/cell_object.py
```diff
@@ -69,6 +69,20 @@ def cells(self, cells: np.ndarray | list | tuple):
...
 
+    @property
+    def centroids(self) -> np.ndarray | None:
+        """
+        Compute the centroids of the cells.
+        """
+        if self.cells is None:
+            return None
+
+        return np.array([self.vertices[conn].mean(axis=0) for conn in self.cells])
```
domfournier, line 80: That's a bit dangerous change since we often check if centroids is None in other repos. Sort out of scope for this issue anyway. Remove unless absolutely necessary

geoh5py/objects/cell_object.py
```diff
@@ -69,6 +69,20 @@ def cells(self, cells: np.ndarray | list | tuple):
...
+
+    @property
+    def locations(self):
+        return self.vertices
```
domfournier, line 84: Would be better defined on the parent `Points` class, although technically already handled by the `BaseObject`

geoh5py/shared/utils.py
```diff
@@ -833,3 +833,35 @@ def str2none(value):
...
+def remove_duplicates_in_list(input_list: list) -> list:
+    """
+    Remove duplicates from a list without changing the sorting.
+
+    :param input_list: the list to remove duplicates from.
+
+    :return: The sorted list
+    """
+    seen = set()
+    return [x for x in input_list if not (x in seen or seen.add(x))]
```
domfournier, line 867:
```suggestion
    return sorted(set(input_list), key=input_list.index)
```

domfournier (commented):
See comments

geoh5py/groups/property_group_table.py
MatthieuCMira: Cell object can have data with different associations:
Cell and Vertex. 
We have to deal with this scenario.

geoh5py/groups/property_group_table.py
MatthieuCMira: same

geoh5py/groups/property_group_table.py
MatthieuCMira: That's too complicated.
The data of drillholes are not associated with the locations but with the depth or the from-to. 
and a drillhole can contain a bunch of them.

geoh5py/groups/property_group_table.py
MatthieuCMira: circular import

geoh5py/objects/cell_object.py
MatthieuCMira: ok, but it would be nice to have the centroids of the cells.

If in a curve object, the data are associated to cells, we want the coordinates of thoses cells, so the mean of the vertices delimiting it.

geoh5py/objects/cell_object.py
MatthieuCMira: That's to avoid the issue with the cell object

geoh5py/shared/utils.py
MatthieuCMira: I did not used list(sorted) because I wanted users to have the same order as their input data. 
If not, it's sorted by uids, which is kind of random.
Of course, this is a choice to make

geoh5py/groups/property_group_table.py
domfournier: Yea you are right.

geoh5py/groups/property_group_table.py
domfournier: Right.

geoh5py/objects/cell_object.py
domfournier: Ok we can try. 
Can you turn this operation to be vectorial though? Looping will be very expensive here.

geoh5py/objects/cell_object.py
domfournier: OK

geoh5py/shared/utils.py
domfournier: I am not following. The function I gave you above should do the same, but using C code instead of a python loop.

geoh5py/groups/property_group_table.py
domfournier: Ok, just deal with it ince here then.

geoh5py/groups/property_group.py
```diff
@@ -56,67 +71,180 @@ class PropertyGroup:
...
+
+        :return: List of unique identifiers for the Data entities.
+        """
+        if data_list is None:
+            return None
+
+        if not isinstance(data_list, Iterable):
+            raise TypeError(
+                f"Properties must be an iterable of Data. Provided {data_list}"
+            )
```
domfournier, line 220: I think we can remove. This would already be covered by _initialize_properties. I wonder if we could not remove this method altogether and just to the list conversion directly in the init?

geoh5py/groups/property_group.py
```diff
@@ -56,67 +71,180 @@ class PropertyGroup:
...
         for elem in data:
-            if isinstance(elem, uuid.UUID):
-                entity = self.parent.get_entity(elem)[0]
-            elif isinstance(elem, Data) and elem in self.parent.children:
-                entity = elem
-            else:
-                continue
-
-            if isinstance(entity, Data) and entity.uid not in properties:
-                properties.append(entity.uid)
```
domfournier, line 110: Nice

geoh5py/objects/object_base.py
```diff
@@ -181,39 +186,19 @@ def add_data_to_group(
...
-            associations = list(set(associations))
-            if not associations:
-                raise ValueError(
-                    "No children data found on the parent object. "
-                    "Verify that the list of data or uuid provided are children entities."
-                )
-
-            if len(associations) != 1:
-                raise ValueError("All input 'data' must have the same association.")
-
```
domfournier, line 209: Fantastic.

domfournier (dismissed):

geoh5py/groups/property_group.py
MatthieuCMira: right, it's not pertinent anymore

domfournier (approved):

PR #623, opened 2024-09-25
GEOPY-1755: Property group refactor

domfournier:
**GEOPY-1755 - create an object to access all the data of a property group in a recarray**

### DIFF
--- a/geoh5py/groups/property_group.py
+++ b/geoh5py/groups/property_group.py
@@ -26,7 +26,6 @@ from warnings import warn
 from ..data import Data, DataAssociationEnum
 from ..shared.utils import (
     find_unique_name,
-    map_attributes,
     remove_duplicates_in_list,
     str2uuid,
 )
@@ -61,7 +60,6 @@ class PropertyGroup:
     :param on_file: Property group is on file.
     :param uid: Unique identifier.
     :param property_group_type: Type of property group.
-    :param kwargs: Additional attributes to add to the group
     """
 
     _attribute_map = {
@@ -75,32 +73,34 @@ class PropertyGroup:
     def __init__(  # pylint: disable=too-many-arguments
         self,
         parent: ObjectBase,
-        name: str | None = None,
+        association: str | DataAssociationEnum | None = None,
+        allow_delete: bool = True,
+        name: str = "Property Group",
         on_file: bool = False,
         uid: UUID | None = None,
-        property_group_type: GroupTypeEnum = GroupTypeEnum.UNKNOWN,
-        **kwargs,
+        property_group_type: GroupTypeEnum | str = GroupTypeEnum.UNKNOWN,
+        properties: list[UUID | Data] | None = None,
+        **_,
     ):
-        self._on_file: bool = False
-        self._allow_delete = True
-        self._association: DataAssociationEnum | None = None
-        self._properties: list[UUID] | None = None
+        self._parent: ObjectBase = self.validate_parent(parent)
+        self._property_group_type = self.validate_group_type(property_group_type)
 
+        self.allow_delete = allow_delete
         self.name = name or "property_group"
+        self.on_file: bool = on_file
         self.uid = uid or uuid4()
-        self.property_group_type = property_group_type
 
-        # define the parent
-        if not hasattr(parent, "_property_groups"):
-            raise AttributeError(
-                f"Parent {parent} must have a 'property_groups' attribute"
-            )
-        self._parent: ObjectBase = parent
-        parent.add_children([self])
+        if isinstance(properties, Iterable):
+            properties_list = [self.reference_to_data(prop) for prop in properties]
+        else:
+            properties_list = None
 
-        map_attributes(self, **kwargs)
-        self.on_file = on_file
+        self._association: DataAssociationEnum = self.validate_association(
+            association, properties_list
+        )
+        self._properties = self.validate_properties(properties_list)
 
+        self.parent.add_children([self])
         self.parent.workspace.register(self)
 
     def add_properties(self, data: str | Data | list[str | Data | UUID] | UUID):
@@ -115,7 +115,7 @@ class PropertyGroup:
 
         properties = self._properties or []
         for elem in data:
-            properties.append(self.verify_data(elem))
+            properties.append(self.validate_data(elem))
 
         if properties:
             self._properties = remove_duplicates_in_list(properties)
@@ -143,24 +143,6 @@ class PropertyGroup:
             return DataAssociationEnum.UNKNOWN
 
         return self._association
-
-    @association.setter
-    def association(self, value: str | DataAssociationEnum):
-        if self._association is not None:
-            raise UserWarning(
-                "Cannot modify association of an existing property group. "
-                "Consider creating a new property group."
-            )
-
-        if isinstance(value, str):
-            value = getattr(DataAssociationEnum, value.upper())
-
-        if not isinstance(value, DataAssociationEnum):
-            raise TypeError(
-                f"Association must be 'VERTEX', 'CELL' or class of type {DataAssociationEnum}"
-            )
-
-        self._association = value
 
     @property
     def attribute_map(self) -> dict:
@@ -227,24 +209,6 @@ class PropertyGroup:
         """
         return self._properties
 
-    @properties.setter
-    def properties(self, uids: list[str | UUID | Data]):
-        if self._properties is not None:
-            raise UserWarning(
-                "Cannot modify properties of an existing property group. "
-                "Consider using 'add_properties'."
-            )
-
-        if not isinstance(uids, Iterable):
-            raise TypeError(f"Properties must be an iterable of UUID. Provided {uids}")
-
-        self._properties = remove_duplicates_in_list(
-            [self.verify_data(uid) for uid in uids]
-        )
-
-        if self._on_file:
-            self.parent.workspace.add_or_update_property_group(self)
-
     @property
     def properties_name(self) -> list[str] | None:
         """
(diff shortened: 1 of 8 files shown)


geoh5py/groups/property_group.py
```diff
@@ -75,32 +73,34 @@ class PropertyGroup:
...
-        self.property_group_type = property_group_type
 
-        # define the parent
-        if not hasattr(parent, "_property_groups"):
-            raise AttributeError(
-                f"Parent {parent} must have a 'property_groups' attribute"
-            )
-        self._parent: ObjectBase = parent
-        parent.add_children([self])
+        if isinstance(properties, Iterable):
```
MatthieuCMira, line 93: why not putting that in validate properties?

geoh5py/groups/property_group.py
```diff
@@ -75,32 +73,34 @@ class PropertyGroup:
...
+            properties_list = [self.reference_to_data(prop) for prop in properties]
+        else:
+            properties_list = None
```
MatthieuCMira, line 96: so basicaly, users can pass anuthing to properties without raising an error.

Why not changing the previsous properties setter to a private self._properties = _validate_properties function?

geoh5py/groups/property_group.py
```diff
@@ -75,32 +73,34 @@ class PropertyGroup:
...
+        property_group_type: GroupTypeEnum | str = GroupTypeEnum.UNKNOWN,
+        properties: list[UUID | Data] | None = None,
+        **_,
     ):
-        self._on_file: bool = False
-        self._allow_delete = True
-        self._association: DataAssociationEnum | None = None
-        self._properties: list[UUID] | None = None
+        self._parent: ObjectBase = self.validate_parent(parent)
+        self._property_group_type = self.validate_group_type(property_group_type)
```
MatthieuCMira, line 86: those functions could be private? (No telling it should just asking)

geoh5py/groups/property_group.py
```diff
@@ -346,7 +321,29 @@ def uid(self, uid: str | UUID):
...
+            value = getattr(DataAssociationEnum, value.upper())
+
+        if not isinstance(value, DataAssociationEnum):
+            raise TypeError(
+                f"Association must be 'VERTEX', 'CELL' or class of type {DataAssociationEnum}"
+            )
+
+        return value
+
+    def validate_data(self, data: Data | UUID | str) -> UUID:
```
MatthieuCMira, line 344: The add_data function was settiing the association to the first data association if association was None. 
It sounds this mecanic disapear

geoh5py/groups/property_group.py
```diff
@@ -355,41 +352,52 @@ def verify_data(self, data: Data | UUID | str) -> UUID:
...
 
-        if not isinstance(data, Data):
+        if not isinstance(value, GroupTypeEnum):
             raise TypeError(
-                f"Data must be of type Data, UUID or str. Provided {type(data)}"
+                f"'Property group type' must be of type {GroupTypeEnum}, "
+                f"provided {type(value)}"
             )
 
-        if self._association is None:
```
MatthieuCMira, line 386: This was quite convenient.

geoh5py/objects/object_base.py
```diff
@@ -48,7 +48,7 @@ class ObjectBase(EntityContainer):
     """
     Object base class.
```
MatthieuCMira, line 50: "povisible"?

geoh5py/groups/property_group.py
```diff
@@ -355,41 +352,52 @@ def verify_data(self, data: Data | UUID | str) -> UUID:
 
         :return: The uuid of the data.
         """
-        data = str2uuid(data)
+        data = self.reference_to_data(data)
```
MatthieuCMira, line 354:
```suggestion
self._association = data.association if self.association is None

```

geoh5py/groups/property_group.py
```diff
@@ -346,7 +321,29 @@ def uid(self, uid: str | UUID):
...
+    @staticmethod
+    def validate_association(
+        value: str | DataAssociationEnum | None, properties: list[Data] | None
+    ) -> DataAssociationEnum:
+        if properties is None and value is None:
+            raise ValueError(
+                "At least one of 'properties' or 'association' must be provided."
+            )
+
+        if value is None and properties is not None:
```
MatthieuCMira, line 333: The problem is properties can be None

geoh5py/groups/property_group.py
```diff
@@ -346,7 +321,29 @@ def uid(self, uid: str | UUID):
 
         self._uid = uid
 
-    def verify_data(self, data: Data | UUID | str) -> UUID:
+    @staticmethod
+    def validate_association(
+        value: str | DataAssociationEnum | None, properties: list[Data] | None
+    ) -> DataAssociationEnum:
```
MatthieuCMira, line 327: docstring

geoh5py/groups/property_group.py
```diff
@@ -61,7 +60,6 @@ class PropertyGroup:
     :param on_file: Property group is on file.
     :param uid: Unique identifier.
     :param property_group_type: Type of property group.
```
MatthieuCMira, line 62: increase docstring

geoh5py/groups/property_group.py
```diff
@@ -355,41 +352,52 @@ def verify_data(self, data: Data | UUID | str) -> UUID:
...
-        if self._association is None:
-            self.association = data.association
+        return value
 
-        if self.association != data.association:
-            raise ValueError(
-                f"Data '{data.name}' association ({data.association}) "
-                f"does not match group association ({self.association})."
+    @staticmethod
+    def validate_parent(parent: ObjectBase) -> ObjectBase:
```
MatthieuCMira, line 383: docstring

geoh5py/groups/property_group.py
```diff
@@ -355,41 +352,52 @@ def verify_data(self, data: Data | UUID | str) -> UUID:
...
-            if len(data_) == 0 and isinstance(data, UUID):
-                data_temp = self.parent.workspace.load_entity(data, "data", self.parent)
-                data_ = [] if data_temp is None else [data_temp]
-            if len(data_) == 0:
-                raise ValueError(f"Data '{data}' not found in parent {self.parent}")
-            if len(data_) > 1:
+        return data.uid
+
+    @staticmethod
+    def validate_group_type(value: str | GroupTypeEnum) -> GroupTypeEnum:
```
MatthieuCMira, line 364: docstring

geoh5py/groups/property_group.py
```diff
@@ -270,33 +234,44 @@ def property_group_type(self) -> GroupTypeEnum:
...
-    def property_group_type(self, value: str | GroupTypeEnum | None):
-        if (
-            getattr(self, "_property_group_type", GroupTypeEnum.UNKNOWN)
-            is not GroupTypeEnum.UNKNOWN
-        ):
-            raise UserWarning(
-                "Cannot modify 'property group type' of an existing property group. "
-                "Consider creating a new property group."
-            )
+    def reference_to_data(self, data: str | Data | UUID) -> Data:
```
MatthieuCMira, line 237: I have the feeling this could be a nice utils function no?
just passing parent as an attribute

geoh5py/objects/object_base.py
```diff
@@ -181,20 +181,18 @@ def add_data_to_group(
 
         :return: The target property group.
         """
-        if isinstance(data, (Data, uuid.UUID)):
+        if isinstance(data, Data | uuid.UUID | str):
             data = [data]
 
         if isinstance(property_group, str):
             associations = []
```
MatthieuCMira, line 188: All of this mecanics are duplicates with the new property_group

geoh5py/shared/concatenation/object.py
```diff
@@ -148,9 +149,13 @@ def property_groups(self) -> list | None:
                 property_groups = []
 
             for key in property_groups:
-                self.fetch_property_group(
-                    **self.concatenator.get_concatenated_attributes(key), on_file=True
-                )
+                attributes = {
+                    INV_KEY_MAP.get(key, key): val
```
MatthieuCMira, line 153: question: Why this option could'nt be implemented into the get_concatenated_attributes function?
Like inv_key_map: dict | None, If isinstance(inv_key_map, dict) => convert

MatthieuCMira (commented):
First review.


Missing some docstring for the function,
I am not confident about the association mecanics.

geoh5py/groups/property_group.py
domfournier: No still there, under validate association

geoh5py/groups/property_group.py
domfournier: Yep

geoh5py/groups/property_group.py
domfournier: Because `validate_properties` calls your `validate_data` method, and checks for the association to match. Here I just want to get a formatted list, so that validate association can set the association if not provided.

geoh5py/groups/property_group.py
domfournier: See answer below.

geoh5py/groups/property_group.py
domfournier: Yes but not the behaviour we want I would say. Association is required by geoh5, so I don't think we want allow for None

geoh5py/objects/object_base.py
domfournier: typo

geoh5py/groups/property_group.py
domfournier: Rather be explicit about checking for association.

geoh5py/groups/property_group.py
domfournier: Yes, but we shouldn't allow both to be None. Either you provide an association or you provide properties. You can't have both be None because it makes property_group undefined

geoh5py/groups/property_group.py
domfournier: I think we could move it to the ObjectBase actually, and just call the `parent.reference_to_data`? I will let you modify it on your PR if you want.

geoh5py/objects/object_base.py
domfournier: Could be, yes. I will let you modify on your side too.

geoh5py/shared/concatenation/object.py
domfournier: I tried, but get concatenated_attribute gets called in other places and it started to be an heavy change. This is bare minimum I needed to get rid of map_keys in the property_group instantiation.

geoh5py/groups/property_group.py
MatthieuCMira: ok I get the logic. If geoh5py requieres it, I'll let it there

PR #625, opened 2024-10-01
GEOPY-1669: Final clean up of Object class

MatthieuCMira:
**GEOPY-1669 - Final clean up of Object class**
I changed the object base a bit. 

I added some comments "#todo: ..." within the code. 
Can you comment back and then, and I'll correct/remove them if needed

### DIFF
--- a/geoh5py/data/data.py
+++ b/geoh5py/data/data.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 import uuid
 from abc import abstractmethod
 from typing import Any
+from warnings import warn
 
 import numpy as np
 
@@ -255,6 +256,34 @@ class Data(Entity):
                 return mask_by_extent(self.parent.centroids, extent, inverse=inverse)
 
         return None
+
+    def remove_values(self, indices: np.ndarray):
+        """
+        Remove values based on indices.
+
+        :param indices:
+        """
+        if not isinstance(indices, np.ndarray):
+            raise TypeError("Indices must be a numpy array.")
+
+        if not isinstance(self.values, np.ndarray):
+            warn(
+                f"The values of {self} are not an array but a {type(self.values)}."
+                "The values cannot be removed."
+            )
+            return
+
+        self._values = np.delete(self.values, indices, axis=0)
+
+        if self._values.size == 0:
+            self._values = None
+
+        self._association = self.validate_association(
+            self.parent.find_association(self._values)
+        )
+
+        if self.on_file:
+            self.workspace.update_attribute(self, "values")
 
     def validate_entity_type(self, entity_type: DataType | None) -> DataType:
         """
--- a/geoh5py/data/data.py
+++ b/geoh5py/data/data.py
@@ -20,7 +20,6 @@ from __future__ import annotations
 import uuid
 from abc import abstractmethod
 from typing import Any
-from warnings import warn
 
 import numpy as np
 
@@ -256,34 +255,6 @@ class Data(Entity):
                 return mask_by_extent(self.parent.centroids, extent, inverse=inverse)
 
         return None
-
-    def remove_values(self, indices: np.ndarray):
-        """
-        Remove values based on indices.
-
-        :param indices:
-        """
-        if not isinstance(indices, np.ndarray):
-            raise TypeError("Indices must be a numpy array.")
-
-        if not isinstance(self.values, np.ndarray):
-            warn(
-                f"The values of {self} are not an array but a {type(self.values)}."
-                "The values cannot be removed."
-            )
-            return
-
-        self._values = np.delete(self.values, indices, axis=0)
-
-        if self._values.size == 0:
-            self._values = None
-
-        self._association = self.validate_association(
-            self.parent.find_association(self._values)
-        )
-
-        if self.on_file:
-            self.workspace.update_attribute(self, "values")
 
     def validate_entity_type(self, entity_type: DataType | None) -> DataType:
         """
--- a/geoh5py/objects/cell_object.py
+++ b/geoh5py/objects/cell_object.py
@@ -17,7 +17,6 @@
 
 from __future__ import annotations
 
-import warnings
 from abc import ABC, abstractmethod
 from uuid import UUID
 
@@ -53,6 +52,7 @@ class CellObject(Points, ABC):
         """
         Array of indices defining connecting vertices.
         """
+        # todo: technically this is not possible anymore
         if self._cells is None and self.on_file:
             self._cells = self.workspace.fetch_array_attribute(self, "cells")
 
@@ -118,11 +118,6 @@ class CellObject(Points, ABC):
         :param indices: Indices of cells to be removed.
         :param clear_cache: Clear cache of data values.
         """
-
-        if self.cells is None:
-            warnings.warn("No cells to be removed.", UserWarning)
-            return
-
         if isinstance(indices, (list, tuple)):
             indices = np.array(indices)
 
@@ -135,19 +130,12 @@ class CellObject(Points, ABC):
         ):
             raise ValueError("Found indices larger than the number of cells.")
 
-        children = []
-        for child in self.children:
-            if (
-                isinstance(child, Data)
-                and child.values is not None
-                and child.association == DataAssociationEnum.CELL
-            ):
-                children.append(child)
-
         cells = np.delete(self.cells, indices, axis=0)
-
+        self.load_children_values()
         self._cells = self.validate_cells(cells)
-        self.remove_children_values(indices, children, clear_cache=clear_cache)
+        self._remove_children_values(
+            indices, DataAssociationEnum.CELL, clear_cache=clear_cache
+        )
         self.workspace.update_attribute(self, "cells")
 
     def remove_vertices(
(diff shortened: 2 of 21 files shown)


MatthieuCMira:
Other general questions:
I have the feeling the "object" is not being updated to the workspace in the functions. 
Is this normal?

geoh5py/objects/cell_object.py
```diff
@@ -53,6 +52,7 @@ def cells(self) -> np.ndarray:
         """
         Array of indices defining connecting vertices.
         """
+        # todo: technically this is not possible anymore
```
domfournier, line 55: It is possible during a copy with `clear_cache` options. It was a feature added to avoid increasing the memory requirements on the copy of large files.

geoh5py/objects/cell_object.py
```diff
@@ -118,11 +118,6 @@ def remove_cells(self, indices: list[int] | np.ndarray, clear_cache: bool = Fals
         :param indices: Indices of cells to be removed.
         :param clear_cache: Clear cache of data values.
         """
-
-        if self.cells is None:
-            warnings.warn("No cells to be removed.", UserWarning)
-            return
-
```
domfournier, line 125: Keep it.

tests/curve_data_test.py
```diff
@@ -191,12 +191,25 @@ def test_remove_cells_data(tmp_path: Path):
         with pytest.raises(TypeError, match="Indices must be a list or numpy array."):
             curve.remove_vertices("abc")
 
+        # print(curve.get_data("cellValues")[0].values)
```
domfournier, line 194: remove commented code

geoh5py/objects/object_base.py
```diff
@@ -71,28 +70,50 @@ def __init__(self, last_focus: str = "None", **kwargs):
...
         prop_group_uids = {prop_group.uid: prop_group for prop_group in property_groups}
         children_uids = {child.uid: child for child in self._children}
 
         for child in children:
-            if child.uid not in children_uids and isinstance(
-                child, (Data, PropertyGroup)
+            if (
+                isinstance(child, (Data, PropertyGroup))
+                and child.uid not in children_uids
+                and child.uid not in prop_group_uids
```
domfournier, line 110: Duplicate check -Technically PropertyGroup are already in the children list.

geoh5py/workspace/workspace.py
```diff
@@ -439,12 +442,12 @@ def create_data(
...
     def create_entity(
         self,
-        entity_class,
+        entity_class: type[TYPE],
         compression: int = 5,
         entity: dict | None = None,
         entity_type: EntityType | dict | None = None,
         save_on_creation: bool = True,
-    ) -> Entity:
+    ) -> TYPE:
```
domfournier, line 450: Why the change?

geoh5py/objects/object_base.py
```diff
@@ -136,26 +163,25 @@ def add_data(
...
-                if key in ["parent", "association", "entity_type", "type"]:
-                    continue
-
-                kwargs[key] = val
+                if key not in ["parent", "entity_type", "type"]:
+                    kwargs[key] = val
 
             data_object = self.workspace.create_entity(
-                Data, entity=kwargs, entity_type=entity_type, compression=compression
+                Data,  # type: ignore
```
domfournier, line 179: Better to keep the typing on workspace create_entity then a blind type ignore

geoh5py/objects/object_base.py
```diff
@@ -136,26 +163,25 @@ def add_data(
...
+                entity=kwargs,
+                entity_type=entity_type,
+                compression=compression,
             )
 
-            if not isinstance(data_object, Data):
-                continue
-
```
domfournier, line 158: Same

geoh5py/objects/object_base.py
```diff
@@ -674,8 +698,9 @@ def validate_entity_type(self, entity_type: ObjectType | None) -> ObjectType:
     @property
     def visual_parameters(self) -> VisualParameters | None:
         """
-        Access the visual parameters of the object.
+        The visual parameters of the object.
         """
+        # todo: it sounds like I could have several visual parameters to object
```
domfournier, line 703: GA wouldn't let you, but it's true that we don't have mechanics to prevent adding more...

geoh5py/objects/object_base.py
```diff
@@ -430,44 +503,63 @@ def last_focus(self, value: str):
...
+        The number of cells.
         """
-        return None
+        raise AttributeError(
+            f"Object {self.__class__.__name__} does not have the attribute 'n_cells'."
+        )
 
     @property
-    def n_vertices(self):
+    def n_vertices(self) -> int:
```
domfournier, line 544: Same here

geoh5py/objects/object_base.py
```diff
@@ -430,44 +503,63 @@ def last_focus(self, value: str):
...
+        self, extent: np.ndarray, inverse: bool = False
+    ) -> np.ndarray | None:
         if self.extent is None or not box_intersect(self.extent, extent):
             return None
 
         return mask_by_extent(self.locations, extent, inverse=inverse)
 
     @property
-    def n_cells(self):
+    def n_cells(self) -> int:
```
domfournier, line 535:
```suggestion
    @abstractmethod
    def n_cells(self) -> int:
```

Maybe we should turn those as @abstracmethod, and force the sub-classes to implement it.

geoh5py/workspace/workspace.py
MatthieuCMira: Because now the typing of this function will be the class type you sent it as attribute.
As there is no risk the type change (or it means an error is raised), it's super convenient!

geoh5py/objects/object_base.py
MatthieuCMira: it's not a type:ignore, it's because Data is an abstact class. 
The typing out of this function is "Data", but because it's an abstract class, mypy don't like that... 
I'll find a way to make it better.

geoh5py/objects/object_base.py
MatthieuCMira: I will add one then

geoh5py/data/data.py
```diff
@@ -256,6 +257,34 @@ def mask_by_extent(
 
         return None
 
+    def remove_values(self, indices: np.ndarray):
```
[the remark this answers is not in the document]
MatthieuCMira: It's used when removing a vertice or a cell from an object.
We have to remove the values of the data.
But it was some dangerous functions on "ObjcetBase"

Now, if we want to remove a values, it won't be unstable.

MatthieuCMira:
I have the feeling the changes made on an ObjectBase are never saved to GA. 

```
        if self.on_file:
            self.workspace.update_attribute(self, 
            ```
            
is this normal?

File: geoh5py/objects/cell_object.py (hunk 2 above)
Comment (MatthieuCMira) [reply geoh5py/objects/cell_object.py:125]: Technically, if the cell is None, it will be loaded from the file. 
So they are no chance cell being None! even with clear_cache,
Because you are reloading cells.
And you can't have a cell object with empty cells now!

File: geoh5py/workspace/workspace.py (hunk 5 above)
Comment (MatthieuCMira) [reply geoh5py/workspace/workspace.py:450]: I just removed the typing

File: geoh5py/objects/object_base.py (hunk 7 above)
Comment (MatthieuCMira) [reply geoh5py/objects/object_base.py:158]: when passing "Data" to create_entity, it's redudndant to check if the outptut is "Data" no?

File: geoh5py/objects/object_base.py (hunk 10 above)
Comment (MatthieuCMira) [reply geoh5py/objects/object_base.py:535]: I simply removed them, has it does not make sense to have a grid with vertices.

File: geoh5py/objects/object_base.py (hunk 9 above)
Comment (MatthieuCMira) [reply geoh5py/objects/object_base.py:544]: removed

**Comment by MatthieuCMira** [comment]
Should we remove faces?
I understand it can be implemented in the future, but for now, it's used nowhere

File: geoh5py/objects/object_base.py [hunk 12]
```diff
@@ -354,6 +335,34 @@ def create_property_group(
...
+        """
+        if self.locations is None:
+            return None
+
+        return np.c_[self.locations.min(axis=0), self.locations.max(axis=0)].T
+
+    @property
+    def faces(self) -> np.ndarray:
+        """Object faces."""
+        # todo: this is used nowhere... remove?
```
domfournier, line 361: Not sure, will need to ask Julien. Might just not be supported by our API... It is technically an accepted association type, so something should be using it.

geoh5py/objects/object_base.py
```diff
@@ -562,34 +601,7 @@ def remove_property_group(self, property_group: PropertyGroup):
...
-
-        return out
-
-    @property
-    def converter(self):
-        """
-        :return: The converter for the object.
-        """
-        return self._converter
+            # todo: this should suppress the property group object too? del property_group
```
domfournier, line 604: Either make the `remove_property_group` method private, or move that little piece of logic back inside the `remove_children`, so that the PropertyGroup is guaranteed to be deleted by the workspace.

domfournier (commented):
Getting there. Few more sticky points.

domfournier (approved):

PR #627, opened 2024-10-02
GEOPY-1739: add new InterpretationSection group

MatthieuCMira:
**GEOPY-1739 - Implement Interpretation group type**
new group


domfournier (commented):
There is a bit more to it I think.

image

MatthieuCMira:
Apparement, le "Section Object UID"
Correspond à un object: "clipping plane"
pas encore implémenté dans geoh5py

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+
+        super().__init__(**kwargs)
+
+    def _load_attributes(self):
+        """
+        Load the data from the metadata only to please mypy.
+        """
+        _ = self.interpretation_curves
+        _ = self.interpretation_sections
+        _ = self.section_object_id
```
domfournier, line 143: Let's try to get around this better. It may help to know that metadata are always loaded by the reader, so it will either be None or a dict if it exsits.

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+        Check if the parameters are in the list.
+
+        :param params_list: The list of parameters.
+
+        :return: The index of the parameters in the list else None.
+        """
+        for idx, param in enumerate(params_list):
+            if param.map_back == self.map_back:
+                return idx
+        return None
```
domfournier, line 72: This is pretty much the same as doing `self in params_list`. I don't think it makes sense to have this logic at the Section level, but rather query the `self._interpretation_sections`

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+    def map_back(self) -> dict:
+        """
+        Map the attributes back to the original keys of attribute map.
+
+        :return: The mapped data.
+        """
+        return {key: getattr(self, value) for key, value in ATTRIBUTE_MAP.items()}
+
+    @staticmethod
+    def verify(
```
domfournier, line 96:
```suggestion
    def validate_section(
```

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+
+        :return: An instance of the class.
+        """
+        if isinstance(params, dict):
+            return InterpretationSectionParams(**params)
+        if not isinstance(params, InterpretationSectionParams):
+            raise TypeError(
+                "Interpretation section must be an InterpretationSectionParams"
+                f"or a dict, {type(params)} provided."
+            )
```
domfournier, line 112: Also seems out of place. This kind of validation should belong to the class holding the sections

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+        }
+    )
+
+    _TYPE_UID = UUID(fields=(0x27EE59E1, 0xB1CE, 0x41EC, 0x8A, 0x86, 0x8BD3D229E198))
+    _default_name = "Interpretation Section"
+
+    def __init__(self, **kwargs):
+        self._can_add_group: bool = False
+        self._interpretation_curves: list[Curve] | None = None
+        self._interpretation_sections: list[InterpretationSectionParams] | None = None
```
domfournier, line 132:
```suggestion
        self._interpretation_curves: tuple[Curve] = ()
        self._interpretation_sections: tuple[InterpretationSectionParams] = ()
```
Makes more sense to be immutable since we don't want the user to alter the list directly.

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+    _attribute_map.update(
+        {
+            "Can add group": "can_add_group",
+        }
+    )
+
+    _TYPE_UID = UUID(fields=(0x27EE59E1, 0xB1CE, 0x41EC, 0x8A, 0x86, 0x8BD3D229E198))
+    _default_name = "Interpretation Section"
+
+    def __init__(self, **kwargs):
```
domfournier, line 129:
```suggestion
    def __init__(self, can_add_group: bool = False, clipping_ids: uid = uuid.UUID4, **kwargs):
```
Following the same logic as we did during the refactor, the init should be able to set the attributes specific to the class

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+        self._can_add_group: bool = False
+        self._interpretation_curves: list[Curve] | None = None
+        self._interpretation_sections: list[InterpretationSectionParams] | None = None
+        self._section_object_id: Slicer | None = None
```
domfournier, line 133:
```suggestion
        self._section_object_id: Slicer | None = None\
        self.clipping_ids = clipping_ids
```

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+        {
+            "Can add group": "can_add_group",
+        }
+    )
+
+    _TYPE_UID = UUID(fields=(0x27EE59E1, 0xB1CE, 0x41EC, 0x8A, 0x86, 0x8BD3D229E198))
+    _default_name = "Interpretation Section"
+
+    def __init__(self, **kwargs):
+        self._can_add_group: bool = False
```
domfournier, line 130:
```suggestion
        self._can_add_group: bool = can_add_group
```

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+    def _load_from_metadata(self, key: str) -> Any:
+        """
+        Load the metadata from the key.
+
+        :param key: the key to load the metadata from.
+
+        :return: The value store in the metadata if exists.
+        """
+        metadata = self.metadata if self.metadata is not None else {}
+        return metadata.get(key, None)
```
domfournier, line 154:
```suggestion
        return self.metadata.get(key, None) if self.metadata else None
```

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+        :param values: The values to update.
+        """
+        if values:
+            self.update_metadata(
+                {
+                    key: [
+                        value.map_back
+                        if isinstance(value, InterpretationSectionParams)
+                        else str(value.uid)
+                        for value in values
```
domfournier, line 170: Too complex. Just implement the metadata update directly on the setter of Sections and Curves instead

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+                    ]
+                }
+            )
+        else:
+            setattr(self, key, None)
+            self.update_metadata({key: None})
```
domfournier, line 176: Not sure what is happening here. We are resetting the curves or sections to None? When is this needed?

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+    def add_interpretation_curve(self, curves: Curve | list[Curve]):
+        """
+        Add an interpretation curve.
+
+        :param curves: The interpretation curve to add.
+        """
+        if not isinstance(curves, list):
+            curves = [curves]
+
+        self._load_attributes()
```
domfournier, line 208: Should already be ready to go.

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+            value = str2uuid(value)
+            value = self.workspace.get_entity(value)[0]
+
+        if getattr(value, "_default_name", None) != key:
+            raise TypeError(
+                f"The {value} object must be a {key} object."
+                f" {type(value)} provided."
+            )
+
+        return value
```
domfournier, line 197: A lot of code to just check `isinstance(obj, Curve)`. Don't really think we need this.

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+    def can_add_group(self, value: bool):
+        """
+        Set the can add group attribute.
+
+        :param value: The can add group attribute.
+        """
+        self._can_add_group = bool(value)
+
+        if self.on_file:
+            self.workspace.update_attribute(self, "can_add_group")
```
domfournier, line 261:
```suggestion
            self.workspace.update_attribute(self, "attributes")
```

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+        """
+        if not isinstance(sections, list):
+            sections = [sections]
+
+        self._load_attributes()
+        for section in sections:
+            section = InterpretationSectionParams.verify(section)
+            if self._interpretation_sections is None:
+                self._interpretation_sections = [section]
+            elif section.in_list(self._interpretation_sections) is None:
```
domfournier, line 235: Flip the logic here.
```suggestion
            elif section not in self._interpretation_sections:
```

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+            self._interpretation_curves = None
+        else:
+            if not isinstance(curves, list):
+                raise TypeError("Interpretation curves must be a list of Curve.")
+
+            self._interpretation_curves = [
+                self._verify_object(curve, "Curve") for curve in curves
+            ]
+
+            self.add_children(self._interpretation_curves)  # type: ignore
```
domfournier, line 295: Doesn't seem to do the trick right now
image

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+
+        if self._interpretation_curves is None:
+            self.interpretation_curves = self._load_from_metadata(
+                "Interpretation curves"
+            )
+
+        return self._interpretation_curves
+
+    @interpretation_curves.setter
+    def interpretation_curves(self, curves: list[Curve] | None):
```
domfournier, line 279: I am not sure if setting the list directly makes sense. Maybe an `add_interpretation_curve` instead? We also need to assign a `clipping_ids` attribute to the curve that corresponds to the slicer. So we sort of need more explicit checks on the assignment.

image

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,411 @@
...
+            return
+
+        if not isinstance(sections, list):
+            sections = [sections]
+
+        for section in sections:
+            section = InterpretationSectionParams.verify(section)
+            idx = section.in_list(self._interpretation_sections)
+            if idx is not None:
+                del self._interpretation_sections[idx]
```
domfournier, line 377:
```suggestion
        new_list = []
        for section in self._interpretation_sections:
            if section not in sections:
                new_list = section
        
        self._update_to_metadata(
            "Interpretation sections", new_list
        )  
```

Also need to cut out the curves that belong to that section.

geoh5py/shared/entity.py
```diff
@@ -310,15 +310,22 @@ def metadata(self, value: dict | np.ndarray | bytes | None):
         if self.on_file:
             self.workspace.update_attribute(self, "metadata")
 
-    def update_metadata(self, value: dict | None):
+    def update_metadata(self, value: dict):
```
domfournier, line 313:
```suggestion
    def update_metadata(self, value: dict | None):
```
The None allows to clear out metadata.

domfournier (commented):
Lots to go over. Maybe we can talk first.

geoh5py/groups/interpretation_section.py
MatthieuCMira: The problem was with mypy. 
The mechanics where working,..

geoh5py/groups/interpretation_section.py
domfournier: I guess I don't see the use case of removing a section by using parameter values - it is both very unrealiable (float comparison is always tricky) and not very practical. Makes sense to let users remove one from the list though.

MatthieuCMira (commented):
Correction Done.

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,312 @@
...
+        params: dict | InterpretationSectionParams,
+    ) -> InterpretationSectionParams:
+        """
+        Verify/Create the interpretation section from a mapper.
+
+        :param params: The parameters to verify.
+
+        :return: An instance of the class.
+        """
+        if isinstance(params, Mapping):
```
domfournier, line 65: Fancy new type `Mapping`. Why is that?

geoh5py/groups/interpretation_section.py
```diff
@@ -0,0 +1,312 @@
...
+    _default_name = "Interpretation Section"
+
+    def __init__(self, can_add_group: bool = False, **kwargs):
+        self._can_add_group: bool = self._validate_can_add_group(can_add_group)
+
+        super().__init__(**kwargs)
+
+        self._interpretation_curves: tuple[Curve] | tuple = ()
+        self._interpretation_sections: tuple[InterpretationSectionParams] | tuple = ()
+        self._section_object_id: Slicer | None = None
```
domfournier, line 95:
```suggestion
        self._section_object: Slicer | None = None
```
Makes more sense to just call it `section_object` I think, since not a uid

domfournier (dismissed):
Minor comment about naming. Moving to test

geoh5py/groups/interpretation_section.py
MatthieuCMira: a Mapping is a mapper containing key and values.
Just to be more fancy than "dict. but it could be "dict" too

domfournier: Sounds special. I ll let you have it ;)

domfournier (dismissed):

domfournier (approved):
Thanks for the partial clean up of contanated classes

domfournier (approved):

PR #632, opened 2024-10-07
GEOPY-1670: change shutil move for copy

MatthieuCMira:
**GEOPY-1670 - cannot finalize workspace if tmp folder is not on the destination disk**
created a function, but I cannot create a unitest on it.

Guess the back and forth of testing will be there...

### DIFF
--- a/geoh5py/shared/utils.py
+++ b/geoh5py/shared/utils.py
@@ -17,7 +17,6 @@
 
 from __future__ import annotations
 
-import shutil
 from abc import ABC
 from collections.abc import Callable
 from contextlib import contextmanager
@@ -865,26 +864,3 @@ def remove_duplicates_in_list(input_list: list) -> list:
     :return: The sorted list
     """
     return sorted(set(input_list), key=input_list.index)
-
-
-def resilient_copy(source: str | Path, destination: str | Path):
-    """
-    Copy a file and ignore errors.
-
-    This function is for the case of a copy in a different folder under unbutu.
-
-    :param source: The source file to copy.
-    :param destination: The destination file to copy to.
-    """
-    permission_error = False
-    try:
-        shutil.copy2(source, destination)
-    except PermissionError as error:
-        permission_error = True
-        warn(
-            f"First copy attempt failed due to: {error}"
-            "Trying again with shutil.copy."
-        )
-
-    if permission_error:
-        shutil.copy(source, destination)
--- a/geoh5py/shared/utils.py
+++ b/geoh5py/shared/utils.py
@@ -17,7 +17,7 @@
 
 from __future__ import annotations
 
-import warnings
+import shutil
 from abc import ABC
 from collections.abc import Callable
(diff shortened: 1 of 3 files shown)


geoh5py/shared/utils.py
```diff
@@ -864,3 +865,26 @@ def remove_duplicates_in_list(input_list: list) -> list:
     :return: The sorted list
     """
     return sorted(set(input_list), key=input_list.index)
+
+
+def resilient_copy(source: str | Path, destination: str | Path):
```
sebhmg, line 870: **issue:** desired behaviour is not to copy systematically

We want a "resiliant move". The code should still do a move when it can.

The suggestion was telling about a `resiliant_copy` function, but only to pass it a fallback to `shutils.move`:
```py
shutils.move(source, destination, resiliant_copy)
```

Now, there are only a few places to change: no need for anything complicated, no need for a new function: simply replace the few calls to `shutils.move(...)` by `shutils.move(..., copy)`

This tells `shutils.move()` to use `copy` as a fallback instead of `copy2`.
Since we "move" from a tmp folder, or for a save_as, we do not care about preserving file attributes though `copy2`

geoh5py/ui_json/utils.py
```diff
@@ -326,7 +325,7 @@ def monitored_directory_copy(
         with Workspace.create(working_path / temp_geoh5) as w_s:
             entity.copy(parent=w_s, copy_children=copy_children)
 
-    shutil.move(
+    resilient_copy(
         working_path / temp_geoh5,
         directory_path / temp_geoh5,
     )
```
sebhmg, line 331:
```suggestion
    shutil.move(
        working_path / temp_geoh5,
        directory_path / temp_geoh5,
        copy
    )
```

geoh5py/workspace/workspace.py
```diff
@@ -212,7 +212,7 @@ def close(self):
                     stdout=subprocess.DEVNULL,
                 )
                 Path(self._h5file).unlink()
-                shutil.move(temp_file, self._h5file)
+                resilient_copy(temp_file, self._h5file)
```
sebhmg, line 215:
```suggestion
                shutil.move(temp_file, self._h5file, copy)
```

geoh5py/workspace/workspace.py
```diff
@@ -1329,7 +1329,7 @@ def save_as(self, filepath: str | Path) -> Workspace:
         elif self.h5file is None:
             raise ValueError("Input 'h5file' file must be specified.")
         else:
-            shutil.copy(self.h5file, filepath)
+            resilient_copy(self.h5file, filepath)
```
sebhmg, line 1332:
```suggestion
            shutil.move(self.h5file, filepath, copy)
```

sebhmg (changes requested):
Te code should still do a move when it can. Please see suggestions

sebhmg (approved):

PR #634, opened 2024-10-08
GEOPY-1733: Block model centroids not computed correctly for negative cell delimiters

benk-mira:
**GEOPY-1733 - Block model centroids not computed correctly for negative cell delimiters**

### DIFF
--- a/geoh5py/objects/block_model.py
+++ b/geoh5py/objects/block_model.py
@@ -69,8 +69,9 @@ class BlockModel(GridObject):
 
         :param axis: Axis to get the centers for.
         """
-        n_cells = getattr(self, f"{axis}_cells")
-        return np.cumsum(n_cells) - n_cells / 2.0
+        out = np.cumsum(getattr(self, f"{axis}_cell_delimiters"))
+        out[2:] = out[2:] - out[:-2]
+        return out[2 - 1 :] / 2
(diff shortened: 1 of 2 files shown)


geoh5py/objects/block_model.py
```diff
@@ -69,8 +69,9 @@ def local_axis_centers(self, axis: str) -> np.ndarray:
 
         :param axis: Axis to get the centers for.
         """
-        n_cells = getattr(self, f"{axis}_cells")
-        return np.cumsum(n_cells) - n_cells / 2.0
+        out = np.cumsum(getattr(self, f"{axis}_cell_delimiters"))
+        out[2:] = out[2:] - out[:-2]
+        return out[2 - 1 :] / 2
```
domfournier, line 74: `[2-1:]`? Is that not just `[1:]`?
Could also do it in one go with a convolution.

`return np.convolve(getattr(self, f"{axis}_cell_delimiters"), np.ones(2)/2)[1:-1]`

benk-mira: whoops, I adapted an n-length window to length two and didn't notice the simplification.  I'll stick with the cumsum implementation because it should be more efficient

domfournier (approved):

PR #641, opened 2024-10-16
GEOPY-523: Update typing for shape properties

sebhmg:
**GEOPY-523 - Move iso-surface app to separate repo**


sebhmg:
branch is obsolete: typing was already applied, or shape functions have moved

PR #642, opened 2024-10-16
GEOPY-1645: Fix metadata read

sebhmg:
**GEOPY-1645 - Update for geoh5py refactor of class instantiation**


sebhmg:
obsolete changes. They have already been applied

PR #648, opened 2024-10-21
GEOPY-1701: Add validation with mesh_type on ObjectForm

benk-mira:
**GEOPY-1701 - Add validation with mesh_type on ObjectForm**


geoh5py/ui_json/validations/uijson.py
```diff
@@ -0,0 +1,90 @@
...
+
+def parent_validation(name: str, data: dict[str, Any], params: dict[str, Any]):
+    """Data is a child of the parent object."""
+
+    form = params[name]
+    child = data[name]
+    parent = data[form["parent"]]
+
+    if isinstance(child, Data):
+        if (
```
domfournier, line 86: Combine logic with `and`

tests/ui_json/uijson_test.py
```diff
@@ -122,3 +122,58 @@ class MyUIJson(BaseUIJson):
 
     assert "my_absent_uid_parameter" in str(err.value)
     assert "my_faulty_data_parameter" in str(err.value)
+
+
+def test_validation(tmp_path):
+    from geoh5py.objects import Curve
```
domfournier, line 128: Imports on top

tests/ui_json/uijson_test.py
```diff
@@ -122,3 +122,58 @@ class MyUIJson(BaseUIJson):
...
+    assert "my_data_parameter data is not a child of my_other_object_parameter" in str(
+        err.value
+    )
+    assert (
+        "Object's mesh type must be one of [<class 'geoh5py.objects.curve.Curve'>]"
+        in str(err.value)
+    )
+    assert "Dependency my_other_object_parameter must be either optional or" in str(
+        err.value
+    )
```
domfournier, line 179: Would you mind splitting this test into three? Will be easier to maintain and append later on. You could have some utility function that generates a UIJson class from a generic one, and just call it in each sub-test?

geoh5py/ui_json/validations/uijson.py
domfournier, line 1: What else is going to be added to this module? Could we just take over the existing validation file and eventually deprecate the InputValidation class?

domfournier (commented):
Very nice, I like!
See comments.

geoh5py/ui_json/validations/uijson.py
benk-mira: done.

tests/ui_json/uijson_test.py
benk-mira: whoops - done.

tests/ui_json/uijson_test.py
benk-mira: done.

geoh5py/ui_json/validations/uijson.py
benk-mira: I think we should store the form validations and update the forms to use the functional form of the field/model validators:
image

domfournier (approved):

PR #649, opened 2024-10-21
GEOPY-1802: Copying drillhole groups crashes

domfournier:
**GEOPY-1802 - Copying drillhole groups crashes**


benk-mira (approved):
Thanks man!

PR #651, opened 2024-10-23
GEOPY-1165: b

domfournier:
**GEOPY-1165 - Check for properties for Property Group Type - Implement Property Group Type: Simple**
Reverting to Enum

### DIFF
--- a/geoh5py/groups/property_group_type.py
+++ b/geoh5py/groups/property_group_type.py
@@ -45,8 +45,6 @@ class PropertyGroupType(ABC):
 
 
 class SimpleType(PropertyGroupType):
-    name: str = "Simple"  # Each subclass will define this
-
     @classmethod
     def verify(cls, children: list[Data]):
         """
@@ -55,8 +53,6 @@ class SimpleType(PropertyGroupType):
 
 
 class DepthType(PropertyGroupType):
-    name = "Depth table"
-
     @classmethod
     def verify(cls, children: list[Data]):
         if (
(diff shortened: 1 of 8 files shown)


geoh5py/groups/property_group_type.py
```diff
@@ -20,13 +20,14 @@
...
 from abc import ABC
+from enum import Enum
 
 from ..data import Data, DataAssociationEnum, FloatData, NumericData
 
 
 class PropertyGroupType(ABC):
     """
-    Class to define the basic structure of a property group type.
+    Abstract base class for defining the type of group that a property group is.
```
MatthieuCMira, line 30:
```suggestion
    Class to define the basic structure of a property group type.
```

geoh5py/groups/property_group_type.py
```diff
@@ -132,26 +133,48 @@ def verify(cls, children: list[Data]):
             )
 
 
-# mapper for group types
 GROUP_TYPES = {
```
MatthieuCMira, line 136: I have the feeling there is some repetition here betwen GROUP_TYPE, the objects of the Enum and the name

MatthieuCMira (dismissed):
As I did it, I am not against it. 
Just feel there is a bit of repetition in property_group_type.

Do you have an idea to make it simpler?

geoh5py/groups/property_group_type.py
domfournier: Right, we can probably remove the name from the classes....

MatthieuCMira:
I'm good with the changes

MatthieuCMira (approved):

PR #652, opened 2024-10-24
GEOPY-1811: Use latest "Tx ID tx property" on all EM surveys. 

domfournier:
**GEOPY-1811 - Use latest "Tx ID tx property" on ground EM metadata. **

### DIFF
--- a/geoh5py/objects/surveys/electromagnetics/base.py
+++ b/geoh5py/objects/surveys/electromagnetics/base.py
@@ -28,8 +28,7 @@ from warnings import warn
 
 import numpy as np
 
-from geoh5py.data import ReferencedData
-from geoh5py.data.float_data import FloatData
+from geoh5py.data import FloatData, IntegerData, ReferencedData
 from geoh5py.groups.property_group import PropertyGroup
 from geoh5py.objects import Curve
 from geoh5py.objects.object_base import ObjectBase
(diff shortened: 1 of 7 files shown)


geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -545,6 +567,38 @@ def _edit_validate_property_groups(
...
+        else:
+            uid = self.metadata["EM Dataset"].get("Tx ID tx property", None)
+
+        if uid is not None:
+            tx_id_property = self.get_entity(uid)[0]
+        else:
+            tx_id_property = self.get_entity("Transmitter ID")[0]
+            # Reset the metadata
+
+        if not isinstance(tx_id_property, ReferencedData | IntegerData | type(None)):
```
benk-mira, line 586: Seems like `type(none)` shouldn't be in the list.  You're asking for `getattr` on `tx_id_property` later

geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -805,7 +843,7 @@ def tx_id_property(self, value: uuid.UUID | ReferencedData | np.ndarray | None):
 
             value = self.add_data({"Transmitter ID": attributes})
 
-        if not isinstance(value, (ReferencedData, type(None))):
+        if not isinstance(value, (ReferencedData, IntegerData, type(None))):
```
benk-mira, line 846: same as above.

benk-mira (changes requested):
just a couple comments about allowing type(None), looks good otherwise.

geoh5py/objects/surveys/electromagnetics/base.py
domfournier: It has to be there since the get_entity can return None, which means that the tx ID is not defined on the object (possible since not mandatory)

benk-mira (approved):

PR #659, opened 2024-10-28
GEOPY-1623: metadata are not displaying when exporting metadata to drillhole group

domfournier:
**GEOPY-1623 - metadata are not displaying when exporting metadata to drillhole group**


MatthieuCMira (approved):
Nice and sweet

PR #663, opened 2024-11-08
GEOPY-1836: return referenced values in table

MatthieuCMira:
**GEOPY-1836 - return referenced values in table**


geoh5py/shared/concatenation/drillholes_group_table.py
```diff
@@ -58,25 +56,33 @@ def _create_empty_structured_array(
...
         names: tuple[str],
         drillhole: bool = True,
+        mapped: bool = False,
     ) -> np.ndarray:
         """
         Create an empty structured array that can contains the data.
 
         :param names: The names to extract.
         :param drillhole: If True, the drillholes are added to the table.
+        :param mapped: Map the referenced data back.
```
domfournier, line 66: Can you be more explicit about what the "data back" means? (i.e. Return the string 'value' instead of the key.

domfournier (approved):

PR #666, opened 2024-11-13
GEOPY-1840: Enforce reference data type shared between the tx_id_property of receiver and complement (transmitter, base station)

domfournier:
**GEOPY-1840 - Enforce reference data type shared between the tx_id_property of receiver and complement (transmitter, base station)**

### DIFF
--- a/geoh5py/objects/surveys/electromagnetics/base.py
+++ b/geoh5py/objects/surveys/electromagnetics/base.py
@@ -466,6 +466,11 @@ class BaseEMSurvey(ObjectBase, ABC):  # pylint: disable=too-many-public-methods
         if receivers.tx_id_property is not None:
             self.edit_em_metadata({"Tx ID property": receivers.tx_id_property.uid})
 
+            if isinstance(
+                self.tx_id_property, ReferencedData | IntegerData
+            ) and isinstance(receivers.tx_id_property, ReferencedData | IntegerData):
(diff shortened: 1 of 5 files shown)


geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -466,6 +466,11 @@ def receivers(self, receivers: BaseEMSurvey):
         if receivers.tx_id_property is not None:
             self.edit_em_metadata({"Tx ID property": receivers.tx_id_property.uid})
 
+            if isinstance(
+                self.tx_id_property, ReferencedData | IntegerData
+            ) and isinstance(receivers.tx_id_property, ReferencedData | IntegerData):
```
MatthieuCMira, line 471: Question: why when IntegerData?

MatthieuCMira (approved):
Seems good for me. 
Can you just explain a bit why we need to passe the entity_type for IntegerData?
Thanks

geoh5py/objects/surveys/electromagnetics/base.py
domfournier: Same as for any data, if they share type, they share visual parameters.

MatthieuCMira: Then why resricting it to IntergerData and ReferencedData only?

domfournier: Enforced by GA

PR #668, opened 2024-11-15
DEVOPS-571: add python packages in draft release, and publish to prod upon release

sebhmg:
**DEVOPS-571 - add python packages in draft release, and publish to prod upon release**


sebhmg:
tested as noted in https://github.com/MiraGeoscience/CI-tools/pull/45#issuecomment-2480049553

and also:
- auto-cancel when retagging with same tag
- auto cancel when rerunning manually for same release

PR #669, opened 2024-11-18
DEVOPS-559: Limit dependabot scan to production dependencies

SophieCurinier:
**DEVOPS-559 - Python packages: until vulnerability on runtime dependencies are mostly address, tell dependabot not to scan dev dependencies**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,5 @@ updates:
     schedule:
       interval: "monthly"
     target-branch: "develop"
+    allow:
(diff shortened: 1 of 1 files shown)


SophieCurinier:
What do you think @sebhmg ? This should be the workaround and only work on production dependencies.

sebhmg:
That is initially what I had in mind. But now, maybe it is enough having this rule dismissing low-impact alerts for dev dependencies:
image

This repo seems currently free from any dependabot alerts.

sebhmg:
no change needed for now

PR #670, opened 2024-11-26
GEOPY-1878: Error creating value map from values

domfournier:
**GEOPY-1878 - Error creating value map from values**

### DIFF
--- a/geoh5py/data/reference_value_map.py
+++ b/geoh5py/data/reference_value_map.py
@@ -63,7 +63,9 @@ class ReferenceValueMap(ABC):
 
         if isinstance(value_map, np.ndarray) and value_map.dtype.names is None:
             if value_map.ndim == 1:
-                value_map = {i: str(val) for i, val in enumerate(set(value_map))}
+                unique_set = set(value_map)
+                unique_set.discard(0)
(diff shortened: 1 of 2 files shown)


geoh5py/data/reference_value_map.py
```diff
@@ -63,7 +63,9 @@ def validate_value_map(cls, value_map: np.ndarray | dict) -> np.ndarray:
 
         if isinstance(value_map, np.ndarray) and value_map.dtype.names is None:
             if value_map.ndim == 1:
-                value_map = {i: str(val) for i, val in enumerate(set(value_map))}
+                unique_set = set(value_map)
+                unique_set.discard(0)
```
MatthieuCMira, line 67: why discard 0?
I guess that's the bug to correct because that's the only change. 

a label can't be named "0" in GA?

MatthieuCMira (approved):
seems ok to me. 

I understood waht you want to do. 
I'm just wondering why GA can't accept "0" as label.

geoh5py/data/reference_value_map.py
domfournier: 0 is reserved for "Unknown"

PR #671, opened 2024-11-27
GEOPY-1813: Change typing for input PropertyGroup.properties to list[Data] | list[UUID] | list[str] instead list[Data | UUID | str]

MatthieuCMira:
**GEOPY-1813 - Change typing for input PropertyGroup.properties to list[Data] | list[UUID] | list[str] instead list[Data | UUID | str]**
change the list type by Sequence as mypy don't like list

### DIFF
--- a/geoh5py/groups/property_group.py
+++ b/geoh5py/groups/property_group.py
@@ -68,7 +68,7 @@ class PropertyGroup:
         on_file: bool = False,
         uid: UUID | None = None,
         property_group_type: GroupTypeEnum | str = GroupTypeEnum.SIMPLE,
-        properties: list[UUID | Data | str] | None = None,
+        properties: Sequence[UUID | Data | str] | None = None,
         **_,
     ):
         self._parent: ObjectBase = self._validate_parent(parent)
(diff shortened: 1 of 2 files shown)


geoh5py/groups/property_group.py
```diff
@@ -68,7 +68,7 @@ def __init__(  # pylint: disable=too-many-arguments
         on_file: bool = False,
         uid: UUID | None = None,
         property_group_type: GroupTypeEnum | str = GroupTypeEnum.SIMPLE,
-        properties: list[UUID | Data | str] | None = None,
+        properties: Sequence[UUID | Data | str] | None = None,
```
domfournier, line 71: Changing to Sequence is better, but won't address the underlying issue of a mixed list

MatthieuCMira: With sequence, mypy does not complain for mix types.

There was not real underlying issue, just mypy complaining

domfournier (approved):
Yep that worked. Didn't know that mypy suggested exactly that fix. Thanks

PR #674, opened 2024-12-03
GEOPY-1881: Refactor desurvey method of drillholes using GA's implementation

domfournier:
**GEOPY-1881 - Refactor desurvey method of drillholes using GA's implementation**


domfournier:
@cmhowlett If you want to start testing GI desurveying, the methods are static, so you should be able to just call with arrays.

```
intervals = Drillhole.compute_intervals(survey: np.ndarray, collar, end_of_hole)
xyz = Drillhole.depths_to_xyz(intervals: dict, depths: np.ndarray)
```
You should get an exact match with GA's

image

cmhowlett (approved):
Nice!

geoh5py/objects/drillhole.py
```diff
@@ -733,60 +721,136 @@ def post_processing(self):
...
-    Compute the vertical deviation.
+    @staticmethod
+    def depths_to_xyz(intervals: dict, depths: np.ndarray) -> np.ndarray:
+        """
+        Convert survey parameters to x, y, z coordinates.
 
-    :param dip: Degree angle positive down from horizontal
+        :param intervals: Dictionary of intervals parameters.
+        :param depths: Array of depth values.
+        :param collar: Array of collar coordinates.
```
cmhowlett, line 844: collar isn't one of the params.

geoh5py/objects/drillhole.py
```diff
@@ -733,60 +721,136 @@ def post_processing(self):
...
+
+        # Densify along intervals if greater than 50 m
+        deltas = np.diff(full_survey[:, 0])
+        new_points = []
+        for ii, delta in enumerate(deltas):
+            if delta < MAXIMUM_DEPTH_INTERVAL:
+                continue
+
+            depths = np.arange(0, delta, MAXIMUM_DEPTH_INTERVAL)
+            d_azm = (full_survey[ii + 1, 1] - full_survey[ii, 1]) / delta
```
MatthieuCMira, line 760: ok, if I understand well every points at depth (depth or interval) has its dip, and azimuth, corresponding to the deviation?

geoh5py/objects/drillhole.py
```diff
@@ -733,60 +721,136 @@ def post_processing(self):
                     self._cells = key_map.reshape((-1, 2)).astype("uint32")
                     self.workspace.update_attribute(self, "cells")
```
MatthieuCMira, line 723: This function is doing much more than just validating no?
It's creating new points isin't it?

geoh5py/objects/drillhole.py
```diff
@@ -733,60 +721,136 @@ def post_processing(self):
...
+            ],
+        }
 
+        return intervals
 
-def deviation_z(_, dip):
-    """
-    Compute the vertical deviation.
+    @staticmethod
+    def depths_to_xyz(intervals: dict, depths: np.ndarray) -> np.ndarray:
```
MatthieuCMira, line 838: why intervals is not loaded from self? (why static?)

the idea of depth_to_xyz is not to do a desurvey?

like drillholes.desurvey(depth_values) => xyz coords?

MatthieuCMira (commented):
i have some question about the structure. 

I have the feeling a lot of static method are used to get values contains on the drillhole. 
Like your example :
```
intervals = Drillhole.compute_intervals(survey: np.ndarray, collar, end_of_hole)
xyz = Drillhole.depths_to_xyz(intervals: dict, depths: np.ndarray)
```

why interval is not a property of the drillhole?
or loaded from a property group?
The `"survey"`of the first functions are the values associated to the Depthg or Interval group isn't it?

geoh5py/objects/drillhole.py
domfournier: The "intervals" are independent of data, just based on geometry (collar and survey). Once we have intervals defined, then we can compute xyz from depths at any location.

geoh5py/objects/drillhole.py
domfournier: The reason being that the methods can be used outside of an instance of Drillhole, like what the GI team needs.

geoh5py/objects/drillhole.py
domfournier: Sure, I can rename.

geoh5py/objects/drillhole.py
domfournier: Nice catch

domfournier:
> 
> why interval is not a property of the drillhole? or loaded from a property group? The `"survey"`of the first functions are the values associated to the Depthg or Interval group isn't it?

Intervals are stored as property of the Drillhole. This is only done if you have an instance of the Drillhole. Otherwise you can just call the methods from arrays.

geoh5py/objects/drillhole.py
MatthieuCMira: Ok.
We don't have "deviation" along a drillhole in geoh5py right?

PR #675, opened 2024-12-04
GEOPY-1851: fail to clip geoh54 with error about CustomGroup

domfournier:
**GEOPY-1851 - fail to clip geoh54 with error about CustomGroup**


MatthieuCMira (approved):
Ok your changes look good

PR #679, opened 2024-12-12
GEOPY-1910: Throw warning if any of the common auto-saving drive is found in mode="r+"

benk-mira:
**GEOPY-1910 - Throw warning if any of the common auto-saving drive is found in mode="r+"**

### DIFF
--- a/geoh5py/workspace/workspace.py
+++ b/geoh5py/workspace/workspace.py
@@ -73,6 +73,16 @@ from ..shared.utils import (
 )
 
 
+NETWORK_DRIVES = [
+    "Egnyte Connect",
+    "One Drive",
+    "Dropbox",
+    "Google Drive",
+    "Box",
+    "iCloud",
+]
+
+
 # pylint: disable=too-many-instance-attributes
 class Workspace(AbstractContextManager):
     """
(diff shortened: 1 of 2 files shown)


domfournier (dismissed):

geoh5py/workspace/workspace.py
```diff
@@ -1236,6 +1246,13 @@ def open(self, mode: str | None = None) -> Workspace:
         if mode is None:
             mode = self._mode
 
+        if mode == "r+" and any(k in str(self.h5file) for k in NETWORK_DRIVES):
```
domfournier, line 1249:
```suggestion
        if mode == "r+" and any(k in str(self.h5file.absolute()) for k in NETWORK_DRIVES):
```
Need to also make sure h5file is a `str` of `Path`.

geoh5py/workspace/workspace.py
```diff
@@ -73,6 +73,16 @@
 )
 
 
+NETWORK_DRIVES = [
+    "Egnyte Connect",
+    "One Drive",
+    "Dropbox",
+    "Google Drive",
```
domfournier, line 80: Mine is `googledrive`

benk-mira: I think you must have updated it when installing, or they changed their default.  The current default is "Google Drive"

domfournier: It says Google Drive in the explorer, but the windows path is lower case. 
image

Anyway, doesn't cost much to add it to the list.

benk-mira: ok, added

domfournier (approved):

PR #684, opened 2025-01-07
GEOPY-1918: Don't load workspace_geoh5 by default.

domfournier:
**GEOPY-1918 - Don't load workspace_geoh5 by default.**


geoh5py/ui_json/utils.py
```diff
@@ -303,6 +303,12 @@ def path2workspace(value):
     return value
 
 
+def str2path(value):
```
benk-mira, line 306: Maybe a good idea to keep this in shared/utils with the other method.

benk-mira (changes requested):

geoh5py/shared/utils.py
```diff
@@ -913,3 +919,9 @@ def remove_duplicates_in_list(input_list: list) -> list:
     :return: The sorted list
     """
     return sorted(set(input_list), key=input_list.index)
+
+
+def str2path(value):
+    if isinstance(value, str) and Path(value).exists():
```
sebhmg, line 925: @domfournier, does the Path has to exist on disk? could it be an output path?

I do not know if there is a way to test if path is valid while not requesting for it to exists

PR #688, opened 2025-01-15
GEOPY-1932: Desurveying can produce a divide by zero warning with some paths

domfournier:
**GEOPY-1932 - Desurveying can produce a divide by zero warning with some paths**


MatthieuCMira (approved):
Code seems ok, go to testing

PR #701, opened 2025-02-25
GEOPY-2002: Add class method to dump UIJson class to defaulted ui.json file

benk-mira:
**GEOPY-2002 - Add class method to dump UIJson class to defaulted ui.json file**


domfournier (changes requested):
Moving similar mechanics to simpeg-drivers for now.

PR #702, opened 2025-02-26
GEOPY-2033: Error copying reference data with GeometryDataValueMap

domfournier:
**GEOPY-2033 - Error copying reference data with GeometryDataValueMap**

### DIFF
--- a/geoh5py/data/referenced_data.py
+++ b/geoh5py/data/referenced_data.py
@@ -52,7 +52,8 @@ class ReferencedData(IntegerData):
         **kwargs,
     ) -> Data:
         """
-        Overwrite the copy method to ensure that the uid is set to None.
+        Overwrite the copy method to ensure that the uid is set to None and
+        transfer the GeometricDataValueMapType to the new entity type.
         """
         kwargs["uid"] = None
 
--- a/geoh5py/data/referenced_data.py
+++ b/geoh5py/data/referenced_data.py
@@ -197,15 +197,17 @@ class ReferencedData(IntegerData):
         if not isinstance(data, dict | np.ndarray):
             raise TypeError("Data map values must be a numpy array or dict")
 
-        reference_data = ReferenceValueMap(data, name=name)
-
         if self.entity_type.value_map is None:
             raise ValueError("Entity type must have a value map.")
 
-        if not set(reference_data.map["Key"]).issubset(
-            set(self.entity_type.value_map.map["Key"])
-        ):
-            raise KeyError("Data map keys must be a subset of the value map keys.")
+        reference_data = ReferenceValueMap(data, name=name)
+        # TODO: Enforce that the keys of the data map are a subset
+        #  of the value map keys once GA changes its behavior
+        # if not set(reference_data.map["Key"]).issubset(
+        #     set(self.entity_type.value_map.map["Key"])
+        # ):
+        #     raise KeyError("Data map keys must be a subset of the value map keys.")
+        #
 
         data_type = GeometricDataValueMapType(
             self.workspace,
--- a/geoh5py/data/referenced_data.py
+++ b/geoh5py/data/referenced_data.py
@@ -24,8 +24,9 @@ from typing import cast
 
 import numpy as np
 
+from ..shared.utils import find_unique_name
 from .data import Data
-from .data_type import GeometricDataValueMapType, ReferenceDataType
+from .data_type import DataType, GeometricDataValueMapType, ReferenceDataType
 from .geometric_data import GeometricDataConstants
 from .integer_data import IntegerData
 from .primitive_type_enum import PrimitiveTypeEnum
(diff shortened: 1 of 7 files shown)


geoh5py/data/referenced_data.py
```diff
@@ -39,6 +43,42 @@ def __init__(self, **kwargs):
...
+        clear_cache: bool = False,
+        mask: np.ndarray | None = None,
+        **kwargs,
+    ) -> Data:
+        """
+        Overwrite the copy method to ensure that the uid is set to None.
+        """
+        kwargs["uid"] = None
+
+        new_data = cast(
```
MatthieuCMira, line 59: I'm not against cast, but is there more "cleaned" way to ensure the type here?

geoh5py/data/referenced_data.py
```diff
@@ -39,6 +43,42 @@ def __init__(self, **kwargs):
...
+        )
+
+        # Always overwrite the entity type name to protect the GeometricDataValueMapType
+        new_data.entity_type.name = find_unique_name(
+            self.entity_type.name,
+            [tp.name for tp in self.workspace.types if isinstance(tp, DataType)],
+        )
+
+        if self.data_maps is not None:
+            for name, child in self.data_maps.items():
```
MatthieuCMira, line 77: data_maps has a setter. You could pass directly 
`new_data.data_maps=self.data_maps` no?

geoh5py/objects/object_base.py
```diff
@@ -309,7 +310,7 @@ def copy(
         if copy_children:
             children_map = {}
             for child in self.children:
-                if isinstance(child, PropertyGroup):
+                if isinstance(child, PropertyGroup | GeometricDataConstants):
```
MatthieuCMira, line 313: Can you explain what's that's `GeometricDataConstants` for my understanding? I understand it's data, but does it behave like PropertyGroup?

geoh5py/data/referenced_data.py
```diff
@@ -39,6 +43,42 @@ def __init__(self, **kwargs):
...
+    def copy(
+        self,
+        parent=None,
+        *,
+        clear_cache: bool = False,
+        mask: np.ndarray | None = None,
+        **kwargs,
+    ) -> Data:
+        """
+        Overwrite the copy method to ensure that the uid is set to None.
```
MatthieuCMira, line 55: "And transfer the data_maps unique to reference data."

geoh5py/workspace/workspace.py
```diff
@@ -283,7 +283,7 @@ def copy_to_parent(
 
         entity_type_kwargs = get_attributes(
             entity.entity_type,
```
MatthieuCMira, line 285: won't this break other process somewhere?

tests/reference_data_test.py
```diff
@@ -179,6 +176,25 @@ def test_add_data_map(tmp_path):
...
+            data.entity_type.value_map.map["Key"],
+            np.random.randn(len(data.entity_type.value_map.map["Key"])),
+        ]
+        data.add_data_map("test2", data_map)
+
+        data.parent.copy()
+        geom_data = workspace.get_entity("test2")
+        assert len(geom_data) == 2
+
+        assert geom_data[0].entity_type.name != geom_data[1].entity_type.name
```
MatthieuCMira, line 195: add another assert to check data_map are well transfered

MatthieuCMira (commented):
Nothing major. 

Except for the transfer of data_map, was there other issues covered by these changes?

geoh5py/data/referenced_data.py
domfournier: We would need to remove the typing on the base class, and type the sub-classes instead.
To be addressed on https://mirageoscience.atlassian.net/browse/GEOPY-2038

geoh5py/data/referenced_data.py
domfournier: The data map setter accepts a dictionary of GeometricData. There are more mechanics involved with the `add_data_map`, such as creating the GeometricData and its data type, which needs to be linked to the reference data. The linkage between the ReferenceData and the GeometricData is pretty complicated on geoh5, and I raised the issue to @krisdavis. We can simplify it once we have a stronger link.

geoh5py/objects/object_base.py
domfournier: It is specific to these
image

Either the locations (created by GA) or the functions. Their respective DataType defines the behaviour.

geoh5py/data/referenced_data.py
MatthieuCMira: ok, maybe you could create an issue then? 
And precise to change the loops when it's ready?

geoh5py/objects/object_base.py
MatthieuCMira: ok I got it

MatthieuCMira (dismissed):
ok for me.

PR #705, opened 2025-02-28
Fixes to make simpeg-drivers uijson round-trip test pass

benk-mira:


geoh5py/shared/validators.py
```diff
@@ -105,10 +105,31 @@ def to_class(
...
+    """None transforms to empty string for serialization."""
+    if value is None:
+        return ""
+    return value
+
+
+def types_to_string(types: list) -> list[str] | str:
+    if len(types) > 1:
+        return [f"{{{k.default_type_uid()}}}" for k in types]
+    return f"{{{types[0].default_type_uid()}}}"
```
domfournier, line 125:
```suggestion
        return [str(k.default_type_uid()) for k in types]
    return str(types[0].default_type_uid())
```
Why so many braces?

geoh5py/shared/validators.py
```diff
@@ -105,10 +105,31 @@ def to_class(
...
+
+
+def uuid_to_string(value: UUID | None) -> str:
+    if value is None:
+        return ""
+    if isinstance(value, UUID):
+        return f"{{{value!s}}}"
```
domfournier, line 132:
```suggestion
        return str(value)
```

geoh5py/shared/validators.py
benk-mira: It's to make the uuid string like "{blahblahblah}"

benk-mira: I don't mind breaking the mold, but it's how we've formatted the uid's up to now.

domfournier (approved):

PR #706, opened 2025-03-03
GEOPY-1987: Error from cmd.exe printed if h5repack is not found

domfournier:
**GEOPY-1987 - Error from cmd.exe printed if h5repack is not found**

### DIFF
--- a/geoh5py/workspace/workspace.py
+++ b/geoh5py/workspace/workspace.py
@@ -32,6 +32,7 @@ from contextlib import AbstractContextManager, contextmanager
 from gc import collect
 from getpass import getuser
 from io import BytesIO
+from logging import getLogger
 from pathlib import Path
 from shutil import copy, move
 from subprocess import CalledProcessError
@@ -75,6 +76,8 @@ from ..shared.utils import (
     str2uuid,
 )
 
+
+logger = getLogger(__name__)
 
 NETWORK_DRIVES = [
     "Egnyte Connect",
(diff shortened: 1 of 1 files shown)


geoh5py/workspace/workspace.py
```diff
@@ -225,6 +225,7 @@ def close(self):
                     check=True,
                     shell=True,
                     stdout=subprocess.DEVNULL,
+                    stderr=subprocess.DEVNULL,
```
sebhmg, line 228: **issue:** error should not be ignored when `CalledProcessError` is ignore (with `pass`)

Suggestion:
- capture output, and report it in exception handling
- ignore `FileNotFoundError` which happens when program is not found (without `shell=True`)

```py
            try:
                subprocess.run(
                    ["h5repack", "--native", str(self._h5file), str(temp_file)],
                    check=True,
                    capture_output=True,
                    encoding="utf8",
                )
                # unlink, move ...
            except FileNotFoundError:
                # h5repack not installed or not found
                pass
            except CalledProcessError as process_error:
                logging.getLogger().error(f"{process_error.stdout}\n{process_error.stderr}")
```

geoh5py/workspace/workspace.py
```diff
@@ -225,6 +225,7 @@ def close(self):
                     check=True,
                     shell=True,
```
sebhmg, line 226: preferably, avoid using `Shell=True`. See [subprocess.html#security-consideration](https://docs.python.org/3/library/subprocess.html#security-considerations)).
Code be changed to:
```py
                subprocess.run(
                    ["h5repack", "--native", str(self._h5file), str(temp_file)],
                    check=True,
```
(`shell=False` is the default)

sebhmg (changes requested):
Should not dismiss error when h5repack is present and fails.
See some suggestion.

geoh5py/workspace/workspace.py
domfournier: Seems to work.

geoh5py/workspace/workspace.py
```diff
@@ -221,15 +224,15 @@ def close(self):
...
-                    f'h5repack --native "{self._h5file}" "{temp_file}"',
+                    ["h5repack", "--native", str(self._h5file), str(temp_file)],
                     check=True,
-                    shell=True,
-                    stdout=subprocess.DEVNULL,
+                    capture_output=True,
+                    encoding="utf8",
                 )
-                Path(self._h5file).unlink()
-                move(temp_file, self._h5file, copy)
```
sebhmg, line 230: **issue:** still need to replace initial geoh5 file by repacked tempfile

sebhmg (changes requested):
missing the final move of repacked file

geoh5py/workspace/workspace.py
domfournier: nice catch

sebhmg (approved):

PR #710, opened 2025-03-07
GEOPY-2039: Deal with data values vector of wrong length

MatthieuCMira:
**GEOPY-2039 - Deal with data values vector of wrong length**
avoid raising a bug if values are too long and truncate them like GA

### DIFF
--- a/geoh5py/data/numeric_data.py
+++ b/geoh5py/data/numeric_data.py
@@ -74,7 +74,12 @@ class NumericData(Data, ABC):
         if self.n_values is None:
             return values
 
+        error_message = (
+            f"Input 'values' of shape ({self.n_values},) expected. "
+            f"Array of shape{values.shape} provided for data {self.name}."
+        )
         if len(values) < self.n_values:
+            warn(error_message)
             full_vector = np.ones(self.n_values, dtype=values.dtype) * self.nan_value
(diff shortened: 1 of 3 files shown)


geoh5py/data/numeric_data.py
```diff
@@ -74,7 +74,12 @@ def format_length(self, values: np.ndarray) -> np.ndarray:
         if self.n_values is None:
             return values
 
+        error_message = (
+            f"Input 'values' of shape ({self.n_values},) expected. "
+            f"Array of shape{values.shape} provided for data {self.name}."
+        )
         if len(values) < self.n_values:
+            warn(error_message)
```
domfournier, line 82: That will happen very often and expected behaviour, let's not warn here.

geoh5py/data/numeric_data.py
```diff
@@ -83,10 +88,8 @@ def format_length(self, values: np.ndarray) -> np.ndarray:
             len(values) > self.n_values
             and self.association is not DataAssociationEnum.OBJECT
         ):
-            raise ValueError(
-                f"Input 'values' of shape({self.n_values},) expected. "
-                f"Array of shape{values.shape} provided.)"
-            )
+            warn(error_message)
```
domfournier, line 91:
```suggestion
            logger.warning(error_message)
```

domfournier (approved):

PR #711, opened 2025-03-10
GEOPY-2065: Crash on monitoring directory copy with reference data

domfournier:
**GEOPY-2065 - Crash on monitoring directory copy with reference data**


MatthieuCMira (approved):
I'll do the test right now, wait before merging

PR #713, opened 2025-03-12
GEOPY-2056: Handle extra fields and deprecations in UIJson version validation

benk-mira:
**GEOPY-2056 - Handle extra fields and deprecations in UIJson version validation**
…ther packages


domfournier (commented):
You are not using it anywhere? Should probably add a unitest for it.

PR #717, opened 2025-03-31
GEOPY-2053: Support colour data type

MatthieuCMira:
**GEOPY-2053 - Support colour data type**
add colour data.

(PS: I corrected a bug in property_group_table.py along the way)


geoh5py/data/colour.py
```diff
@@ -0,0 +1,183 @@
...
+        if values.ndim != 2 or values.shape[1] not in (3, 4):
+            raise ValueError("Values must be a 2D numpy array containing RGB bands.")
+
+        if values.dtype != np.uint8:
+            values = min_max_scaler(values, 0, 255, axis=0)
+
+        values = (
+            values[:, :3]
+            .astype(np.uint8)
+            .view([("r", "u1"), ("g", "u1"), ("b", "u1")])
```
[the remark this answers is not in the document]
MatthieuCMira: I talk with Ben, and he told me the alpha is not supported by analyst. 

They are just stacking 255 values for this channel.

MatthieuCMira:
image

geoh5py/data/colour.py
```diff
@@ -0,0 +1,172 @@
...
+            values = min_max_scaler(values, 0, 255, axis=0)
+
+        if values.shape[1] == 3:
+            values = np.column_stack(
+                (values, np.full(values.shape[0], 255, dtype=np.uint8))
+            )
+
+        values = (
+            values[:, :4]
+            .astype(np.uint8)
```
domfournier, line 79: The conversion to int is an issue is the array contains nan from the NDV of GA. We should augment the tests to cover the case.

geoh5py/data/colour.py
```diff
@@ -0,0 +1,172 @@
...
+        if values.dtype != np.uint8:
+            values = min_max_scaler(values, 0, 255, axis=0)
+
+        if values.shape[1] == 3:
+            values = np.column_stack(
+                (values, np.full(values.shape[0], 255, dtype=np.uint8))
+            )
+
+        values = (
+            values[:, :4]
```
domfournier, line 78: Shouldn't have to slice since already validating the shape above.

geoh5py/data/colour.py
```diff
@@ -0,0 +1,172 @@
...
+            .astype(np.uint8)
+            .view([("r", "u1"), ("g", "u1"), ("b", "u1"), ("a", "u1")])
+            .reshape(-1)
```
domfournier, line 81: That's an interesting way to do it. I would vote that we keep things explicit like other places, so we can easily refactor later.
```
            values = np.core.records.fromarrays(
                values.T, names=self._names, formats=self._formats
            )
```

geoh5py/data/colour.py
```diff
@@ -0,0 +1,172 @@
...
+                    f"The allowed parents are {cls.allowed_parent}"
+                )
+
+    @property
+    def nan_value(self):
+        return np.array(
+            [
+                (90, 90, 90, 0),
+            ],
+            dtype=[("r", "u1"), ("g", "u1"), ("b", "u1"), ("a", "u1")],
```
domfournier, line 146: Can we store that information as class attribute, like is done elsewhere such as: https://github.com/MiraGeoscience/geoh5py/blob/07722a173159bb412c664e07d4e793743adaec81/geoh5py/data/color_map.py#L40

geoh5py/data/colour.py
MatthieuCMira: Sure. 
I'll remove the nan values by 90 if in rgb, by 0 if in a,  as no data values is  90, 90, 90, 0

geoh5py/data/colour.py
MatthieuCMira: view is soooo much more efficient....

Are you sure you want to use your method??
`
import timeit
import tracemalloc
# Constants
N = 10_000_000
dtype = np.uint8
# Generate a big RGBA array (N, 4) of uint8 values
raw = np.random.randint(0, 256, size=(N, 4), dtype=dtype)
# Conversion using .view()
def method_view():
    return raw.view([("r", "u1"), ("g", "u1"), ("b", "u1"), ("a", "u1")]).reshape(-1)
# Conversion using np.core.records.fromarrays
def method_records():
    return np.core.records.fromarrays(
        [raw[:, i] for i in range(4)],
        names=["r", "g", "b", "a"],
        formats=["u1", "u1", "u1", "u1"]
    )
# Benchmark time
print("Timing (10 runs each):")
print("view():    ", timeit.timeit(method_view, number=10))
print("records(): ", timeit.timeit(method_records, number=10))
# Benchmark memory
def measure_memory(fn):
    tracemalloc.start()
    _ = fn()
    current, peak = tracemalloc.get_traced_memory()
    tracemalloc.stop()
    return current / 1024 / 1024, peak / 1024 / 1024  # MB
print("\nMemory usage (MB):")
curr, peak = measure_memory(method_view)
print(f"view():     current = {curr:.2f}, peak = {peak:.2f}")
curr, peak = measure_memory(method_records)
print(f"records():  current = {curr:.2f}, peak = {peak:.2f}")
`

results:
`
Timing (10 runs each):
view():     4.579999949783087e-05
records():  0.4267208999954164
Memory usage (MB):
view():     current = 0.00, peak = 0.00
records():  current = 38.15, peak = 38.15
`

MatthieuCMira: I'll definitely vote to change the other usages...

domfournier: To be addressed on GEOPY-2112

domfournier (approved):

PR #724, opened 2025-04-10
GEOPY-1900: Implement Text Data object

domfournier:
**GEOPY-1900 - Implement Text Data object**

### DIFF
--- a/geoh5py/objects/text.py
+++ b/geoh5py/objects/text.py
@@ -64,8 +64,10 @@ class TextEntry(BaseModel):
         alias_generator=AliasGenerator(  # type: ignore
             validation_alias=to_snake,
             serialization_alias=to_pascal,
-        )
+        ),
+        validate_assignment=True,
     )
+
     starting_point: str
     annotation: int = 0
     color: str = "#ffffff"
@@ -85,6 +87,9 @@ class TextEntry(BaseModel):
         """
         Convert a list of floats to a string representation.
         """
+        if isinstance(float_list, tuple) and len(float_list) == 1:
+            float_list = float_list[0]
+
         if isinstance(float_list, np.ndarray | list):
             float_list = "{" + ",".join([str(val) for val in float_list]) + "}"
 
@@ -201,7 +206,7 @@ class TextObject(Points):
             value = self._validate_value_length(value)
 
             for entry, elem in zip(self.text_mesh_data.text_data, value, strict=False):
-                entry.__setattr__(key, elem)
+                setattr(entry, key, elem)
 
             if self.on_file:
                 self.workspace.update_attribute(self, "attributes")
--- a/geoh5py/objects/text.py
+++ b/geoh5py/objects/text.py
@@ -1,0 +1,225 @@
+# ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+#  Copyright (c) 2025 Mira Geoscience Ltd.                                     '
+#                                                                              '
+#  This file is part of geoh5py.                                               '
+#                                                                              '
+#  geoh5py is free software: you can redistribute it and/or modify             '
+#  it under the terms of the GNU Lesser General Public License as published by '
+#  the Free Software Foundation, either version 3 of the License, or           '
+#  (at your option) any later version.                                         '
+#                                                                              '
+#  geoh5py is distributed in the hope that it will be useful,                  '
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of              '
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the               '
+#  GNU Lesser General Public License for more details.                         '
+#                                                                              '
+#  You should have received a copy of the GNU Lesser General Public License    '
+#  along with geoh5py.  If not, see <https://www.gnu.org/licenses/>.           '
+# ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+
+from __future__ import annotations
+
+import uuid
+from json import loads
+from typing import TYPE_CHECKING
+
+import numpy as np
+from pydantic import AliasChoices, BaseModel, Field, field_validator
+
+from .points import Points
+
+
+if TYPE_CHECKING:
+    from numpy import ndarray
+
+
+class TextEntry(BaseModel):
+    """
+    Core parameters for text mesh data.
+    """
+
+    starting_point: str = Field(
+        validation_alias=AliasChoices("starting_point", "StartingPoint"),
+        serialization_alias="StartingPoint",
+    )
+    annotation: int = Field(
+        validation_alias=AliasChoices("annotation", "Annotation"),
+        serialization_alias="Annotation",
+        default=0,
+    )
+    color: str | None = Field(
+        validation_alias=AliasChoices("color", "Color"),
+        serialization_alias="Color",
+        default="#ffffff",
+    )
+    direction: str | None = Field(
+        validation_alias=AliasChoices("direction", "Direction"),
+        serialization_alias="Direction",
+        default="{1,0,0}",
+    )
+    font: str | None = Field(
+        validation_alias=AliasChoices("font", "Font"),
+        serialization_alias="Font",
+        default="builtin_simple",
+    )
+    font_size: int | None = Field(
+        validation_alias=AliasChoices("font_size", "FontSize"),
+        serialization_alias="FontSize",
+        default=10,
+    )
+    layer_name: str | None = Field(
+        validation_alias=AliasChoices("layer_name", "LayerName"),
+        serialization_alias="LayerName",
+        default="",
+    )
+    normal: str | None = Field(
+        validation_alias=AliasChoices("normal", "Normal"),
+        serialization_alias="Normal",
+        default="{0,0,1}",
+    )
+    obliquing_angle: int | None = Field(
+        validation_alias=AliasChoices("obliquing_angle", "ObliquingAngle"),
+        serialization_alias="ObliquingAngle",
+        default=0,
+    )
+    rotation_angle: int | None = Field(
+        validation_alias=AliasChoices("rotation_angle", "RotationAngle"),
+        serialization_alias="RotationAngle",
+        default=0,
+    )
+    text: str | None = Field(
+        validation_alias=AliasChoices("text", "Text"),
+        serialization_alias="Text",
+        default="<no-data>",
+    )
+    width_scale_factor: float | None = Field(
+        validation_alias=AliasChoices("width_scale_factor", "WidthScaleFactor"),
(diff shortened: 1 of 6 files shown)


geoh5py/objects/text.py
```diff
@@ -0,0 +1,225 @@
...
+if TYPE_CHECKING:
+    from numpy import ndarray
+
+
+class TextEntry(BaseModel):
+    """
+    Core parameters for text mesh data.
+    """
+
+    starting_point: str = Field(
```
benk-mira, line 42: Pydantic has an alias generator to handle this kind of aliasing.  From our forms class:
image

geoh5py/objects/text.py
```diff
@@ -0,0 +1,225 @@
...
+
+class TextObject(Points):
+    """
+    Text object for annotated text labels in viewport.
+    """
+
+    _attribute_map: dict = Points._attribute_map.copy()
+    _attribute_map.update({"TextMesh Data": "text_mesh_data"})
+    _default_name = "Text Object"
+    _TYPE_UID: uuid.UUID | None = uuid.UUID("{c00905d1-bc3b-4d12-9f93-07fcf1450270}")
```
benk-mira, line 130: why is there an `| None` in the annotation?

geoh5py/objects/text.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        Geography bounding box of the object.
+        """
+        return None
+
+    def mask_by_extent(self, extent: ndarray, inverse: bool = False) -> None:
+        """
+        Sub-class extension of :func:`~geoh5py.shared.entity.Entity.mask_by_extent`.
+        """
+        return None
```
benk-mira, line 169: Should this just return an array of all true?

benk-mira: I guess same comment goes for the base class behaviour:
image

benk-mira: Shouldn't have to check if the result is None everywhere before we use the mask.

benk-mira (changes requested):
Looks good.  A couple of suggestions

geoh5py/objects/text.py
domfournier: Yep, brilliant

geoh5py/objects/text.py
domfournier: Bad copy I guess. Removing

geoh5py/objects/text.py
domfournier: The goal of `mask_by_extent` was to have a generic method for all entities, even those that don't have coordinates (like labels and groups). If a group does have objects, then its `mask_by_extent` will return the combine extent of everything, otherwise it won't be copied. 

By setting to None here I am saying that this object can't be copied by extent (until we write the mechanics to update the locations in the text_data at the same time as vertices).

geoh5py/objects/text.py
```diff
@@ -0,0 +1,192 @@
...
+from pydantic.alias_generators import to_pascal, to_snake
+
+from .points import Points
+
+
+if TYPE_CHECKING:
+    from numpy import ndarray
+
+
+class TextEntry(BaseModel):
```
MatthieuCMira, line 45: add the docstrings

geoh5py/objects/text.py
```diff
@@ -0,0 +1,192 @@
...
+    layer_name: str = ""
+    normal: str = "{0,0,1}"
+    obliquing_angle: int = 0
+    rotation_angle: int = 0
+    text: str = "<no-data>"
+    width_scale_factor: float = 1.0
+
+    @field_validator("starting_point", "normal", "direction", mode="before")
+    @classmethod
+    def array_to_str(cls, float_list: np.ndarray | list):
```
MatthieuCMira, line 71: output type + docstring

geoh5py/objects/text.py
```diff
@@ -0,0 +1,192 @@
...
+        if isinstance(float_list, np.ndarray | list):
+            float_list = "{" + ",".join([str(val) for val in float_list]) + "}"
+
+        return float_list
+
+
+class TextData(BaseModel):
+    """
+    Text mesh data.
```
MatthieuCMira, line 80: docstrings

geoh5py/objects/text.py
```diff
@@ -0,0 +1,192 @@
...
+
+    _attribute_map: dict = Points._attribute_map.copy()
+    _attribute_map.update({"TextMesh Data": "text_mesh_data"})
+    _default_name = "Text Object"
+    _TYPE_UID: uuid.UUID = uuid.UUID("{c00905d1-bc3b-4d12-9f93-07fcf1450270}")
+
+    def __init__(
+        self,
+        vertices: np.ndarray | list | tuple | None = None,
+        text_mesh_data: str | None = None,
```
MatthieuCMira, line 102: add docstrings for new attributes

geoh5py/objects/text.py
```diff
@@ -0,0 +1,192 @@
...
+
+class TextObject(Points):
+    """
+    Text object for annotated text labels in viewport.
+    """
+
+    _attribute_map: dict = Points._attribute_map.copy()
+    _attribute_map.update({"TextMesh Data": "text_mesh_data"})
+    _default_name = "Text Object"
+    _TYPE_UID: uuid.UUID = uuid.UUID("{c00905d1-bc3b-4d12-9f93-07fcf1450270}")
```
MatthieuCMira, line 97: _Type_uid has different implementation for different class:

eg:

in geo_image: 
```
    _TYPE_UID = uuid.UUID(
        fields=(0x77AC043C, 0xFE8D, 0x4D14, 0x81, 0x67, 0x75E300FB835A)
    )
```

would be nice to make it the same in every classes

geoh5py/objects/text.py
```diff
@@ -0,0 +1,192 @@
...
+
+    @text_mesh_data.setter
+    def text_mesh_data(self, value: str | None):
+        """
+        Set the text mesh data.
+        """
+
+        self._text_mesh_data = self._validate_text_data(value)
+
+    @property
```
MatthieuCMira, line 125: dfo some trivial test for extent is None (vallable for all None returning functions)

geoh5py/objects/text.py
```diff
@@ -0,0 +1,192 @@
...
+        return None
+
+    def mask_by_extent(self, extent: ndarray, inverse: bool = False) -> None:
+        """
+        Sub-class extension of :func:`~geoh5py.shared.entity.Entity.mask_by_extent`.
+        """
+        return None
+
+    def _validate_text_data(self, value: str | dict | None) -> TextData:
+        """
```
MatthieuCMira, line 139: docstrings

geoh5py/objects/text.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def __init__(
+        self,
+        vertices: np.ndarray | list | tuple | None = None,
+        text_mesh_data: str | None = None,
+        **kwargs,
+    ):
+        self._text_mesh_data: TextData
+
+        super().__init__(vertices=vertices, text_mesh_data=text_mesh_data, **kwargs)
```
MatthieuCMira, line 107: I guess the text mesh data is using the map atrribute function

I am not against, but won't it be more transparent to call it in the init, after the super()

self.text_mesh_data = text_mesh_data ?

geoh5py/objects/text.py
```diff
@@ -0,0 +1,192 @@
...
+        """
+        return None
+
+    def mask_by_extent(self, extent: ndarray, inverse: bool = False) -> None:
+        """
+        Sub-class extension of :func:`~geoh5py.shared.entity.Entity.mask_by_extent`.
+        """
+        return None
+
+    def _validate_text_data(self, value: str | dict | None) -> TextData:
```
MatthieuCMira, line 138: is validate text data used somewhere else thant the setter?

If not, the code could be in the setter

geoh5py/objects/text.py
```diff
@@ -0,0 +1,192 @@
...
+        return mesh_data
+
+    def _validate_value_length(self, values) -> list:
+        """
+        Validate the length of the values.
+        """
+        if not isinstance(values, (list, np.ndarray)):
+            values = [values] * self.n_vertices
+
+        if len(values) != self.n_vertices:
```
MatthieuCMira, line 172: Just an opened question:

why text data is dependant on points?
Couldn't be associated with another type of object?

tests/text_object_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+        assert isinstance(text.text_mesh_data, TextData)
+        for entry, label in zip(
+            text.text_mesh_data.text_data, random_labels, strict=False
+        ):
+            assert entry.text == label
+            assert entry.color == color
+
+
+def test_text_data_length_mismatch(tmp_path):
```
MatthieuCMira, line 62: could you create another function to tests the errors and the trivial mechanisms for covergae please?

MatthieuCMira (changes requested):
Sound like an object I could use for petro-lingo ;-)

Can you explain to me the global idea of the object please. I am  a bit confused with the nuance with a TextData, etc. 

The code seems good. Minor optional comments.

Globally, docstrings and some typing are missing, I might not have comment all. 
Augment the text to ensure the errors/None return values are working as intended (let's keep an high coverage from the start!)

Good job!

geoh5py/objects/text.py
domfournier: I would like to be able to let the user set parameters for the text data in the construction of the class, like this

```
TextObject.create(
            workspace, vertices=xyz, text=random_labels, name="Text Object", color=color
        )
```
instead for forcing to set the attributes after construction. In order for this to work, we need to be specific about the order of the kwargs.

geoh5py/objects/text.py
domfournier: Agreed, moving to setter.

geoh5py/objects/text.py
domfournier: Points have the mechanics to deal with vertices, which this object also has.

MatthieuCMira:
> > Can you explain to me the global idea of the object please. I am a bit confused with the nuance with a TextData, etc.
> 
> The TextObject has been implemented by GA to handle DXF files that contain labels. It is a bit of a mix between Points, TextData and Visual parameters, as it displays the text directly in the viewport, and not in a table like TextData would do. The "values/parameters" are stored as a list of dictionaries (one per text label)
> 
> Main caveats at current stage of implementation are:
> 
>     * users cannot directly modify the content or position of the labels in GA
> 
>     * elevation of the labels are always snapped at 0 (issue raised)
> 
>     * info about the position is duplicated in the `vertices` and `starting position` of the labels.

ok, sounds strange a "text data" cannot be used as label, and we have to pass by another specific object to show our labels.

But that's not on our side, we are just replicating what GA does...

PR #725, opened 2025-04-30
GEOPY-2153: Increase length of possible string in value map

domfournier:
**GEOPY-2153 - Increase length of possible string in value map**


sebhmg (approved):

sebhmg (dismissed):
👍

sebhmg (approved):

PR #733, opened 2025-06-10
GEOPY-2256

MatthieuCMira:
correct bug in geoh5py

I replicate the way to raise the bug in the tests


geoh5py/data/data_type.py
```diff
@@ -687,7 +687,8 @@ def get_parent_reference(self, parent: ObjectBase):
         ref_data = []
         for child in parent.children:
             if (
-                isinstance(child.entity_type, ReferencedValueMapType)
+                hasattr(child, "data_maps")
```
domfournier, line 690:
```suggestion
                hasattr(child, "entity_type")
```
A bit more direct.

PR #743, opened 2025-07-03
GEOPY-2292: Use scientific format with fixed number of decimals to store means

MatthieuCMira:
**GEOPY-2292 - Use scientific format with fixed number of decimals to store means**
deal with no data values

### DIFF
--- a/geoh5py/shared/utils.py
+++ b/geoh5py/shared/utils.py
@@ -1033,7 +1033,7 @@ def array_is_colour(values: np.ndarray) -> bool:
 
 
 def format_numeric_values(
-    values: np.ndarray, n_decimals: int, max_chars: int
+    input_values: np.ndarray, n_decimals: int, max_chars: int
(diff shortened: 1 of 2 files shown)


geoh5py/shared/utils.py
```diff
@@ -1066,7 +1068,10 @@ def format_numeric_values(
     sci_str = np.char.add(np.char.add(man, "e"), man_exp[:, 1])
 
     # choose between normal and scientific notation
-    use_sci = np.char.str_len(normal_str) > max_chars
-    result = np.where(use_sci, sci_str, normal_str)
+    result = np.where(np.char.str_len(normal_str) > max_chars, sci_str, normal_str)
+
+    # replace NaN values with empty strings
+    final_result = np.full(input.shape, "", dtype=object)
```
domfournier, line 1074:
```suggestion
    final_result = np.full(input_values.shape, "", dtype=object)
```

PR #744, opened 2025-07-09
GEOPY-2091

domfournier:


geoh5py/__init__.py
```diff
@@ -18,25 +18,6 @@
 # ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
 
 
-# flake8: noqa
-
 __version__ = "0.12.0a1"
```
MatthieuCMira, line 22: This is quite a commitment to change the init of geoh5py in this issue. 

Are we sure it's used nowhere?

Should be an issue on itself to test our packages

geoh5py/workspace/__init__.py
```diff
@@ -1,24 +0,0 @@
-# ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
```
MatthieuCMira, line 1: oh you moved workspace to geoh5py

But if somehwere in one of our package we are calling from workspace.workspace import Workspace?

geoh5py/data/__init__.py
```diff
@@ -45,3 +46,21 @@
 from .text_data import CommentsData, MultiTextData, TextData
 from .unknown_data import UnknownData
 from .visual_parameters import VisualParameters
+
+
+class TypeEnum(Enum):
```
MatthieuCMira, line 51: it's seems a duplicate with PrimitiveTypeEnum no?
What the difference?
seems redundant

geoh5py/workspace.py
```diff
@@ -423,39 +439,30 @@ def create_data(
...
 
-                if member is TextData and any(
-                    isinstance(val, str) and "Visual Parameters" == val
-                    for val in entity.values()
-                ):
-                    member = VisualParameters
+        member = type_enum.value
 
-                created_entity = member(entity_type=entity_type, **entity)
+        if self.version > 1.0 and isinstance(entity["parent"], ConcatenatedObject):
```
MatthieuCMira, line 451: At some point we should change this object creation...

geoh5py/workspace.py
```diff
@@ -542,41 +549,34 @@ def create_object_or_group(
...
-            ):
-                if self.version > 1.0:
-                    if member in (DrillholeGroup, IntegratorDrillholeGroup):
-                        member = type("Concatenator" + name, (Concatenator, member), {})
-                    elif member is Drillhole and isinstance(
-                        entity.get("parent"),
-                        (DrillholeGroup, IntegratorDrillholeGroup),
-                    ):
-                        member = ConcatenatedDrillhole
+        member = TYPE_UID_TO_CLASS.get(entity_type_uid, None)
```
MatthieuCMira, line 552: cool, way more simpler

geoh5py/workspace/__init__.py
domfournier: Most packages do from `from geoh5py import Workspace` or `from geoh5py.workspace import Workspace`, which is still supported here.

domfournier: But fair enough. I will raise a deprecation warning so we have time to adjust.

geoh5py/data/data.py
```diff
@@ -266,11 +267,6 @@ def modifiable(self, value: bool):
         if self.on_file:
             self.workspace.update_attribute(self, "attributes")
 
-    @classmethod
```
MatthieuCMira, line 269: deprecation warning using the Primitive typeEnum?

geoh5py/objects/surveys/electromagnetics/base.py
```diff
@@ -710,7 +708,7 @@ def _format_transmitter_ids(self, values, attributes):
             value_map[0] = "Unknown"
             attributes.update(
                 {
-                    "primitive_type": "REFERENCED",
```
MatthieuCMira, line 713: I guess the "primitive_key" is still accepted here?

MatthieuCMira (commented):
It's getting huge. 

It would be great if we can test if it break anything in other packages automatically....
For now, someone must do it manually before we merge...

PR #747, opened 2025-07-16
GEOPY-2272

MatthieuCMira:
add incremental (n) on add data and add files.

I added tests.

I comments error based on naming.
Might be possible to still raise those errors doing
data.name = "duplicate name" after add_data
But do we want that?

### DIFF
--- a/geoh5py/objects/object_base.py
+++ b/geoh5py/objects/object_base.py
@@ -47,6 +47,7 @@ from ..shared.utils import (
     array_is_colour,
     box_intersect,
     clear_array_attributes,
+    find_unique_name,
     mask_by_extent,
     str2uuid,
 )
@@ -177,12 +178,23 @@ class ObjectBase(EntityContainer):
 
         property_groups: dict[PropertyGroup | None, list[Data]] = {}
         data_objects = []
+
+        names = [
+            child.name
+            for child in self.children
+            if isinstance(child, Data) and child.name != "Visual Parameters"
+        ]
+
         for name, attr in data.items():
             if not isinstance(attr, dict):
                 raise TypeError(
                     f"Given value to data {name} should of type {dict}. "
                     f"Type {type(attr)} given instead."
                 )
+
+            name = find_unique_name(name, names)
+            if name != "Visual Parameters":
+                names.append(name)
 
             attr, validate_property_group = self.validate_association(
                 {**attr, "name": name}, property_group=property_group, **kwargs
@@ -235,8 +247,10 @@ class ObjectBase(EntityContainer):
         """
         data_maps = data.data_maps or {}
 
-        if name in data_maps:
-            raise KeyError(f"Data map '{name}' already exists.")
+        names = [
+            child.name for child in self.children if isinstance(child, ReferencedData)
+        ]
+        name = find_unique_name(name, names)
 
         if not isinstance(values, dict | np.ndarray):
             raise TypeError("Data map values must be a numpy array or dict")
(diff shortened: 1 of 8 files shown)


tests/drillhole_v4_0_test.py
```diff
@@ -366,15 +366,16 @@ def test_create_drillhole_data(tmp_path):  # pylint: disable=too-many-statements
...
+        # note: not possible anymore! removing?
+        # with pytest.raises(ValueError, match="already present on the drillhole"):
+        #     well.add_data(
+        #         {
+        #             "my_log_values/": {
+        #                 "depth": np.arange(0, 50.0),
+        #                 "values": np.random.randn(50),
+        #             },
+        #         }
+        #     )
```
domfournier, line 378: Yep, you can delete for good.

tests/drillhole_v4_0_test.py
```diff
@@ -470,19 +471,20 @@ def test_append_data_to_tables(tmp_path):
...
+        #     well_b.add_data(
+        #         {
+        #             "Depth Data": {
+        #                 "values": np.random.randn(10),
+        #                 "depth": np.sort(
+        #                     np.random.uniform(low=0.05, high=100, size=(10,))
+        #                 ),
+        #             },
+        #         }
+        #     )
```
domfournier, line 487: Same, clean it out.

geoh5py/objects/object_base.py
```diff
@@ -271,7 +285,7 @@ def add_data_map(self, data: ReferencedData, name: str, values: dict):
         )
         data_maps[name] = geom_data
         data.data_maps = data_maps
-        return geom_data
+        return data_type
```
domfournier, line 288: You are returning the type? I think we should still return the data itself, like `add_data` does.
Curious why we don't have a test that catches that.

tests/reference_data_test.py
```diff
@@ -146,8 +146,9 @@ def test_add_data_map(tmp_path):
         data.entity_type.value_map = value_map
         data.add_data_map("test", data_map)
 
-        with pytest.raises(KeyError, match="Data map 'test' already exists."):
-            data.add_data_map("test", data_map)
+        # not possible anymore
+        # with pytest.raises(KeyError, match="Data map 'test' already exists."):
+        #     data.add_data_map("test", data_map)
```
domfournier, line 151: Here too

geoh5py/objects/object_base.py
```diff
@@ -177,13 +178,24 @@ def add_data(
...
+        ]
+
         for name, attr in data.items():
             if not isinstance(attr, dict):
                 raise TypeError(
                     f"Given value to data {name} should of type {dict}. "
                     f"Type {type(attr)} given instead."
                 )
 
+            name = find_unique_name(name, names)
```
domfournier, line 195: We probably want to prevent VisualParameters all together. Here it could add another `Visual Parameters` with the same name.

domfournier: So right now, the way it's written
1- You pull all names other than Visual Parameters
2- Loop over all entries and increment the names (if needed). If the user passes a "Visual Parameter" data, then the name won't be incremented and potentially end up with two "Visual Parameters" if one exists already, which is a problem.

My point was to simply raise a `UserWarning` right away if the name "Visual Parameter" is encountered in the `add_data`. It should never be encountered here in the first place. Either it is coming from GA, or it is forced to go through the `add_default_visual_parameters` on the object to force formatting. And you can remove the check below as well.

domfournier (changes requested):

domfournier (approved):

PR #752, opened 2025-07-31
GEOPY-2317: Expose filter basement visual parameters to the VPmesh object.

domfournier:
**GEOPY-2317 - Expose filter basement visual parameters to the VPmesh object.**
…model

### DIFF
--- a/geoh5py/data/visual_parameters.py
+++ b/geoh5py/data/visual_parameters.py
@@ -36,6 +36,14 @@ class VisualParameters(TextData):
     _xml: ET.Element | None = None
     _association = DataAssociationEnum.OBJECT
 
+    def __init__(
+        self,
+        association: DataAssociationEnum = DataAssociationEnum.OBJECT,
+        public: bool = False,
+        **kwargs,
+    ):
+        super().__init__(association=association, public=public, **kwargs)
+
     @property
     def xml(self) -> ET.Element:
         """
--- a/geoh5py/data/visual_parameters.py
+++ b/geoh5py/data/visual_parameters.py
@@ -110,6 +110,30 @@ class VisualParameters(TextData):
 
         self.set_tag("Colour", str(value))
 
+    @property
+    def filter_basement(self) -> None | float:
+        """
+        Filter basement tag for VPmodel.
+        """
+        element = self.get_tag("Filterbasement")
+
+        if element is None or not element.text:
+            return None
+
+        return float(element.text)
+
+    @filter_basement.setter
+    def filter_basement(self, value: float):
+        """
+        Set the filter basement value.
+
+        :param value: The filter basement value.
+        """
+        if not isinstance(value, (int, float)):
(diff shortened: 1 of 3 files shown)


geoh5py/data/visual_parameters.py
```diff
@@ -110,6 +110,30 @@ def colour(self, rgb: list | tuple | np.ndarray):
...
         self.set_tag("Colour", str(value))
 
+    @property
+    def filter_basement(self) -> None | float:
+        """
+        Filter basement tag for VPmodel.
+        """
+        element = self.get_tag("Filterbasement")
+
+        if element is None or not element.text:
```
MatthieuCMira, line 120: are we sure `element.text` is always float?

geoh5py/objects/vp_model.py
```diff
@@ -165,6 +165,28 @@ def _promote_uuid_attribute(
...
+
+        :param workspace: Workspace to be added to.
+        :param kwargs: List of keyword arguments defining the properties of a class.
+
+        :return entity: Registered Entity to the workspace.
+        """
+        new_object = super().create(workspace, **kwargs)
+
+        viz_params = new_object.add_default_visual_parameters()
+        viz_params.filter_basement = (
```
MatthieuCMira, line 184: can you just explain the logic behind this defualt value?

tests/vp_mesh_test.py
```diff
@@ -106,6 +106,8 @@ def test_create_vp_model(tmp_path: Path):
     ):
         VPModel.create(ws, layers=layers, prisms=prisms, unit_property_id="abc")
 
+    assert isinstance(drape.visual_parameters.filter_basement, float)
```
MatthieuCMira, line 109: Shouldn't you test the default value? 
filter.basement == ...

What happens in the case of a VP mesh containing a visual parameter with no filter_basement value? 
Shouldn't we assign it automatically in @property? with a warning for example?

MatthieuCMira (changes requested):
good for me. 
Just missing testing the values

And some questions

geoh5py/data/visual_parameters.py
domfournier: Currently, VisualParameters is agnostic of the parent object. So if `element` is None, that means it's not a VPmodel, so no we don't want to set it.

geoh5py/objects/vp_model.py
domfournier: See the docstrings of the method above. We need to assign some "small" value, which I decided to define as 5% of the total Vpmodel depth.

tests/vp_mesh_test.py
domfournier: GA always creates it, and our `create` method also always adds it too. So it's expected that Filterbasement is always present/

MatthieuCMira (dismissed):
good to go

MatthieuCMira (approved):
going to test

PR #753, opened 2025-08-06
GEOPY-2301: The fxmean attached to a referenced data is getting always the same name

MatthieuCMira:
**GEOPY-2301 - The fxmean attached to a referenced data is getting always the same name**
data_maps now  increment names
can be made public/private (default True => actual behavious)

tested.

I also add a text test to show you what I would like to do with petro-lingo. But we cannot visualize text data (when containing only 1 text) in GA...

### DIFF
--- a/geoh5py/objects/object_base.py
+++ b/geoh5py/objects/object_base.py
@@ -240,7 +240,7 @@ class ObjectBase(EntityContainer):
         return data_objects
 
     def add_data_map(
-        self, data: ReferencedData, name: str, values: dict
+        self, data: ReferencedData, name: str, values: dict, public: bool = True
     ) -> GeometricDataConstants:
         """
         Add a data map to the reference data under the object.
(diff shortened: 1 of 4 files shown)


tests/text_data_test.py
```diff
@@ -155,3 +155,37 @@ def test_create_byte_text_data(tmp_path: Path):
 
             word.values = np.array([b"b word"])[0]
             assert word.values == "b word"
+
+
+def test_create_one_text_data(tmp_path: Path):
```
domfournier, line 160: This one looks out of scope?

geoh5py/objects/object_base.py
```diff
@@ -240,19 +240,22 @@ def add_data(
...
         :param values: The values to add to the data map.
+        :param public: Whether the data map is public or not.
         """
         data_maps = data.data_maps or {}
 
         names = [
-            child.name for child in self.children if isinstance(child, ReferencedData)
+            child.name
+            for child in self.children
+            if isinstance(child, (ReferencedData, GeometricDataConstants))
```
MatthieuCMira, line 258:
```suggestion
            if isinstance(child,  GeometricDataConstants)
```

domfournier (approved):

PR #760, opened 2025-08-26
GEOPY-2413: Error copying Geometric data on cell_objects

domfournier:
**GEOPY-2413 - Error copying Geometric data on cell_objects**


MatthieuCMira (commented):
I am a bit afraid of the new typîgn for such high level function. 

They are pre-commit errors to fix first

MatthieuCMira (approved):
good to go

PR #769, opened 2025-09-11
GEOPY-2440: add a copy group+object base in inputfile as it exists uin uijhson group

MatthieuCMira:
**GEOPY-2440 - add a copy group+object base in inputfile as it exists uin uijhson group**
add a copy_relative method in Inputfile based on the uijson group.

Move the mechanisms to shared utils


domfournier (dismissed):

geoh5py/groups/uijson.py
```diff
@@ -121,7 +107,7 @@ def copy(
         parent=None,
         *,
         copy_children: bool = False,
-        copy_relatives: bool = True,
+        copy_relatives: bool = False,
```
domfournier, line 110: That won't work because the nested UIJGroup won't copy theirs

image

domfournier (approved):

PR #780, opened 2025-10-08
GEOPY-2453: object stock on display after use of geoh5py

domfournier:
**GEOPY-2453 - object stock on display after use of geoh5py**


geoh5py/objects/object_base.py
```diff
@@ -222,6 +222,7 @@ def add_data(
             if isinstance(data_object, VisualParameters):
                 self.visual_parameters = data_object
 
+            data_object.visible = attr.get("visible", False)
```
MatthieuCMira: the first data visible is not set to visible = False in GA

PR #790, opened 2025-10-22
GEOPY-1261: Return None if path2workspace does not find a file

MatthieuCMira:
**GEOPY-1261 - Return None if path2workspace does not find a file**
pass in path2workspace if the workspace does not exist.

Agentic typing and docstrings


benk-mira (approved):
path2workspace changes look fine.  I know you want to update docstrings for your AI agent to digest, but can we open new issues to handle this?  I don't want to miss details of the actual PR because the diff is flooded with docstring changes!

PR #802, opened 2025-11-04
GEOPY-2499: BaseUIJson crashes in to_params validation when skipping over disabled fields.

benk-mira:
**GEOPY-2499 - BaseUIJson crashes in to_params validation when skipping over disabled fields.**

### DIFF
--- a/geoh5py/ui_json/ui_json.py
+++ b/geoh5py/ui_json/ui_json.py
@@ -35,6 +35,8 @@ from pydantic import (
 )
 
 from geoh5py import Workspace
+from geoh5py.groups import PropertyGroup
+from geoh5py.shared import Entity
 from geoh5py.shared.utils import fetch_active_workspace
 from geoh5py.shared.validators import empty_string_to_none, none_to_empty_string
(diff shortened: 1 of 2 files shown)


geoh5py/ui_json/ui_json.py
```diff
@@ -220,6 +220,8 @@ def validate_data(
 
         ui_json = self.model_dump(exclude_unset=True)
         for field in self.model_fields_set:
+            if self.is_disabled(field):
```
MatthieuCMira: does is disable is also checking for :
group dependencies? object dependencies? 
If so I'm ok! :-)

MatthieuCMira (changes requested):
See my question in the comment. 


Maybe you could do a unitest to replicate the initial crash to ensure it works?

benk-mira:
> See my question in the comment.
> 
> Maybe you could do a unitest to replicate the initial crash to ensure it works?

Added a line to my unit test for is_disabled (call `to_params()` which would have failed before this PR)

MatthieuCMira (approved):
ok for me

PR #804, opened 2025-11-10
GEOPY-2488: UIJson should have a flatten method to retrieve value/property of all forms

benk-mira:
**GEOPY-2488 - UIJson should have a flatten method to retrieve value/property of all forms**


domfournier (dismissed):
Nice

PR #808, opened 2025-11-14
GEOPY-2564: Basement added to the referenced value map not visible in GA

domfournier:
**GEOPY-2564 - Basement added to the referenced value map not visible in GA**


benk-mira (changes requested):
couple suggestions

PR #813, opened 2025-11-19
GEOPY-2575: Add data_map crashes on float values trying to rename

domfournier:
**GEOPY-2575 - Add data_map crashes on float values trying to rename**

### DIFF
--- a/geoh5py/shared/utils.py
+++ b/geoh5py/shared/utils.py
@@ -965,13 +965,16 @@ def find_unique_name(name: str, names: list[str], case_sensitive=True) -> str:
     :param names: List of names to avoid.
     :return: A unique name.
     """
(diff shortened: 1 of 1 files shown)


geoh5py/shared/utils.py
```diff
@@ -971,7 +971,7 @@ def find_unique_name(name: str, names: list[str], case_sensitive=True) -> str:
         names_list = names
 
     checkname = name.lower() if not case_sensitive and isinstance(name, str) else name
-    if checkname not in names_list:
+    if checkname not in names_list or not isinstance(name, str):
```
sebhmg: testing name being a string should be the first thing done in the function:
```py
if not isinstance(name, str):
    return name
```

sebhmg (approved):

PR #817, opened 2025-11-26
GEOPY-2540: Copy from extent GeoImage takes forever.

MatthieuCMira:
**GEOPY-2540 - Copy from extent GeoImage takes forever.**
Changing the geoimage method to do something way faster and cleaner

### DIFF
--- a/geoh5py/shared/cut_by_extent.py
+++ b/geoh5py/shared/cut_by_extent.py
@@ -27,29 +27,14 @@ from geoh5py.shared.utils import (
     clean_extent_for_intersection,
     ensure_counter_clockwise,
     normalize,
+    validate_3d_array,
+    validate_normalized_vector,
 )
 
 
 if TYPE_CHECKING:
     from geoh5py.objects.geo_image import GeoImage
     from geoh5py.objects.grid2d import Grid2D
-
-
-def validate_3d_array(value: np.ndarray) -> np.ndarray:
-    """Validate that input is a 3D numpy array."""
-    if not isinstance(value, np.ndarray):
-        raise TypeError(f"Expected numpy array, got {type(value).__name__}")
-    if value.shape != (3,):
-        raise ValueError(f"Expected shape (3,), got {value.shape}")
-    return value
-
-
-def validate_normalized_vector(value: np.ndarray) -> np.ndarray:
-    """Validate that input is a normalized 3D vector."""
-    value = validate_3d_array(value)
-    if not np.isclose(np.linalg.norm(value), 1.0, atol=1e-6):
-        raise ValueError("Vector is not normalized.")
-    return value
 
 
 class Plane(BaseModel):
--- a/geoh5py/shared/utils.py
+++ b/geoh5py/shared/utils.py
@@ -1299,3 +1299,32 @@ def ensure_counter_clockwise(polygon: np.ndarray) -> np.ndarray:
         return polygon[::-1]
 
     return polygon
+
+
+def validate_3d_array(value: np.ndarray) -> np.ndarray:
+    """
+    Validate that input is a 3D numpy array.
+
+    :param value: Input array to validate.
+
+    :return: The validated 3D array.
+    """
+    if not isinstance(value, np.ndarray):
+        raise TypeError(f"Expected numpy array, got {type(value).__name__}")
+    if value.shape != (3,):
+        raise ValueError(f"Expected shape (3,), got {value.shape}")
+    return value
+
+
+def validate_normalized_vector(value: np.ndarray) -> np.ndarray:
+    """
+    Validate that input is a normalized 3D vector.
+
+    :param value: Input array to validate.
+
+    :return: The validated normalized 3D vector.
+    """
+    value = validate_3d_array(value)
+    if not np.isclose(np.linalg.norm(value), 1.0, atol=1e-6):
+        raise ValueError("Vector is not normalized.")
+    return value
--- a/tests/cut_by_extent_test.py
+++ b/tests/cut_by_extent_test.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 
 import numpy as np
 import pytest
+from pydantic import ValidationError
 
 from geoh5py.shared.cut_by_extent import Plane
 
@@ -27,42 +28,21 @@ from geoh5py.shared.cut_by_extent import Plane
 @pytest.mark.parametrize(
     "origin,u,v,match",
     [
-        # origin shape error
+        # bad shape
         (
             np.array([0.0, 0.0]),
             np.array([1.0, 0.0, 0.0]),
             np.array([0.0, 1.0, 0.0]),
-            r"origin must be shape \(3,\), got",
+            r"Expected shape \(3,\), got \(2,\)",
         ),
-        # u_vector shape error
-        (
-            np.array([0.0, 0.0, 0.0]),
-            np.array([1.0, 0.0]),
-            np.array([0.0, 1.0, 0.0]),
-            r"u_vector must be shape \(3,\), got",
-        ),
-        # v_vector shape error
-        (
-            np.array([0.0, 0.0, 0.0]),
-            np.array([1.0, 0.0, 0.0]),
-            np.array([0.0, 1.0]),
-            r"v_vector must be shape \(3,\), got",
-        ),
-        # u not normalized
+        # non-normalized
         (
             np.array([0.0, 0.0, 0.0]),
             np.array([2.0, 0.0, 0.0]),
             np.array([0.0, 1.0, 0.0]),
-            r"u_vector is not normalized\.",
+            r"Vector is not normalized\.",
         ),
-        # v not normalized
-        (
-            np.array([0.0, 0.0, 0.0]),
-            np.array([1.0, 0.0, 0.0]),
-            np.array([0.0, 2.0, 0.0]),
-            r"v_vector is not normalized\.",
-        ),
-        # not orthogonal
+        # non-orthogonal
         (
             np.array([0.0, 0.0, 0.0]),
             np.array([1.0, 0.0, 0.0]),
@@ -73,7 +53,7 @@ from geoh5py.shared.cut_by_extent import Plane
 )
 def test_plane_validate_uv_errors(origin, u, v, match):
     """Ensure Plane model_validator raises for invalid origin/u/v inputs with message match."""
-    with pytest.raises(ValueError, match=match):
+    with pytest.raises(ValidationError, match=match):
         Plane(origin=origin, u_vector=u, v_vector=v)
 
 


geoh5py/objects/geo_image.py
```diff
@@ -150,42 +157,74 @@ def copy_from_extent(
...
+        clipped_uv, plane = compute_clipped_polygon_uv(
+            self.vertices, extent, eps_plane=1e-5, eps_collinear=1e-6
         )
 
-        if grid_transformed is None:
-            grid.workspace.remove_entity(grid)
-            warnings.warn("Image could not be cropped.")
+        # short-circuit degenerate cases
+        if clipped_uv.shape[0] < 3:
+            warnings.warn("The image and the extent doesn't intersect.")
```
domfournier:
```suggestion
```
Otherwise when we batch process a whole workspace, we will get a ton of prints to screen.

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,551 @@
...
+#                                                                              '
+#  You should have received a copy of the GNU Lesser General Public License    '
+#  along with geoh5py.  If not, see <https://www.gnu.org/licenses/>.           '
+# ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+from __future__ import annotations
+
+import numpy as np
+
+
+def normalize(v: np.ndarray) -> np.ndarray:
```
domfournier:
```suggestion
def normalize(vector: np.ndarray) -> np.ndarray:
```
Avoid single-letter variables

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,551 @@
...
+    v = normalize(vertices[0] - p0)
+    n = normalize(np.cross(u, v))
+    return p0, u, v, n
+
+
+def intersect_segment_plane(
+    a: np.ndarray,
+    b: np.ndarray,
+    p0: np.ndarray,
+    n: np.ndarray,
```
domfournier:
```suggestion
    start: np.ndarray,
    end: np.ndarray,
    origin: np.ndarray,
    normal: np.ndarray,
```
Same here, more descriptive variables are prefered

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,551 @@
...
+    if extent.shape[1] == 2:
+        # For 2D extents, add Z bounds with slight expansion for safety
+        z_bounds = np.array([[zmin - 1], [zmax + 1]], dtype=np.float64)
+        extent = np.column_stack([extent.astype(np.float64), z_bounds])
+
+    return extent
+
+
+#  pylint: disable=too-many-locals
+def compute_plane_box_intersections(
```
domfournier: Let's see if we can clean up the function to avoid too many locals
```suggestion
def plane_box_intersection(
```

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,551 @@
...
+        return np.zeros((0, 3), dtype=np.float64)
+
+    return np.asarray(points, dtype=np.float64)
+
+
+def project_to_uv(
+    p: np.ndarray,
+    p0: np.ndarray,
+    u: np.ndarray,
+    v: np.ndarray,
```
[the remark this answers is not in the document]
domfournier: How about we define a `Plane` dataclass, so we can just pass it around all the functions.

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,551 @@
...
+    :param uv: UV coordinates of polygon vertices, shape (N, 2).
+    :param u_cell: Size of one pixel in U direction.
+    :param v_cell: Size of one pixel in V direction.
+    :param width: Image width in pixels (U direction).
+    :param height: Image height in pixels (V direction).
+
+    :return: Tuple of (xmin, ymin, xmax, ymax) pixel coordinates for PIL crop.
+    """
+    if len(uv) == 0:
+        return 0, 0, 0, 0
```
domfournier: Still return an index if no overlap?

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,551 @@
...
+    if xmin >= xmax or ymin >= ymax:
+        return 0, 0, 0, 0
+
+    return xmin, ymin, xmax, ymax
+
+
+# pylint: disable=too-many-locals
+def pixel_rect_to_world(
+    extent: tuple[int, int, int, int],
+    plane: tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray],
```
domfournier: Another place where having a `Plane` object would be nicer.

geoh5py/objects/geo_image.py
```diff
@@ -150,42 +157,74 @@ def copy_from_extent(
...
+            if hasattr(kwargs["parent"], "h5file"):
+                workspace = kwargs.pop("parent")
+            else:
+                workspace = kwargs["parent"].workspace
+        else:
+            workspace = self.workspace
+
+        new_attributes = GeoImageConversion.verify_kwargs(self, **kwargs)
+
+        return GeoImage.create(workspace, **new_attributes)
```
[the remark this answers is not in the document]
MatthieuCMira: small detial: parent can also be a group, must tackle this

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,551 @@
...
+
+    # Convert UV bounds to pixel indices
+    xmin = max(0, int(np.floor(u_min / u_cell - 0.5)) + 1)
+    xmax = min(width, int(np.floor(u_max / u_cell - 0.5)) + 2)
+    ymin = max(0, int(np.floor(v_min / v_cell - 0.5)) + 1)
+    ymax = min(height, int(np.floor(v_max / v_cell - 0.5)) + 2)
+
+    # Ensure we have a valid rectangle
+    if xmin >= xmax or ymin >= ymax:
+        return 0, 0, 0, 0
```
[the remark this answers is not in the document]
MatthieuCMira: pixel_extent_from_polygon:

=> x_min, x_max, y_min, y_min => centers of the image's cells might be in the extent. 

=> 0, 0, 0, 0 => no extent, no cell center to keep.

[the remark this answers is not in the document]
MatthieuCMira: I'll change it to None, and return None if 0,0,0,0

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,518 @@
...
+        :param vertices: Array of shape (4, 3) containing the four vertices of the quad.
+
+        :return: Plane object containing origin and basis vectors.
+
+        :raises ValueError: If vertices is not a (4, 3) array.
+        """
+        if vertices.shape != (4, 3):
+            raise ValueError("vertices must be a (4, 3) array")
+
+        vertices = np.asarray(vertices)
```
benk-mira: Should vertices be typed `list | np.ndarray` then?

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,518 @@
...
+    :param normal: Normal vector of the plane.
+    """
+
+    origin: np.ndarray
+    u_vector: np.ndarray
+    v_vector: np.ndarray
+    normal: np.ndarray
+
+    @classmethod
+    def from_quad_origin_last(cls, vertices: np.ndarray) -> Plane:
```
benk-mira: Seems more natural to pass origin and an array of co-planar vertices in and let the caller sort out that the last vertex is the origin.

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,518 @@
...
+def normalize(vector: np.ndarray) -> np.ndarray:
+    """
+    Normalize a vector to unit length.
+
+    :param vector: Input vector to normalize.
+    :param dtype: NumPy data type for the normalized vector.
+
+    :return: Normalized vector with unit length.
+    """
+    vector = np.asarray(vector, dtype=np.float64)
```
benk-mira: vector can be typed `list | np.ndarray`

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,518 @@
...
+    return vector / np.linalg.norm(vector)
+
+
+def intersect_segment_plane(
```
benk-mira: all the functions feel like they should be methods of the Plane class, any reason to keep it a dataclass?

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,518 @@
...
+    intersection_points: list[np.ndarray] = []
+    for corner_a, corner_b in box_edges:
+        intersection = intersect_segment_plane(
+            box_corners[corner_a], box_corners[corner_b], plane, eps_plane=eps_plane
+        )
+        if intersection is not None:
+            intersection_points.append(intersection)
+
+    if not intersection_points:
+        return np.zeros((0, 3), dtype=np.float64)
```
benk-mira:
```suggestion
        return np.empty((0, 3))
```

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,518 @@
...
+    :return: 2D coordinates in the UV system.
+    """
+    displacement = np.asarray(point, dtype=np.float64) - plane.origin
+    return np.array(
+        [np.dot(displacement, plane.u_vector), np.dot(displacement, plane.v_vector)],
+        dtype=np.float64,
+    )
+
+
+def ensure_ccw(polygon: np.ndarray) -> np.ndarray:
```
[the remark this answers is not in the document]
benk-mira: also clipping...
image

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,518 @@
...
+    subject_polygon = ensure_ccw(subject_polygon)
+    clipping_polygon = ensure_ccw(clipping_polygon)
+
+    output_vertices = subject_polygon
+    for i, current_clip_edge_start in enumerate(clipping_polygon):
+        input_vertices = output_vertices
+        current_clip_edge_end = clipping_polygon[(i + 1) % len(clipping_polygon)]
+
+        if len(input_vertices) == 0:
+            return np.zeros((0, 2), dtype=subject_polygon.dtype)
```
benk-mira:
```suggestion
            return np.empty((0, 2), dtype=subject_polygon.dtype)
```

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,551 @@
...
+    vertices = np.asarray(vertices)
+
+    p0 = vertices[3]
+    u = normalize(vertices[2] - p0)
+    v = normalize(vertices[0] - p0)
+    n = normalize(np.cross(u, v))
+    return p0, u, v, n
+
+
+def intersect_segment_plane(
```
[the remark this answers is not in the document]
benk-mira: True, but I think we should look at packages like Shapely and TriMesh for 3rd party implementations to reduce the number of utils to maintain as well.

geoh5py/shared/utils.py
```diff
@@ -654,6 +654,39 @@ def box_intersect(
...
+) -> np.ndarray:
+    """
+    Clean and prepare extent array for 3D intersection calculations.
+
+    :param extent: Input extent array, shape (2, 2) or (2, 3)
+    :param locations: Array of vertices to extract Z bounds from, shape (N, 3)
+
+    :return: Cleaned extent array with shape (2, 3)
+    """
+    # raises the eventual errors
```
benk-mira: Not sure I understand your comment here, the vstack is expected to fail?

geoh5py/objects/geo_image.py
```diff
@@ -323,6 +362,8 @@ def georeferencing_from_tiff(self):
             v_cell_size = float(self.tag[33550][1])
             u_count = float(self.tag[256][0])
             v_count = float(self.tag[257][0])
+
+            # todo: not anymore because it's not always a rectangle
```
benk-mira: Address now or in another PR?

benk-mira: ..just checking because sometimes I leave myself a todo comment to come back in the same issue.  If it isn't the case, maybe you can create an issue to address the todo and add the issue number in the comment.

geoh5py/objects/geo_image.py
```diff
@@ -524,9 +570,9 @@ def set_tag_from_vertices(self):
         if self.image is None:
             raise AttributeError("An 'image' must be set before georeferencing.")
 
-        width, height = self.image.size
-        self._tag[256] = (width,)
-        self._tag[257] = (height,)
+        self._tag[256] = (self.u_count,)
+        self._tag[257] = (self.v_count,)
+        # todo: It should be a 6 * 4 tuples defining the affine transformation
```
benk-mira: Now or later?

benk-mira (changes requested):
Some nice geometry utilities in there.  Did you check to see if any other packages could handle them first.  I think a package like TriMesh might be able to do the intersection stuff, and there's Shapely for any 2d clipping, intersection, ccw checks.

geoh5py/objects/geo_image.py
MatthieuCMira: later GEOPY-2591

geoh5py/objects/geo_image.py
MatthieuCMira: later [GEOPY-2591](https://mirageoscience.atlassian.net/browse/GEOPY-2591)

MatthieuCMira:
"True, but I think we should look at packages like Shapely and TriMesh for 3rd party implementations to reduce the number of utils to maintain as well."

=> agree, but the politic is to keep geoh5py with a minimum of dependency. 
That's something to decide...

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,518 @@
...
+                        next_vertex,
+                        current_clip_edge_start,
+                        current_clip_edge_end,
+                        eps_collinear,
+                    )
+                )
+                clipped_vertices.append(next_vertex)
+
+        if not clipped_vertices:
+            return np.zeros((0, 2), dtype=subject_polygon.dtype)
```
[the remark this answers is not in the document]
MatthieuCMira: not changing anything as it's an (0, 2) shape array, but as you want

geoh5py/shared/cut_by_extent.py
MatthieuCMira: The problem is that the origin is ALWAYS the last for geoimage.

So you would like to pass the index of the origin as an attribute?

`def from_quad_origin_last(cls, vertices: np.ndarray, origin:in = 3) -> Plane:`

MatthieuCMira: oh, ok, pass origin, u and v points. Yes, could be!

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,616 @@
...
+        )
+        v_vector = normalize(
+            np.asarray(point_v, dtype=float) - np.asarray(origin, dtype=float)
+        )
+
+        return cls(
+            origin=np.asarray(origin, dtype=float), u_vector=u_vector, v_vector=v_vector
+        )
+
+    @classmethod
```
benk-mira: Unless it's a constructor, or you are modifying the class itself, just use a staticmethod

geoh5py/shared/cut_by_extent.py
```diff
@@ -0,0 +1,616 @@
...
+            raise ValueError("u_vector is not normalized.")
+        if not np.isclose(np.linalg.norm(self.v_vector), 1.0, atol=1e-6):
+            raise ValueError("v_vector is not normalized.")
+        if not np.isclose(float(np.dot(self.u_vector, self.v_vector)), 0.0, atol=1e-8):
+            raise ValueError("u_vector and v_vector are not orthogonal.")
+
+        return self
+
+    @classmethod
+    def create_from_points(
```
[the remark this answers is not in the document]
benk-mira: ..and for convenience for use with extents, why not add a constructor `Plane.from_extent()` that indexes the origin and u, v points.

PR #848, opened 2026-03-06
GEOPY-2746: implement "Group Value" in BaseUIJSON

MatthieuCMira:
**GEOPY-2746 - "Group Value" form is not implemented in BaseUIJSON**
add a new form for multi Data Group

accept tooltips as list

### DIFF
--- a/geoh5py/ui_json/forms.py
+++ b/geoh5py/ui_json/forms.py
@@ -574,12 +574,12 @@ class MultiDataGroupForm(BaseForm):
     group_value: OptionalUUID
 
     data_type: DataType | list[DataType]
-    value: list[str]
+    value: str | list[str]
     multi_select: bool = True
 
     @field_validator("value", mode="before")
     @classmethod
-    def to_list(cls, value: str | list[str]) -> list[str]:
+    def to_list(cls, value: str | list[str]) -> str | list[str]:
         """
         Validate that value is a list, converting it if it's a string.
 
@@ -587,6 +587,8 @@ class MultiDataGroupForm(BaseForm):
 
         :return: A list of strings representing the value.
         """
+        if not value or value == [""]:
+            return ""
         if not isinstance(value, list):
             value = [value]
         return value
--- a/geoh5py/ui_json/forms.py
+++ b/geoh5py/ui_json/forms.py
@@ -699,16 +699,51 @@ def all_subclasses(type_object: type[BaseForm]) -> list[type[BaseForm]]:
     return collection
 
 
-def model_fields_difference(parent: type[BaseForm], child: type[BaseForm]) -> set[str]:
-    """Isolate fields added in the child class."""
-    return set(child.model_fields) - set(parent.model_fields)
+def get_mandatory_attributes(to_inspect: type[BaseForm]) -> set[str]:
+    """
+    Isolate mandatory attributes for a class.
+
+    :param to_inspect: The pydantic class to check.
+
+    :return: The attributes with no default values.
+    """
+    return {
+        key
+        for key, value in to_inspect.model_fields.items()
+        if value.default is None and not value.is_required()
+    }
 
 
 def indicator_attributes(
     parent: type[BaseForm], children: list[type[BaseForm]]
-) -> list[set[str]]:
-    """List all the attributes defined in a subclass."""
-    return [model_fields_difference(parent, c) for c in children]
+) -> list[list[set[str]]]:
+    """
+    List all the mandatory attributes defined in a subclass.
+
+    The function return a list of 2 lists:
+    - the first contains the sets of attributes
+        that are different between the parent and each child class.
+    - the second contains the sets of mandatory attributes
+        that are different between the parent and each child class.
+
+    :param parent: The parent class to compare against.
+    :param children: The list of child classes to compare.
+
+    :return: A list of mandatory attributes defined in a subclass.
+    """
+    parent_mandatory_attributes = get_mandatory_attributes(parent)
+    parent_attributes = set(parent.model_fields)
+
+    mandatory_differences = [
+        get_mandatory_attributes(child) - parent_mandatory_attributes
+        for child in children
+    ]
+
+    full_differences = [
+        set(child.model_fields) - parent_attributes for child in children
+    ]
+
+    return [full_differences, mandatory_differences]
 
 
 def filter_candidates_by_indicator_polling(
@@ -722,6 +757,9 @@ def filter_candidates_by_indicator_polling(
     any ambiguity between non-unique indicators such as 'choice_list'
     and 'multi_select'.
 
+    :param data: The form data to check for matching indicators.
+
+        :return: An array of candidate subclasses with the most matching indicators.
     """
     counts = count_indicators(INDICATORS, data)
     candidates = np.array(FORM_TYPES)[counts == np.max(counts)]
@@ -736,9 +774,28 @@ def filter_candidates_by_indicator_polling(
     return candidates
 
 
-def count_indicators(indicators: list[set[str]], data: dict[str, Any]) -> np.ndarray:
-    """Return the number of matching indicators for each child class."""
-    return np.array([len(i.intersection(data)) for i in indicators])
+def count_indicators(
+    indicators: list[list[set[str]]], data: dict[str, Any]
+) -> np.ndarray:
+    """
+    Count the number of matching indicators for each child class.
+
+    1. count the number of mandatory indicators present in the form data for each subclass.
+        It allows to select the valid classes.
+    2. count the number of indicators not present in the form data for each
+        It allows to rank the incompatibility per classes.
+        (Note: technically, we could just keep the one equal to 0 but risky)
+    3. return the difference between the two counts for each subclass.
+
+    :param indicators: A list of sets of indicator attributes for each subclass.
+    :param data: The form data to check for matching indicators.
+
+    :return: An array of counts of matching indicators for each subclass.
+    """
+    count_interection = np.array([len(i.intersection(data)) for i in indicators[0]])
+    count_difference = np.array([len(i.difference(data)) for i in indicators[1]])
+
+    return count_interection - count_difference
 
 
 def filter_candidates_by_type_checking(
--- a/geoh5py/ui_json/forms.py
+++ b/geoh5py/ui_json/forms.py
@@ -108,7 +108,7 @@ class BaseForm(BaseModel):
     optional: bool = False
     enabled: bool = True
     main: bool = True
-    tooltip: str = ""
+    tooltip: str | list[str] = ""
     group: str = ""
     group_optional: bool = False
     dependency: str = ""
@@ -160,21 +160,6 @@ class BaseForm(BaseModel):
 
     def validate_data(self, params: dict[str, Any]):
         """Validate the form data."""
-
-    @field_validator("tooltip", mode="before")
-    @classmethod
-    def join_tooltip(cls, value: str | list[str]) -> str:
-        """
-        As tooltip is often written as a list of strings for readability,
-        this validator merges  them.
-
-        :param value: The tooltip string or list of strings to merge.
-
-        :return: The tooltip string.
-        """
-        if isinstance(value, list):
-            return "".join(value)
-        return value
 
 
 class StringForm(BaseForm):
@@ -555,13 +540,13 @@ class DataGroupForm(DataForm):
     data_group_type: GroupTypeEnum | list[GroupTypeEnum]
 
 
-class MultiDataGroupForm(BaseForm):
+class GroupMultiDataForm(BaseForm):
     """
     Geoh5py uijson form for data associated with a group.
 
-    Shares documented attributes with the BaseForm and GroupForm.
+    Shares documented attributes with the BaseForm.
 
-    :param group_value: The group UUID containing the data.
+    :param group_value: The group containing the data.
     :param group_type: List of group types that restricts the options in the
         Geoscience ANALYST ui.json dropdown.
     :param data_type: The data type, eg: 'Integer', 'Float', that filters the options
@@ -590,7 +575,7 @@ class MultiDataGroupForm(BaseForm):
         if not value or value == [""]:
             return ""
         if not isinstance(value, list):
-            value = [value]
+            raise TypeError(f"'value' must be a list of strings; got '{type(value)}'")
         return value
 
 
@@ -714,7 +699,7 @@ def get_mandatory_attributes(to_inspect: type[BaseForm]) -> set[str]:
 
 def indicator_attributes(
     parent: type[BaseForm], children: list[type[BaseForm]]
-) -> list[list[set[str]]]:
+) -> tuple[list[set[str]], list[set[str]]]:
     """
     List all the mandatory attributes defined in a subclass.
 
@@ -732,7 +717,6 @@ def indicator_attributes(
     parent_mandatory_attributes = get_mandatory_attributes(parent)
     parent_attributes = set(parent.model_fields)
 
-    # base on the data, which one are pertinent?
     full_differences = [
         set(child.model_fields) - parent_attributes for child in children
     ]
@@ -742,7 +726,7 @@ def indicator_attributes(
         for child in children
     ]
 
-    return [full_differences, mandatory_differences]
+    return full_differences, mandatory_differences
 
 
 def filter_candidates_by_indicator_polling(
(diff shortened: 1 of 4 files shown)


geoh5py/ui_json/forms.py
```diff
@@ -540,6 +555,45 @@ class DataGroupForm(DataForm):
...
+
+        :param value: The value to validate, which can be a string or a list of strings.
+
+        :return: A list of strings representing the value.
+        """
+        if not value or value == [""]:
+            return ""
+        if not isinstance(value, list):
+            value = [value]
+        return value
```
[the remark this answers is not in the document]
MatthieuCMira: nop

tests/geo_image_test.py
```diff
@@ -375,7 +375,7 @@ def test_georeference_image(tmp_path):
         assert isinstance(geoimage.image_georeferenced, Image.Image)
 
         # test grid2d errors
-        with pytest.raises(ValueError, match="image has wrong mode"):
+        with pytest.raises(ValueError, match="conversion from RGB to bidon"):
```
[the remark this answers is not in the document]
MatthieuCMira: add to correct that as it was a bug in the tests.
Probably due to Pillow version

geoh5py/ui_json/forms.py
```diff
@@ -161,6 +161,21 @@ def flatten(self):
...
+    @classmethod
+    def join_tooltip(cls, value: str | list[str]) -> str:
+        """
+        As tooltip is often written as a list of strings for readability,
+        this validator merge them.
+
+        :param value: The tooltip string or list of strings to merge.
+
+        :return: The tooltip string.
+        """
```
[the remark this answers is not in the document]
MatthieuCMira: the tests is already in the MultiGroup test I added

geoh5py/ui_json/forms.py
```diff
@@ -675,9 +775,27 @@ def filter_candidates_by_indicator_polling(
...
+    3. return the difference between the two counts for each subclass.
+
+    :param indicators: A list of sets of indicator attributes for each subclass.
+    :param data: The form data to check for matching indicators.
+
+    :return: An array of counts of matching indicators for each subclass.
+    """
+    count_intersection = np.array([len(i.intersection(data)) for i in indicators[0]])
+    count_difference = np.array([len(i.difference(data)) for i in indicators[1]])
+    return count_intersection - count_difference
```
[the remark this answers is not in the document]
MatthieuCMira: count intersection  => get all the possibilities. 
There might be some forms equals if they contains the different possibilities

count_difference => a valid indicator should be 0, so any invalid will have a minus n

geoh5py/ui_json/forms.py
```diff
@@ -161,6 +161,21 @@ def flatten(self):
...
+    def join_tooltip(cls, value: str | list[str]) -> str:
+        """
+        As tooltip is often written as a list of strings for readability,
+        this validator merges  them.
+
+        :param value: The tooltip string or list of strings to merge.
+
+        :return: The tooltip string.
+        """
+        if isinstance(value, list):
```
benk-mira: Should we account for whitespace?

benk-mira: What about the serialization?  Can we go back to a list?

geoh5py/ui_json/forms.py
```diff
@@ -540,6 +555,45 @@ class DataGroupForm(DataForm):
     data_group_type: GroupTypeEnum | list[GroupTypeEnum]
 
 
+class MultiDataGroupForm(BaseForm):
+    """
+    Geoh5py uijson form for data associated with a group.
+
+    Shares documented attributes with the BaseForm and GroupForm.
```
benk-mira:
```suggestion
    Shares documented attributes with the BaseForm and BaseForm.
```

geoh5py/ui_json/forms.py
```diff
@@ -540,6 +555,45 @@ class DataGroupForm(DataForm):
     data_group_type: GroupTypeEnum | list[GroupTypeEnum]
 
 
+class MultiDataGroupForm(BaseForm):
+    """
+    Geoh5py uijson form for data associated with a group.
+
+    Shares documented attributes with the BaseForm and GroupForm.
+
+    :param group_value: The group UUID containing the data.
```
benk-mira:
```suggestion
    :param group_value: The group containing the data.
```

geoh5py/ui_json/forms.py
```diff
@@ -540,6 +555,45 @@ class DataGroupForm(DataForm):
...
+    Geoh5py uijson form for data associated with a group.
+
+    Shares documented attributes with the BaseForm and GroupForm.
+
+    :param group_value: The group UUID containing the data.
+    :param group_type: List of group types that restricts the options in the
+        Geoscience ANALYST ui.json dropdown.
+    :param data_type: The data type, eg: 'Integer', 'Float', that filters the options
+        in the Geoscience ANALYST ui.json dropdown.
+    :param value: The list of data names or the name of the data.
```
benk-mira: Kris writes a list of names, not UIDs?

geoh5py/ui_json/forms.py
```diff
@@ -540,6 +555,45 @@ class DataGroupForm(DataForm):
...
+        """
+        Validate that value is a list, converting it if it's a string.
+
+        :param value: The value to validate, which can be a string or a list of strings.
+
+        :return: A list of strings representing the value.
+        """
+        if not value or value == [""]:
+            return ""
+        if not isinstance(value, list):
```
benk-mira, line 585: What about serialization?  I tried to design uijson/forms so that a read then write would reproduce an identical uijson file.  I dealt with this on other muli-select forms by splitting the form type into a regular and 'multi' class as it cuts down on validators/logic (see FileForm/MultiFileForm for example).

geoh5py/ui_json/forms.py
```diff
@@ -598,7 +652,14 @@ def only_multi_select(cls, value: bool) -> bool:
 
     @field_validator("value", mode="before")
     @classmethod
-    def to_list(cls, value):
+    def to_list(cls, value: str | list[str]) -> list[str]:
```
benk-mira, line 648: I know that you're doing this to provide data to an agent, but can't an agent get all of that information from the type annotation alone?

geoh5py/ui_json/forms.py
```diff
@@ -661,9 +756,13 @@ def filter_candidates_by_indicator_polling(
     any ambiguity between non-unique indicators such as 'choice_list'
     and 'multi_select'.
 
+    :param data: The form data to check for matching indicators.
+
+        :return: An array of candidate subclasses with the most matching indicators.
     """
```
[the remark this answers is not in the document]
benk-mira: ^^^ this!

geoh5py/ui_json/forms.py
```diff
@@ -540,6 +555,45 @@ class DataGroupForm(DataForm):
     data_group_type: GroupTypeEnum | list[GroupTypeEnum]
 
 
+class MultiDataGroupForm(BaseForm):
```
benk-mira: Can we call this GroupMultiDataForm.  This makes it seem like it's a multi-select version of the DataGroupForm

geoh5py/ui_json/forms.py
```diff
@@ -638,16 +699,50 @@ def all_subclasses(type_object: type[BaseForm]) -> list[type[BaseForm]]:
...
+        key for key, value in to_inspect.model_fields.items() if value.is_required()
+    }
 
 
 def indicator_attributes(
     parent: type[BaseForm], children: list[type[BaseForm]]
-) -> list[set[str]]:
-    """List all the attributes defined in a subclass."""
-    return [model_fields_difference(parent, c) for c in children]
+) -> list[list[set[str]]]:
```
benk-mira: Can you justify these changes? I created a new branch off develop called 'test_2746', added your new form type, and fed it through the BaseForm.infer in the 'test_base_form_infer' unit test.  The old method seems to handle the new form just fine.  Can you add a unit test that shows the old 'infer' breaking on my branch to help me understand the need for the 'mandatory_differences' counting?

benk-mira (changes requested):
You'll have to show me how the extra 'mandatory' counting is necessary.  I tested the old infer with the new form type (see comment), and it seems to work without augmenting the infer.  Can you add a unit test to my 'test_2746' branch that fails?

geoh5py/ui_json/forms.py
MatthieuCMira: Why? white space should be deal on user's size

MatthieuCMira: We could.
But I don't think that's so pertinent. Lot of logic for not much interest. A tooltip is a tooltip

geoh5py/ui_json/forms.py
MatthieuCMira: yes, we can use name for types. 
Hopefully, it will be the case with objects and group soon!

geoh5py/ui_json/forms.py
MatthieuCMira: I tested, it will always be a list of string unless we have no value selected.

removing this part then

geoh5py/ui_json/forms.py
MatthieuCMira: Or we just let the tooltip being a list of string. That's a convenient possibility too. 
I'll do that

benk-mira: because it's easy to forget to add a space when creating multi-line strings.. not a huge deal - not blocking

geoh5py/ui_json/forms.py
MatthieuCMira: Docstrings are here to auto-generate the documentation.
This is why we must add them in every public method.

Still I agree repeating the typing is not necessary.

In pep, it's says "comment all public method/class/"

geoh5py/ui_json/forms.py
```diff
@@ -675,9 +758,27 @@ def filter_candidates_by_indicator_polling(
...
-    return np.array([len(i.intersection(data)) for i in indicators])
+def count_indicators(
+    indicators: tuple[list[set[str]], list[set[str]]], data: dict[str, Any]
+) -> np.ndarray:
+    """
+    Count the number of matching indicators for each child class.
+
+    1. count the number of mandatory indicators present in the form data for each subclass.
+        It allows to select the valid classes.
+    2. count the number of indicators not present in the form data for each
```
domfournier:
```suggestion
    1. Count the number of mandatory indicators present in the form data for each subclass.
        It allows to select the valid classes.
    2. Count the number of indicators not present in the form data for each
```
Grammar

geoh5py/ui_json/forms.py
benk-mira: I get the use of names for types, but this is the name of a data channel the user is entering.  @domfournier are we moving towards handling names for object/group/data instead of uids?

geoh5py/ui_json/forms.py
benk-mira: It's not public method though, it's a validator.  Validation methods should include a leading underscore i guess (my bad), otherwise autodocs will get cluttered with validators.

benk-mira: not blocking.. moving on

geoh5py/ui_json/forms.py
domfournier: This is specific to DrillholeGroups, where the selection of data from a table has to be done by name, since data belongs to multiple sub-Data entities.

We should add a validation on the group_type here, and only accept DrillholeGroups. Nothing else can hold property_groups

geoh5py/ui_json/forms.py
```diff
@@ -108,7 +108,7 @@ class BaseForm(BaseModel):
     optional: bool = False
     enabled: bool = True
     main: bool = True
-    tooltip: str = ""
+    tooltip: str | list[str] = ""
```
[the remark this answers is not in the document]
MatthieuCMira: used in InputFile. shouild be deleted soon

geoh5py/ui_json/forms.py
benk-mira: Ok, didn't realize this was DrillholeGroup specific, let's give the form a different name then `DrillholeDataForm`? and update the docstrings

geoh5py/ui_json/forms.py
```diff
@@ -661,9 +740,13 @@ def filter_candidates_by_indicator_polling(
     any ambiguity between non-unique indicators such as 'choice_list'
     and 'multi_select'.
 
+    :param data: The form data to check for matching indicators.
+
+    :return: An array of candidate subclasses with the most matching indicators.
     """
     counts = count_indicators(INDICATORS, data)
```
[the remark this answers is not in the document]
MatthieuCMira: This templates could be supress no @domfournier ?

geoh5py/ui_json/forms.py
MatthieuCMira: Are we 100% sure about that?
Then why a GroupType would be needed?

I agree I'm using it for Drillholegroup, but maybe it could work with another group!?

What Kris created on his side?
We should ask him!

geoh5py/ui_json/forms.py
```diff
@@ -540,6 +540,45 @@ class DataGroupForm(DataForm):
     data_group_type: GroupTypeEnum | list[GroupTypeEnum]
 
 
+class GroupMultiDataForm(BaseForm):
+    """
+    Geoh5py uijson form for data associated with a group.
```
benk-mira:
```suggestion
    Geoh5py uijson form for selecting (multi) data within a drillhole property group.
```

geoh5py/ui_json/forms.py
```diff
@@ -540,6 +540,45 @@ class DataGroupForm(DataForm):
     data_group_type: GroupTypeEnum | list[GroupTypeEnum]
 
 
+class GroupMultiDataForm(BaseForm):
+    """
+    Geoh5py uijson form for data associated with a group.
+
+    Shares documented attributes with the BaseForm.
+
+    :param group_value: The group containing the data.
```
benk-mira:
```suggestion
    :param group_value: The property group containing the data.
```

geoh5py/ui_json/forms.py
```diff
@@ -540,6 +540,45 @@ class DataGroupForm(DataForm):
...
 
 
+class GroupMultiDataForm(BaseForm):
+    """
+    Geoh5py uijson form for data associated with a group.
+
+    Shares documented attributes with the BaseForm.
+
+    :param group_value: The group containing the data.
+    :param group_type: List of group types that restricts the options in the
```
benk-mira, line 554:
```suggestion
    :param group_type: Must be 'PropertyGroup'.
```

geoh5py/ui_json/forms.py
```diff
@@ -540,6 +540,45 @@ class DataGroupForm(DataForm):
...
+        Geoscience ANALYST ui.json dropdown.
+    :param data_type: The data type, eg: 'Integer', 'Float', that filters the options
+        in the Geoscience ANALYST ui.json dropdown.
+    :param value: The list of data names or the name of the data.
```
benk-mira:
```suggestion
    :param value: The list of data names or the name of the data stored across the drillholes.
```

geoh5py/ui_json/forms.py
```diff
@@ -540,6 +540,45 @@ class DataGroupForm(DataForm):
...
+    :param multi_select: If True, the ui.json dropdown will allow for multi-selection.
+    """
+
+    group_type: GroupTypes
```
benk-mira, line 562:
```suggestion
    group_type: type(PropertyGroup) = PropertyGroup
```

geoh5py/ui_json/forms.py
```diff
@@ -540,6 +540,45 @@ class DataGroupForm(DataForm):
     data_group_type: GroupTypeEnum | list[GroupTypeEnum]
 
 
+class GroupMultiDataForm(BaseForm):
```
benk-mira, line 543:
```suggestion
class DrillholeDataForm(BaseForm):
```

benk-mira (changes requested):
Some suggestions to account for the fact that the form is only used with DrillholeGroup/PropertyGroup

geoh5py/ui_json/forms.py
benk-mira: Seems like it.  I tried with ContainerGroup and the ui doesn't play along:
[test_uijson.ui.json](https://github.com/user-attachments/files/25918797/test_uijson.ui.json)
There's nothing in the dropdown:

image

geoh5py/ui_json/forms.py
MatthieuCMira: See my previous comment.

Can't we imageine this Form on other groupType.

Imagine an InterpretationSectionGroup
We have curves int it with data on them, per lithology. 
We might want to select them this way right?

geoh5py/ui_json/forms.py
MatthieuCMira: no, it's not a property group.
It could be a DrillholeGroup, an InterpretationSectionGroup and so on right?

geoh5py/ui_json/forms.py
MatthieuCMira: I'll add a note to say it's working on drillholes for now but could be extended

PR #855, opened 2026-03-11
GEOPY-2746-unit-test-fill

benk-mira:


tests/ui_json/forms_test.py
```diff
@@ -724,31 +701,63 @@ def test_base_form_infer(tmp_path):
...
+            "dataGroupType": "Strike & dip",
+        }
     )
-    assert form == DataRangeForm
-
+    assert form == DataGroupForm
+    form = BaseForm.infer(
+        {
+            "label": "test",
+            "value": str(uuid.uuid4()),
```
MatthieuCMira: in MulitDataForm, value should NOT be a uid but a string refering to the data.

This form was made for the drillhole group, and there is one data per drillhole

MatthieuCMira (changes requested):
It's not blocking the test, but the value of a GroupMultiData should be the dataname, not the uid

tests/ui_json/forms_test.py
benk-mira: gotcha, fixed

PR #860, opened 2026-03-21
GEOPY-2739: Accept BaseUIJson in the start of driver

domfournier:
**GEOPY-2739 - Accept BaseUIJson in the start of driver**


geoh5py/ui_json/ui_json.py
```diff
@@ -114,20 +119,56 @@ def valid_geoh5_extension(cls, path: Path | None) -> Path | None:
...
+                params,
+                parent,
+                clear_cache=clear_cache,
+            )
+
+    @staticmethod
+    def infer(title="UnknownUIJson", **kwargs) -> type[UIJson]:
+        """
+        Create a UIJson class based on inferred forms.
         """
```
MatthieuCMira, line 170: docstrings

geoh5py/shared/utils.py
```diff
@@ -945,7 +924,7 @@ def nan2str(value):
 
 
 def str2none(value):
```
MatthieuCMira: typing and docstrings

MatthieuCMira (changes requested):
ok with the changes.

only typical docstrings comments

Still, that's a lot of function's names changes. Isn't it a bit dangerous to add it to 4.8?

geoh5py/ui_json/annotations.py
```diff
@@ -47,55 +52,61 @@ def deprecate(value, info):
...
+]
+
+OptionalString = Annotated[
+    str | None,
+    BeforeValidator(str2none),
+    PlainSerializer(none2str, when_used="json"),
 ]
 
 OptionalUUID = Annotated[
     UUID | None,
```
MatthieuCMira, line 101: This is from a past merge, but when multiselect is True (eg. in DataForm), we should be able to accept a list of UUIDs

MatthieuCMira (changes requested):
check for the Form accepting multiselect, and accept list of expected values

geoh5py/ui_json/validation.py
```diff
@@ -360,71 +363,98 @@ def throw(self):
...
+    form = getattr(ui_json, name)
+    mesh_types = form.mesh_type
 
     obj = data[name]
     if not isinstance(obj, tuple(mesh_types)):
         raise UIJsonError(f"Object's mesh type must be one of {mesh_types}.")
 
 
-def parent_validation(name: str, data: dict[str, Any], json_dict: dict[str, Any]):
+def parent_validation(name: str, data: dict[str, Any], ui_json: UIJson):
```
MatthieuCMira, line 412: this gave me an error because if child is in list, it raises an error!

geoh5py/ui_json/validation.py
```diff
@@ -360,71 +363,98 @@ def throw(self):
...
+    parent_name = form.parent
     child = data[name]
-    parent = data[form["parent"]]
+    parent = data[parent_name]
 
-    if not isinstance(parent, ObjectBase) or (
-        isinstance(child, Data) and parent.get_entity(child.uid)[0] is None
-    ):
-        raise UIJsonError(f"{name} data is not a child of {form['parent']}.")
+    if not isinstance(parent, ObjectBase) or (child not in parent.children):
```
MatthieuCMira:
```suggestion
        child = child if isinstance(child, list) else [child]
    missing_children = len(list(set(child) - set(parent.children))) > 0
    if not isinstance(parent, ObjectBase) or missing_children:
```

geoh5py/ui_json/annotations.py
domfournier: Have to be careful because some forms don't allow a list. For example, the DataOrValueForm can only accept single UUID, no list.

domfournier:
@copilot

MatthieuCMira (commented):
give me a new error:

Image

pulling the string:

- 1. because, in form dependency, if a dependency is set latter in the form, it's not in the dictionary you created. 
- 2. because dependencies comes from non active fields (use flatten active_only=False)

geoh5py/ui_json/forms.py
```diff
@@ -536,6 +542,27 @@ def to_list(cls, value: str | list[str]) -> str | list[str]:
...
         return value
 
+    def flatten(self) -> dict:
+        """Returns the property, data and is_complement values for the form."""
+        return {
+            "group_value": self.group_value,
+            "value": self.value,
+        }
+
+    def set_value(self, value: Any):
```
MatthieuCMira, line 552: in set_value.
Should we set enabled to true when we do the set_value?

geoh5py/ui_json/forms.py
```diff
@@ -571,20 +598,22 @@ def flatten(self) -> UUID | float | int | None:
         return self.value
 
     def set_value(self, value: Any):
```
MatthieuCMira, line 600: same

geoh5py/ui_json/forms.py
MatthieuCMira: it could be set_value(None)
=> set_enabled to False?

geoh5py/ui_json/ui_json.py
```diff
@@ -61,235 +64,274 @@ class BaseUIJson(BaseModel):
...
+            enabled = self._group_dependencies[group].enabled
 
-    def flatten(self, skip_disabled=False, active_only=False) -> dict[str, Any]:
+        # Then check on direct dependency
+        for name, mirror in self._form_dependencies[field].items():
+            # Not enabled as soon as False is encountered
+            if not enabled:
+                return False
+
+            codependent = getattr(self, name)
```
MatthieuCMira, line 218: yep better than my 3 dictionaries. :-)

geoh5py/ui_json/ui_json.py
```diff
@@ -61,235 +64,274 @@ class BaseUIJson(BaseModel):
...
-        for key, value in kwargs.items():
-            form = getattr(uijson, key, None)
+        for field, value in kwargs.items():
+            form = getattr(uijson, field, None)
             if isinstance(form, BaseForm):
-                form.set_value(value)
+                if not (form.is_optional and value is None):
+                    form.set_value(value)
+
+                self.set_enabled(copy=False, **{field: value is not None})
```
MatthieuCMira, line 309: oh ok, you set it here to deal with dependencies

geoh5py/ui_json/forms.py
MatthieuCMira: ok, my bad you do it in `set_values`

geoh5py/ui_json/forms.py
MatthieuCMira: ok, my bad you do it in `set_values`

geoh5py/ui_json/ui_json.py
```diff
@@ -342,11 +382,39 @@ def to_ui_json_group(
 
             return ui_json_group
 
-    def validate_data(
-        self, params: dict[str, Any] | None = None, errors: dict[str, Any] | None = None
+    @field_validator("geoh5", mode="after")
+    @classmethod
+    def valid_geoh5_extension(cls, path: Path | None) -> Path | None:
```
MatthieuCMira, line 388: docstrings

geoh5py/ui_json/ui_json.py
```diff
@@ -342,11 +382,39 @@ def to_ui_json_group(
...
+        if path is not None and path.suffix != ".geoh5":
+            raise ValueError(
+                f"Workspace path: {path} must have a '.geoh5' file extension."
+            )
+        return path
+
+    @field_validator("geoh5", mode="after")
+    @classmethod
+    def workspace_path_exists(cls, path: Path | None) -> Path | None:
```
MatthieuCMira, line 403: docstrings

geoh5py/ui_json/ui_json.py
```diff
@@ -342,11 +382,39 @@ def to_ui_json_group(
...
+        if path is not None and not path.exists():
+            raise FileNotFoundError(f"geoh5 path {path} does not exist.")
+        return path
+
+    def write(self, path: Path) -> Path:
+        """
+        Write the UIJson object to file.
+
+        :param path: Path to write the .ui.json file.
```
MatthieuCMira, line 418: :return:  ...

geoh5py/ui_json/ui_json.py
```diff
@@ -355,46 +423,113 @@ def validate_data(
...
+            # If optional, enabled state only influences the form
+            if dependents_on and not getattr(form, "optional", False):
+                mirrors = getattr(form, "dependency_type", None) in [
+                    DependencyType.ENABLED,
+                    DependencyType.SHOW,
+                ]
+
+                # Add reverse linkage
+                if dependents_on not in form_dependencies:
+                    form_dependencies[dependents_on] = {}
```
MatthieuCMira, line 505: This will be overwrite during the loop by l462

geoh5py/ui_json/ui_json.py
```diff
@@ -355,46 +423,113 @@ def validate_data(
...
+        A direct dependency controls the enabled state tow ways, while the group dependency
+        controls the enabled state only from the lead form to its dependents.
+
+        :returns: Tuple of group dependencies and direct form dependencies.
         """
+        form_dependencies: dict[str, dict[str, bool]] = {}
+        group_dependencies: dict[str, BaseForm] = {}
+
+        for name in self.__class__.model_fields.keys():
+            form_dependencies[name] = {}
```
MatthieuCMira: Or you will overwrite l489

```suggestion

            if name not in form_dependencies:
                     form_dependencies[name] = {}
```

geoh5py/ui_json/validation.py
```diff
@@ -360,71 +363,111 @@ def throw(self):
...
-    dependency = json_dict[name]["dependency"]
-    dependency_form = json_dict[dependency]
+    form = getattr(ui_json, name)
+    dependency = form.dependency
+    dependency_form = getattr(ui_json, dependency)
 
-    if "optional" not in dependency_form and not isinstance(data[dependency], bool):
+    # note: if the data is unset, it does not appear in the data dictionary
+    # so it only works with dependency disabled
+    # should we do another check => if dependency enabled and False raises error?
```
MatthieuCMira: What about my comment, 
if it's resolve, remove it. 
If not, what we do?

MatthieuCMira (commented):
Some changes to tackl.e
Good job almost here!

geoh5py/ui_json/ui_json.py
domfournier: Right, do all the checks at once.

geoh5py/ui_json/validation.py
domfournier: Yes thanks, the second logic doesn't make sense to me. I will remove for now.

domfournier: Ok yes, this has to do with a BoolForm. It can be used as a dependency without itself beiung optional.
I changed the logic to not depende on data.

geoh5py/ui_json/ui_json.py
```diff
@@ -61,235 +64,274 @@ class BaseUIJson(BaseModel):
...
+
+            if mirror:
+                enabled = codependent.enabled
+            else:
+                enabled = not codependent.enabled
+
+        return enabled
+
+    def model_post_init(self, context: Any, /) -> None:
+        self._group_dependencies, self._form_dependencies = self._get_dependency_links()
```
MatthieuCMira, line 228: Can you explain what that is?
The '/'
Maybe add docstrings?

geoh5py/ui_json/validation.py
```diff
@@ -360,71 +364,108 @@ def throw(self):
             raise UIJsonError(message)
 
 
-def dependency_type_validation(
-    name: str, data: dict[str, Any], json_dict: dict[str, Any]
-):
+def dependency_type_validation(name: str, _, ui_json: UIJson):
```
MatthieuCMira: Question:

is this function used for something else than ui_json?
Maybe data could just be removed from its calls instead of "_"?
It strange to have it as a positional argument this way no?

MatthieuCMira (commented):
Some question but seems good to me. 
Let's consider it as approved.
I'll just wait for your answer before doing the tests

geoh5py/ui_json/validation.py
domfournier: The current design calls all the data validation functions on promoted entities in a loop, and passes the same arguments to all. So we would need a positional for this one too. 

That being said, your comment indicates that this is not really a validation for the outputs, but more a validation of the forms that should be done at the onset by the UIJson class. I will move.

geoh5py/ui_json/ui_json.py
domfournier: I did not write that part and hadn't noticed. I can remove

domfournier: Ok so that is just how the base class defines the post_init, forcing the position arguments with /

image

PR #868, opened 2026-03-30
GEOPY-2762: Add mechanisms to update the UIJson class with values

domfournier:
**GEOPY-2762 - Add mechanisms to update the UIJson class with values**


geoh5py/ui_json/forms.py
```diff
@@ -90,10 +90,10 @@ class BaseForm(BaseModel):
 
     model_config = ConfigDict(
         extra="allow",
-        frozen=True,
         populate_by_name=True,
         loc_by_alias=True,
         alias_generator=to_camel,
+        validate_assignment=True,
```
MatthieuCMira, line 96: yes! I like that!

geoh5py/ui_json/forms.py
```diff
@@ -556,17 +561,24 @@ def property_if_not_is_value(self):
...
             return self.property
         return self.value
 
+    def set_value(self, value: Any):
+        """Set the form value."""
+        try:
+            self.value = value
+            self.is_value = True
+        except ValidationError:
+            self.is_value = value is None
```
MatthieuCMira: if value is None (so is_value True and no values nor properties) 
Should we no set enabled to False if optional?

geoh5py/ui_json/ui_json.py
```diff
@@ -96,7 +88,7 @@ def __str__(self) -> str:
         """String level shows the full json representation."""
 
         json_string = self.model_dump_json(indent=4, exclude_unset=True)
-        for field in type(self).model_fields:
+        for field in type(self).model_fields.keys():
```
MatthieuCMira: why adding keys? model_fields is a dict right? 
When you loop on a dict, you always get the keys.

MatthieuCMira: (same comment everywhere esle)

MatthieuCMira (changes requested):
only major comment is the set_value of dataorvalue form

geoh5py/ui_json/forms.py
domfournier: My thought was that the `value` in this case is always a float, so `set_value` receiving None implies that the `property` is not used and we revert to is_value. For this specific form, we basically have to set the `enable` state directly if it's optional, can't safely do it through the value setter.

domfournier:
@copilot

MatthieuCMira (approved):
good for me!

geoh5py/ui_json/forms.py
MatthieuCMira: like it better this way!

PR #876, opened 2026-04-21
GEOPY-2816: data_map.entity_type.value_map can have map is None leading to error

domfournier:
**GEOPY-2816 - data_map.entity_type.value_map can have map is None leading to error**
… to match value_map

### DIFF
--- a/geoh5py/data/data_type.py
+++ b/geoh5py/data/data_type.py
@@ -23,6 +23,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from copy import copy
 from typing import TYPE_CHECKING, Literal, get_args
 from uuid import UUID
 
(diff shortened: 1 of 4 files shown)


geoh5py/data/data_type.py
```diff
@@ -681,6 +682,18 @@ def __init__(
         )
         self._parent = parent
 
+    def set_parent_reference(self, data: Data, new_name: str):
```
MatthieuCMira, line 685: docstrings

MatthieuCMira (changes requested):
Good news: it's working and I tested it. 

Just one docstring missing

MatthieuCMira (approved):
Ok for me, go to test. 
Just for my information, could you explained me what was happening and how you tackled it? It's a bit obscure to me....

PR #893, opened 2026-05-12
GEOPY-2860: Accept form dependency type based on group_optional

domfournier:
**GEOPY-2860 - Accept form dependency type based on group_optional**


MatthieuCMira (approved):
good for me.

I already tested it with GEOPY-2841 issue

PR #904, opened 2026-05-26
GEOPY-2815: address report of updated code linters

domfournier:
**GEOPY-2815 - address report of updated code linters**


MatthieuCMira (dismissed):
ok

PR #911, opened 2026-06-02
GEOPY-2893: h5 compression level seems ignore (from mira-omf)

domfournier:
**GEOPY-2893 - h5 compression level seems ignore (from mira-omf)**

### DIFF
--- a/geoh5py/io/h5_writer.py
+++ b/geoh5py/io/h5_writer.py
@@ -603,6 +603,8 @@ class H5Writer:
         :param file: Name or handle to a geoh5 file.
         :param entity: Target entity.
         :param attribute: Name of the attribute to be written to geoh5
+        :param values: Values to be written to geoh5
+        :param compression: Compression level `compression_opts` of the h5 file.
         """
         with fetch_h5_handle(file, mode="r+") as h5file:
             entity_handle = H5Writer.fetch_handle(h5file, entity)
--- a/geoh5py/io/h5_writer.py
+++ b/geoh5py/io/h5_writer.py
@@ -354,7 +354,9 @@ class H5Writer:
                 "vertices",
                 "z_cell_delimiters",
             ]:
-                H5Writer.write_array_attribute(h5file, entity, attribute, **kwargs)
+                H5Writer.write_array_attribute(
+                    h5file, entity, attribute, compression=compression, **kwargs
+                )
             elif attribute == "property_groups":
                 H5Writer.write_property_groups(h5file, entity)
             elif attribute == "color_map":
@@ -588,7 +590,12 @@ class H5Writer:
 
     @staticmethod
     def write_array_attribute(
-        file: str | h5py.File, entity, attribute, values=None, **kwargs
+        file: str | h5py.File,
+        entity,
+        attribute,
+        values=None,
+        compression: int = 9,
+        **kwargs,
     ) -> None:
         """
         Add :obj:`~geoh5py.objects.object_base.ObjectBase.surveys` of an object.
@@ -624,7 +631,7 @@ class H5Writer:
                     KEY_MAP[attribute],
                     data=values,
                     compression="gzip",
-                    compression_opts=9,
+                    compression_opts=compression,
                     **kwargs,
                 )
 
--- a/geoh5py/shared/utils.py
+++ b/geoh5py/shared/utils.py
@@ -46,7 +46,8 @@ if TYPE_CHECKING:
     from .entity import Entity
     from .entity_container import EntityContainer
 
-DEFAULT_PAGE_SIZE = 65536
+DEFAULT_PAGE_SIZE = 65536  # Default page_size for H5, in bytes
+
 INV_KEY_MAP = {
     "Allow delete": "allow_delete",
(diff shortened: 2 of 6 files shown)


geoh5py/io/h5_writer.py
```diff
@@ -624,7 +631,7 @@ def write_array_attribute(
                     KEY_MAP[attribute],
                     data=values,
                     compression="gzip",
-                    compression_opts=9,
+                    compression_opts=compression,
                     **kwargs,
                 )
```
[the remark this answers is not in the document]
domfournier: Tested in workspace_test.test_compression_below_page

geoh5py/workspace/workspace.py
```diff
@@ -1605,3 +1617,20 @@ def active_workspace(workspace: Workspace):
...
+    Check if a page size is valid value.
+
+    :param value: A positive integer multiple of 2, >=512.
+
+    :return: Page size value
+    """
+    if not isinstance(value, int):
+        raise TypeError("Page size must be an integer.")
+
+    if value // 512 < 1 or value % 2 != 0:
```
sebhmg: **suggestion:** easier read as comparison integer division and comparison to 1
```suggestion
    if value < 512 or value % 2 != 0:
```

geoh5py/workspace/workspace.py
```diff
@@ -1605,3 +1617,20 @@ def active_workspace(workspace: Workspace):
     previous_active = previous_active_ref()
     if previous_active is not None:
         previous_active.activate()  # pylint: disable=no-member
+
+
+def validate_page_size(value: int) -> int:
```
sebhmg, line 1627: **question:** why a dedicate function instead on validating in the property setter

geoh5py/workspace/workspace.py
```diff
@@ -1605,3 +1617,20 @@ def active_workspace(workspace: Workspace):
...
+    """
+    Check if a page size is valid value.
+
+    :param value: A positive integer multiple of 2, >=512.
+
+    :return: Page size value
```
sebhmg:
```suggestion
    Check if a page size is valid value. Raise an error if not valid, alse return the value as-is.
    :param value: A positive integer multiple of 2, >=512.
    :return: Page size value, same as :param value:
```

geoh5py/workspace/workspace.py
```diff
@@ -152,7 +154,7 @@ def __init__(
         self._repack: bool = repack
         self._types: dict[uuid.UUID, ReferenceType[EntityType]] = {}
         self._version: float = version
-
+        self._page_size: int = validate_page_size(page_size)
```
sebhmg, line 157: **suggestion:** use a setter, in charge of the validation
```suggestion
        self.page_size = page_size
```

geoh5py/shared/utils.py
```diff
@@ -46,7 +46,7 @@
     from .entity import Entity
     from .entity_container import EntityContainer
 
-DEFAULT_PAGE_BUF_SIZE = 16777216
+DEFAULT_PAGE_SIZE = 65536
```
sebhmg: **suggestion:** add a docstring telling which unit this is

```suggestion
DEFAULT_PAGE_SIZE = 65536
"""int: default page size for H5, in bytes"""

```

geoh5py/workspace/workspace.py
```diff
@@ -1303,13 +1305,23 @@ def _create_h5(self) -> h5py.File:
             self.h5file,
             "x",
             fs_strategy="page",
-            page_buf_size=DEFAULT_PAGE_BUF_SIZE,
+            page_buf_size=self._page_size * 256,
+            fs_page_size=self._page_size,
```
sebhmg, line 1314: **note:** some explanation, for posterity (as I had to look for it)

Both `page_buf_size` and `fs_page_size` expects value in bytes.
Here, the page buffer will be able to hold 256 pages.

geoh5py/workspace/workspace.py
```diff
@@ -115,6 +115,7 @@ class Workspace(AbstractContextManager):
     :param name: Name of the project.
     :param repack: Repack the *geoh5* file after closing.
     :param version: Version of the project.
+    :param page_size: Page size of the h5 file.
```
sebhmg:
```suggestion
    :param page_size: Page size of the h5 file, in bytes.
```

sebhmg (changes requested):
logic looks correct. Nice coverage with tests.
See some questions and suggestions

geoh5py/workspace/workspace.py
domfournier: I don't think we want users to change the `page_size` after instantiation (can't change the file after creation anyway). So it could be a setter, but would have to check if set already, making the `_page_size = None` at first (akward typing). 

Having a dedicated function with fixed return type let us set the value and protects the @property page_size.

geoh5py/workspace/workspace.py
domfournier: Answered above

sebhmg (approved):

PR #931, opened 2026-06-30
GEOPY-2924: Support label text (no value) form type

domfournier:
**GEOPY-2924 - Support label text (no value) form type**

### DIFF
--- a/geoh5py/ui_json/forms.py
+++ b/geoh5py/ui_json/forms.py
@@ -170,7 +170,8 @@ class LabelForm(BaseForm):
     Shares documented attributes with the BaseForm.
     """
 
-    value: None = None
+    value: None = None
+
 
 class StringForm(BaseForm):
     """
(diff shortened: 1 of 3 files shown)


sebhmg (approved):

geoh5py/ui_json/forms.py
```diff
@@ -163,6 +163,16 @@ def is_optional(self) -> bool:
         return self.optional or self.group_optional or len(self.dependency) > 0
 
 
+class LabelForm(BaseForm):
```
benk-mira, line 166: Are we sure a LabelForm can inherit all the attributes of the BaseForm?  What does analyst do with label/dependencies/optional/etc..?

sebhmg: I believe so. The code of Analyst is not making a special case for this one: visible conditions, etc. should all apply

PR #940, opened 2026-07-07
GEOPY-2991: Add label form to the plate match uijson for download location

domfournier:
**GEOPY-2991 - Add label form to the plate match uijson for download location**

### DIFF
--- a/geoh5py/ui_json/forms.py
+++ b/geoh5py/ui_json/forms.py
@@ -96,7 +96,7 @@ class BaseForm(BaseModel):
 
     __pydantic_extra__: dict[str, str | float | int]  # For autodoc
 
-    label: str
+    label: str | list[str]
     value: Any
     optional: bool = False
     enabled: bool = True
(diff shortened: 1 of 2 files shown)


geoh5py/ui_json/forms.py
```diff
@@ -409,13 +409,6 @@ class DirectoryForm(BaseForm):
...
-    @field_validator("value")
-    @classmethod
-    def valid_directory(cls, value: Path) -> Path:
-        if not value.exists() or not value.is_dir():
-            raise ValueError(f"Provided path {value} is not a valid directory.")
-        return value
-
     @field_validator("directory_only", mode="before")
     @classmethod
     def force_directory_only(cls, _):
```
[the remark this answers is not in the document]
domfournier: Not the responsibility of UIJson, but Options class. Possible that Path is to be created by the driver

geoh5py/ui_json/forms.py
benk-mira:
```suggestion
    :param label: Label(s) for ui element.  A list of labels will be rendered as a single space separated string.
```

benk-mira: Does this have a different meaning for the LabelForm?

benk-mira (changes requested):

geoh5py/ui_json/forms.py
domfournier: no, same behaviour for all

PR #941, opened 2026-07-08
GEOPY-2944: POC: Instantiate Points object as pydantic class

gmcga:
**GEOPY-2944 - POC: Instantiate Points object as pydantic class**


geoh5py_pydantic/points.py
```diff
@@ -0,0 +1,195 @@
...
+    type_uid: UUID = POINTS_TYPE_UID
+    name: str = "Points"
+    last_focus: str = Field(
+        default="None", validation_alias=AliasChoices("last_focus", "Last focus")
+    )
+    vertices: np.ndarray | LazyArray = Field(
+        default=None,
+        validate_default=True,
+        validation_alias=AliasChoices("vertices", "Vertices"),
+    )
```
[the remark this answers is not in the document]
domfournier: Correct - Points should always have vertices, so we either use a default like current `(0.0, 0.0, 0.0)`, or remove the default to force values to be provided

geoh5py_pydantic/arrays.py
```diff
@@ -0,0 +1,162 @@
...
+
+# Purpose of lazy loading:
+# let PointsModel hold vertices without necessarily loading them into memory right away.
+# Maybe also allow for accessing/modifying only a specific section of the array.
+
+
+ArrayValidator = Callable[[np.ndarray], np.ndarray]
+
+
+class ArraySource(Protocol):  # pylint: disable=R0903
```
domfournier:
```suggestion
class ArraySource(Protocol):  # pylint: disable=too-few-public-methods
```

geoh5py_pydantic/arrays.py
```diff
@@ -0,0 +1,162 @@
...
+        self.uid = uid
+        self.key = key
+        self.validator = validator
+        self._value = self._validate(value) if value is not None else None
+
+    def _validate(self, value: np.ndarray) -> np.ndarray:
+        if self.validator is None:
+            return value
+
+        return self.validator(value)
```
domfournier:
```suggestion
        self._validate(value)

    def _validate(self, value: np.ndarray | None) -> np.ndarray:
        self.value = self.validator(value)
```

geoh5py_pydantic/arrays.py
```diff
@@ -0,0 +1,162 @@
...
+    array conveniences, while keeping ownership of IO outside the pydantic model.
+    """
+
+    def __init__(
+        self,
+        source: ArraySource,
+        uid: UUID,
+        key: str,
+        *,
+        validator: ArrayValidator | None = None,
```
domfournier:
```suggestion
        validator: list[ArrayValidator],
```
I think it is fair that we will always have a validator - even if it ends up just being a type checker. We could make this a list such that `with_validator` can attach more to it? 
For example, the base validator for FloatArray would check for the `TypeValidation`, then the parent could attach a `ShapeValidation`?

geoh5py_pydantic/arrays.py
```diff
@@ -0,0 +1,162 @@
...
+        return self.fetcher(uid, key)
+
+
+class LazyArray:
+    """
+    Array proxy that loads and validates only when accessed.
+
+    The proxy is small, it provides numpy coercion and a few common
+    array conveniences, while keeping ownership of IO outside the pydantic model.
+    """
```
domfournier, line 88: Can you docstring what those attributes are? I am actually not sure what `uid` and `key` would be used for...

geoh5py_pydantic/arrays.py
```diff
@@ -0,0 +1,162 @@
...
+        return self
+
+    @property
+    def is_loaded(self) -> bool:
+        """
+        Whether the array has been fetched into memory.
+        """
+        return self._value is not None
+
+    def load(self) -> np.ndarray:
```
domfournier:
```suggestion
    @property
    def value(self) -> np.ndarray:
```

geoh5py_pydantic/arrays.py
```diff
@@ -0,0 +1,162 @@
...
+                    f"Array '{self.key}' for entity '{self.uid}' could not be loaded."
+                )
+
+            self._value = self._validate(value)
+
+        return self._value
+
+    @property
+    def shape(self) -> tuple[int, ...]:
+        return self.load().shape
```
domfournier:
```suggestion
        return self.value.shape
```

geoh5py_pydantic/arrays.py
```diff
@@ -0,0 +1,162 @@
...
+            if value is None:
+                raise ValueError(
+                    f"Array '{self.key}' for entity '{self.uid}' could not be loaded."
+                )
+
+            self._value = self._validate(value)
+
+        return self._value
+
+    @property
```
domfournier:
```suggestion
        return self._value
    
    @value.setter
    def value(self, value):
        self._value = self.validator(value)    
    
    @property
```

geoh5py_pydantic/arrays.py
domfournier: I think it would look cleaner if we use a property with a setter, instead of always calling a function load?

geoh5py_pydantic/entity.py
```diff
@@ -0,0 +1,146 @@
...
+
+    model_config = ConfigDict(
+        arbitrary_types_allowed=True,
+        extra="allow",
+        populate_by_name=True,
+        validate_assignment=True,
+    )
+
+    uid: UUID = Field(default_factory=uuid4, validation_alias=AliasChoices("uid", "ID"))
+    name: str = Field(default="Entity", validation_alias=AliasChoices("name", "Name"))
```
domfournier: Yes! Although I think we want 
```suggestion
    uid: UUID = Field(default_factory=uuid4, alias=AliasChoices("uid", "ID"))
    name: str = Field(default="Entity", alias=AliasChoices("name", "Name"))
```
such that the alias is used for the serialization as well.

geoh5py_pydantic/entity.py
```diff
@@ -0,0 +1,146 @@
...
+                "Input metadata must be of type dict, bytes, str, ndarray or None."
+            )
+
+        return value
+
+    @classmethod
+    def from_legacy_entity(cls, entity: Any, **overrides) -> Self:
+        """
+        Build a pydantic model from an existing geoh5py Entity-like object.
+        """
```
domfournier, line 134: As we discussed, we wouldn't maintain both versions at the same time. Ok to keep for now while we test though
```suggestion
    @classmethod
    def from_legacy_entity(cls, entity: Any, **overrides) -> Self:
        """
        Build a pydantic model from an existing geoh5py Entity-like object.
        
        TODO: Won't be needed at the end of development. Remove after final design.
        """
```

geoh5py_pydantic/entity.py
```diff
@@ -0,0 +1,146 @@
...
+            "Name": self.name,
+            "Allow delete": self.allow_delete,
+            "Allow move": self.allow_move,
+            "Allow rename": self.allow_rename,
+            "Clipping IDs": self.clipping_ids,
+            "Metadata": self.metadata,
+            "Partially hidden": self.partially_hidden,
+            "Public": self.public,
+            "Visible": self.visible,
+        }
```
domfournier: If we use `alias`, then that's just
`self.model_dump(by_alias=True)`

```suggestion
```

geoh5py_pydantic/points.py
```diff
@@ -0,0 +1,195 @@
...
+            )
+
+        return np.asarray(value, dtype="<f8")
+
+    @property
+    def vertices_array(self) -> np.ndarray:
+        """
+        Vertices as an eager numpy array.
+        """
+        return np.asarray(self.vertices)
```
domfournier, line 174: Feels like this could be the responsibility of LazyArray?

geoh5py_pydantic/points.py
```diff
@@ -0,0 +1,195 @@
...
+
+    @property
+    def extent(self) -> np.ndarray:
+        """
+        Bounding box of vertices, shaped ``(2, 3)``.
+        """
+        vertices = self.vertices_array
+        return np.c_[vertices.min(axis=0), vertices.max(axis=0)].T
+
+    def as_geoh5_vertices(self) -> np.ndarray:
```
domfournier, line 198: Similar here - could this be the responsibility of LazyArray to serialize itself with the proper dtype?
As in, the `Points` class could specify its own flavour of LazyArray to store its vertices. Then on serialization, the Points.vertices gets serialized by the LazyArray in chain.

domfournier (commented):
Excellent first pass. That's going to be SO much cleaner than the current framework.
See comments, mostly about moving responsability to LazyArray

geoh5py_pydantic/points.py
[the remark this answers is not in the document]
gmcga: done

geoh5py_pydantic/arrays.py
gmcga: makes sense

geoh5py_pydantic/arrays.py
gmcga: added a docstring explaining each param

geoh5py_pydantic/arrays.py
gmcga: agree

geoh5py_pydantic/entity.py
gmcga: I'm not sure this would work. alias affects validation and serialization but it must be a String. validation_alias is the only one that allows AliasChoices. I added a serialization_alias to each field which should allow for model_dump(by_alias=True) to work the way we want.

geoh5py_pydantic/points.py
gmcga: added a couple of methods  that should improve this system and make vertices reading/writing/validation and such simpler.

domfournier (approved):

PR #948, opened 2026-07-13
GEOPY-3000: Support Trend and plunge Property Group

domfournier:
**GEOPY-3000 - Support Trend and plunge Property Group**


geoh5py/groups/property_group_type.py
```diff
@@ -117,7 +117,21 @@ def verify(cls, children: list[Data]):
         ):
             raise TypeError(
                 "Children of 'Strike & dip' property group type "
-                "must be a list of 2 NumericData entities"
+                "must be a list of 2 NumericData entities."
+            )
+
+
+class TrendPlungeType(PropertyGroupType):
```
gmcga, line 124: Should the trend & plunge type be added to the `ui_validations` in `ui_json/constants` and to `data_parameters` in `ui_json/templates`?

gmcga (changes requested):
Looks good, just this one potential issue.

geoh5py/groups/property_group_type.py
domfournier: Good point. We are going to deprecate those soon, but need to support for now....

gmcga (approved):

PR #951, opened 2026-07-16
GEOPY-2731: Replace InputFile for BaseUIJson class

domfournier:
**GEOPY-2731 - Replace InputFile for BaseUIJson class**


geoh5py/ui_json/ui_json.py
```diff
@@ -364,30 +394,45 @@ def to_ui_json_group(
...
+            out_group: UUID | GroupForm = uuid4()
+            if self.out_group is None:
+                out_group = GroupForm(
+                    label="UIJson Group",
+                    group_type=UIJsonGroup,
+                    optional=True,
+                    enabled=True,
+                    main=False,
+                    value=uuid4(),
+                )
```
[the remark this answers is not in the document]
domfournier: Will make _prepare_options less restrictive instead

geoh5py/ui_json/ui_json.py
```diff
@@ -141,27 +140,32 @@ def flatten(self, skip_disabled=False, active_only=False) -> dict[str, Any]:
...
         """
         kwargs = {key: (item if item != "" else None) for key, item in data.items()}
 
         ui_json_class = cls.infer(**kwargs)
 
-        return ui_json_class(**kwargs)
+        if validate:
+            return ui_json_class(**kwargs)
+        else:
+            return ui_json_class.model_construct(**kwargs)  # type: ignore[return-value, arg-type]
```
benk-mira: Surprised to see in issue with 'return-value', is it not recognizing `Self`?

geoh5py/ui_json/ui_json.py
```diff
@@ -314,15 +323,43 @@ def set_values(self, copy: bool = False, **kwargs) -> BaseUIJson:
...
                 uijson.set_enabled(copy=False, **{field: value is not None})
             else:
-                setattr(uijson, field, dict_mapper(value, [entity2uuid]))
+                setattr(uijson, field, stringify(value))
 
         return uijson
 
+    def serialize(self, mode: Literal["json", "python", "str"] | str = "python"):
+        """
+        Return a demoted uijson dictionary representation the params data.
```
benk-mira:
```suggestion
        Return a demoted dictionary representation of the uijson data.
```

geoh5py/ui_json/ui_json.py
```diff
@@ -314,15 +323,43 @@ def set_values(self, copy: bool = False, **kwargs) -> BaseUIJson:
 
                 uijson.set_enabled(copy=False, **{field: value is not None})
             else:
-                setattr(uijson, field, dict_mapper(value, [entity2uuid]))
+                setattr(uijson, field, stringify(value))
 
         return uijson
 
+    def serialize(self, mode: Literal["json", "python", "str"] | str = "python"):
```
benk-mira:
```suggestion
    def serialize(self, mode: Literal["json", "python", "str"] | str = "python") -> dict | str:
```

geoh5py/ui_json/ui_json.py
```diff
@@ -314,15 +323,43 @@ def set_values(self, copy: bool = False, **kwargs) -> BaseUIJson:
...
+        :param mode: Define the schema used for serialization. Either 'json', 'python', 'str'
+
+        :return: Serialized uijson dictionary or string.
+        """
+        if mode == "str":
+            return self.model_dump_json(exclude_unset=True, by_alias=True, indent=4)
+
+        return self.model_dump(exclude_unset=True, by_alias=True, mode=mode)
+
+    def to_file_data(
```
benk-mira, line 340:
```suggestion
    def attach_file_data(
```

geoh5py/ui_json/ui_json.py
```diff
@@ -314,15 +323,43 @@ def set_values(self, copy: bool = False, **kwargs) -> BaseUIJson:
...
+        self, entity: EntityContainer, name: str | None = None
```
benk-mira, line 341: cool, don't think I knew about `EntityContainter`.  It's generic to `Objects` and `Groups`?

geoh5py/ui_json/ui_json.py
```diff
@@ -364,30 +401,45 @@ def to_ui_json_group(
...
+                    label="UIJson Group",
+                    group_type=UIJsonGroup,
+                    optional=True,
+                    enabled=True,
+                    main=False,
+                    value=uuid4(),
+                )
+
+            uijson_copy = self.set_values(
+                copy=True, **{"geoh5": geoh5, "out_group": out_group}
```
benk-mira, line 422: I don't get it..  If self.out_group is not None, this will serialize out_group as a blank UUID instead of existing out_group Form?

benk-mira (changes requested):
Looking good.  Couple suggestions/questions. What's the plan to roll this out so that we don't break every package referring to BaseUIJson?

geoh5py/ui_json/ui_json.py
domfournier: Yea same. Another weird quirk of mypy and pydantic I guess

geoh5py/ui_json/ui_json.py
domfournier: Correct

geoh5py/ui_json/ui_json.py
domfournier: I wanted to keep it consistent with the other `to_` transformers -> `to_params` , `to_ui_json_group`... It is really converting the uijson to a FilenameData equivalent, so it seems appropriate

geoh5py/ui_json/ui_json.py
domfournier: If `out_group` is not a form in the current state of uijson, then it will add it as a group form for the outgoing `option`. If it is already a form, then it just updates the "value"

domfournier:
> Looking good. Couple suggestions/questions. What's the plan to roll this out so that we don't break every package referring to BaseUIJson?

One PR per repo, until it's done.

geoh5py/ui_json/ui_json.py
benk-mira: Still don't think this is right.  If self.out_group is not None, you'll be passing `"out_group": uuid4()` to the `set_values()` call.  Shouldn't you pass the uid of the existing self.out_group object?

benk-mira (approved):

PR #957, opened 2026-07-21
GEOPY-2998: Update docs with release notes from GA 4.8 release

domfournier:
**GEOPY-2998 - Update docs with release notes from GA 4.8 release**


gmcga (changes requested):
Other than the copilot comment, the only thing I noticed was that some bullet points end with a period and others don't.

PR #962, opened 2026-07-29
GEOPY-3013: Support Collar dip and Collar azimuth drillhole attributes

domfournier:
**GEOPY-3013 - Support Collar dip and Collar azimuth drillhole attributes**

### DIFF
--- a/geoh5py/objects/drillhole.py
+++ b/geoh5py/objects/drillhole.py
@@ -361,14 +361,17 @@ class Drillhole(Points):
         if isinstance(values, (list, tuple)):
             values = np.array(values, ndmin=2)
 
-        # Already a recarray with proper type
-        if values.dtype.names == dtype.names:
-            return values
-
         if np.issubdtype(values.dtype, np.number):
             n_fields = values.shape[1]
         else:
             n_fields = len(values.dtype.names)
+            dtype = np.dtype(SURVEYS_FIELDS[:n_fields])
+            if values.dtype.descr != dtype.descr:
+                raise TypeError(f"The type of survey array must be numeric or {dtype}")
+
+        # Already a recarray with proper type
+        if values.dtype.names == dtype.names:
+            return values
 
         if n_fields not in [3, 4]:
             raise ValueError("'surveys' requires an ndarray of shape (*, 3) or (*, 4)")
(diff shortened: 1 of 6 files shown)


geoh5py/objects/drillhole.py
```diff
@@ -167,6 +177,20 @@ def collar(self, value: list | np.ndarray | None):
...
+        """
+        return self._collar_azimuth
+
+    @property
+    def collar_dip(self) -> float | None:
+        """
+        Dip of the drillhole at the collar, in degrees.
+        """
+        return self._collar_dip
+
```
gmcga, line 193: is it intentional that collar_azimuth and collar_dip have no setters?

geoh5py/objects/drillhole.py
```diff
@@ -290,25 +314,21 @@ def surveys(self) -> np.ndarray:
...
+            ]
+        ).T
 
         return surveys.astype(float)
 
     @surveys.setter
-    def surveys(self, array: np.ndarray | list | tuple):
+    def surveys(self, array: np.ndarray | list | tuple | None):
+        if array is None:
+            array = np.c_[0, self._collar_azimuth or 0.0, self._collar_dip or -90.0]
```
gmcga: if `self._collar_dip` can be given as 0.0, the `self._collar_dip or -90.0` statement would end up replacing the collar_dip value with -90.0. Not sure if this is something that could realistically arise though. Maybe it'd be better to specifically test for `self._collar_dip is None`?

gmcga (changes requested):
See comments. I also think it'd be good to try and increase the test coverage as there's a fair amount missing.

geoh5py/objects/drillhole.py
domfournier: It is on purpose, yes. A possible issue is that if we allow them to be changed, then we need to reset the `surveys`. But surveys can (most often) be an actual real array of values which we would want to preserve at all cost. These two collar values are just a fall back in case no survey is provided....

geoh5py/objects/drillhole.py
domfournier: Good catch

tests/objects/drillhole_data_test.py
```diff
@@ -273,14 +295,37 @@ def test_no_survey(tmp_path):
...
 
+def test_collar_azimuth_dip(tmp_path):
+    collar = np.r_[0.0, 10.0, 0.0]
+    h5file_path = tmp_path / f"{__name__}.geoh5"
+    max_depth = 100.0
+    n_data = 10
+    with Workspace(version=1.0).save_as(h5file_path) as workspace:
+        well = Drillhole.create(
+            workspace, collar=collar, collar_azimuth=90.0, collar_dip=-45.0
+        )
```
gmcga, line 316: Would it make sense to add a regression test that specifically tests a collar_dip value of 0.0?

gmcga (changes requested):

gmcga (approved):
Looks good

PR #966, opened 2026-08-06
GEOPY-3022: Crash opening workspace containing PropertyGroup with no name

domfournier:
**GEOPY-3022 - Crash opening workspace containing PropertyGroup with no name**

### DIFF
--- a/geoh5py/groups/property_group.py
+++ b/geoh5py/groups/property_group.py
@@ -68,19 +68,20 @@ class PropertyGroup:
         *,
         association: str | DataAssociationEnum | None = None,
         allow_delete: bool = True,
-        name: str = "Property Group",
+        name: str = "Data Group",
         on_file: bool = False,
         uid: UUID | None = None,
-        property_group_type: GroupTypeEnum | str = GroupTypeEnum.SIMPLE,
+        property_group_type: GroupTypeEnum | str = GroupTypeEnum.MULTI,
         properties: Sequence[UUID | Data | str] | None = None,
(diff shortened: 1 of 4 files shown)


geoh5py/groups/property_group.py
```diff
@@ -89,6 +90,7 @@ def __init__(  # pylint: disable=too-many-arguments
         self._properties = self._validate_properties(properties_list)
 
         self.parent.add_children([self])
+        self.on_file = on_file
```
benk-mira, line 93: Is the double self.on_file setting intentional?

benk-mira (changes requested):
just a question, pushing back for copilot suggestions - suppress or fix.

geoh5py/groups/property_group.py
domfournier: The initial False is to avoid updating the file during instantiation. This one makes it official whether it is in memory or on disk

geoh5py/objects/object_base.py
```diff
@@ -433,23 +433,17 @@ def copy(
 
     def create_property_group(
         self,
-        name=None,
-        property_group_type: GroupTypeEnum | str = GroupTypeEnum.MULTI,
         **kwargs,
     ) -> PropertyGroup:
```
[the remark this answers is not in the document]
domfournier: Removing from Concatenated - passing a kwargs strainght

benk-mira (approved):

PR #967, opened 2026-08-11
GEOPY-3039: Implement validation for GroupMultiData (Drillhole+Data) form types

domfournier:
**GEOPY-3039 - Implement validation for GroupMultiData (Drillhole+Data) form types**


gmcga (approved):
All seems good

PR #968, opened 2026-08-12
GEOPY-3013: Support Collar dip and Collar azimuth drillhole attributes

domfournier:
**GEOPY-3013 - Support Collar dip and Collar azimuth drillhole attributes**

### DIFF
--- a/tests/objects/drillhole_data_test.py
+++ b/tests/objects/drillhole_data_test.py
@@ -312,7 +312,11 @@ def test_collar_azimuth_dip(tmp_path, azimuth, dip, expected):
     n_data = 10
     with Workspace(version=1.0).save_as(h5file_path) as workspace:
         well = Drillhole.create(
-            workspace, collar=collar, collar_azimuth=azimuth, collar_dip=dip
+            workspace,
+            collar=collar,
+            collar_azimuth=azimuth,
+            collar_dip=dip,
+            end_of_hole=100.0,
         )
         well.add_data(
             {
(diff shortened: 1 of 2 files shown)


tests/objects/drillhole_data_test.py
```diff
@@ -312,7 +312,11 @@ def test_collar_azimuth_dip(tmp_path, azimuth, dip, expected):
     n_data = 10
     with Workspace(version=1.0).save_as(h5file_path) as workspace:
         well = Drillhole.create(
-            workspace, collar=collar, collar_azimuth=azimuth, collar_dip=dip
+            workspace,
+            collar=collar,
+            collar_azimuth=azimuth,
+            collar_dip=dip,
+            end_of_hole=100.0,
         )
```
gmcga, line 320: This test doesn't seem to actually cover the fix making it so that the end_of_hole value is set/maintained correctly. The test still passed on the unfixed version of `Drillhole`, which I believe is because `well.desurvey(100)` calculates/extrapolates the coordinate at the requested depth of 100 (the depth doesn't need to be less than the end of hole value) regardless of if the drillhole actually reaches that point. The parameter combinations test the direction but not length. I think some additional assertions to test that the end_of_hole value is correct should be added. 


```suggestion
        )
        assert well.end_of_hole == 100.0
        assert well.surveys[0, 0] == 100.0
```

gmcga (changes requested):

tests/objects/drillhole_data_test.py
domfournier: As a bit of context, I added the end of hole to test a bug for GA. The end of hole actually doesn't matter on the GA side - all it does is make the trace to show, and right now there is an issue if the hole is straight with X at the orgin (don't ask me why)

But sure, I can extend the test...

gmcga (approved):

PR #973, opened 2026-08-28
GEOPY-3037: Create geoh5 file with latest pydantic classes

gmcga:
**GEOPY-3037 - Create geoh5 file with latest pydantic classes**


geoh5py_pydantic/file.py
```diff
@@ -0,0 +1,144 @@
...
+    Geoh5EntityPayload,
+    Geoh5Writer,
+)
+
+
+def create_geoh5(
+    path: str | Path,
+    *,
+    project: Geoh5Project | None = None,
+    root: RootModel | None = None,
```
domfournier:
```suggestion
    project: Geoh5Project = Geoh5Project(),
    root: RootModel = RootModel(),
```
Why not directly setting a default? Would be better that project and root and never None

geoh5py_pydantic/file.py
```diff
@@ -0,0 +1,144 @@
...
+    :class:`Geoh5Writer`.
+
+    :param path: New file path. Existing files are never overwritten.
+    :param project: Project metadata and HDF5 creation settings.
+    :param root: Optional customized Root entity.
+    :return: Path to the newly created file.
+    """
+    output_path = Path(path)
+    project = project or Geoh5Project()
+    root = root or RootModel()
```
domfournier:
```suggestion
```

geoh5py_pydantic/file.py
```diff
@@ -0,0 +1,144 @@
...
+from .serialization import (
+    CHILD_COLLECTIONS,
+    TYPE_COLLECTIONS,
+    Geoh5EntityPayload,
+    Geoh5Writer,
+)
+
+
+def create_geoh5(
+    path: str | Path,
```
domfournier: Feels like this should be a method of a new `Workspace` class that is also a ContextManager with an `open` and `close` methods. The project class would hold the `ProjectAttributes`, a `Root` group and potentially a path to a geoh5. You don't implement all the mechanics that Workspace currently has, but at least be responsible for managing the file open/close when needed. If we keep the h5file on the Geoh5Writer, then the workspace should still be pickable.

geoh5py_pydantic/project.py
```diff
@@ -0,0 +1,122 @@
...
+
+DEFAULT_PAGE_SIZE = 65_536  # Same as legacy workspace default
+
+
+def _default_contributors() -> tuple[str, ...]:
+    """Evaluate the current user when a project is created, not at import time."""
+    return (getuser(),)
+
+
+class Geoh5Project(BaseModel):
```
domfournier:
```suggestion
class ProjectAttributes(BaseModel):
```

geoh5py_pydantic/file.py
```diff
@@ -0,0 +1,144 @@
...
+            output_path.unlink(missing_ok=True)
+        raise
+
+    return output_path
+
+
+def _initialize_project(
+    h5file: h5py.File,
+    project: Geoh5Project,
+    root_payload: Geoh5EntityPayload,
```
domfournier: Could this be part of the Geoh5Writer?

geoh5py_pydantic/root.py
```diff
@@ -0,0 +1,104 @@
...
+    @field_validator("uid")
+    @classmethod
+    def validate_uid(cls, value: UUID) -> UUID:
+        """The Root type UID is part of the geoh5 format contract."""
+        if value != ROOT_TYPE_UID:
+            raise ValueError(f"Root type UID must be {ROOT_TYPE_UID}.")
+        return value
+
+
+class RootModel(PydanticEntity):
```
domfournier:
```suggestion
class Root(PydanticEntity):
```

geoh5py_pydantic/file.py
gmcga: Done

geoh5py_pydantic/file.py
gmcga: Created

geoh5py_pydantic/file.py
gmcga: done

geoh5py_pydantic/workspace.py
```diff
@@ -0,0 +1,201 @@
...
+
+    path: Path | None = None
+    name: str = "GEOSCIENCE"
+    page_size: int = Field(default=DEFAULT_PAGE_SIZE, strict=True)
+    project: ProjectAttributes = Field(default_factory=ProjectAttributes)
+    root: Root = Field(default_factory=Root)
+
+    # The live writer owns h5py's unpicklable file handle. Keeping it private
+    # separates transient I/O state from the serializable Workspace model.
+    _writer: Geoh5Writer | None = PrivateAttr(default=None)
```
domfournier, line 57: Oh that's smart.

domfournier (commented):
Very nice

domfournier (approved):

PR #979, opened 2026-09-16
GEOPY-3065: Implement giftools gzfor3d forward group

domfournier:
**GEOPY-3065 - Implement giftools gzfor3d forward group**


geoh5py/groups/giftools/base.py
```diff
@@ -74,6 +74,37 @@ def merge_field(
 # This dict is never spread whole into a group's parameter dict; individual
 # fields are plucked by key or using merge_field when needed
 BASE_PARAMETERS: dict[str, Any] = {
+    "mesh": {
+        "default": "",
+        "label": "Mesh",
+        "main": True,
+        "meshType": "{b020a277-90e2-4cd7-84d6-612ee3f25051}",
```
benk-mira, line 81: can we use geoh5py class names here?

benk-mira (changes requested):

geoh5py/groups/giftools/base.py
domfournier: Doesn't look like GA respects the meshType at all. I reach out to Kris and changed them all at once if it's possible.

benk-mira (approved):

